PR #324, opened 2014-12-05
Remove expectation data from mongodb

vinayvinay:
www.agileplannerapp.com/boards/173808/cards/8852

an [earlier story](www.agileplannerapp.com/boards/173808/cards/8707) migrated expectations to a govspeak field `need_to_know`, and this got deployed to production. so we no longer need this data in mongodb.

### DIFF
--- a/db/migrate/20141205103230_remove_expectations.rb
+++ b/db/migrate/20141205103230_remove_expectations.rb
@@ -1,0 +1,9 @@
+class RemoveExpectations < Mongoid::Migration
+  def self.up
+    Edition.where(:expectation_ids.exists => true).each { |e| e.unset(:expectation_ids) }
+    Mongoid.master.collection(:expectations).drop
+  end
+
+  def self.down
+  end
+end


vinayvinay:
@alext: the task you mentioned above has been deployed to production, and travel advice publisher is at goveuk_content_models v27.1.0 which is way ahead of [26.0.0](https://github.com/alphagov/govuk_content_models/blob/master/CHANGELOG.md#2600) when we removed the `Expectation` model. good to merge?

db/migrate/20141205103230_remove_expectations.rb
```diff
@@ -0,0 +1,13 @@
...
+    Expectation.collection.drop
+  end
+
+  def self.down
+  end
+end
+
+class Expectation
+  include Mongoid::Document
+end
```
alext: Would be good to avoid this class if possible. There's a risk that running future migrations will re-create the collection because of it.  There should be a way to drop the collection without needing a model that uses it.

PR #340, opened 2014-12-23
Downtime manager

vinayvinay:
www.agileplannerapp.com/boards/173808/cards/8805

As a content designer
I want a better way to indicate on start pages when a service is down for planned maintenance
So that I don't have to re-edition multiple pages and push these through workflow, then repeat the process again once the downtime is over

related:
https://github.com/alphagov/govuk_content_models/pull/277
https://github.com/alphagov/govuk_content_api/pull/209
https://github.com/alphagov/frontend/pull/728

screen shot 2014-12-23 at 10 47 42
screen shot 2014-12-23 at 10 47 29

### DIFF
--- a/app/views/downtimes/index.html.erb
+++ b/app/views/downtimes/index.html.erb
@@ -1,3 +1,5 @@
+<% content_for :page_title, 'Downtime' %>
+
 <div class="page-title">
   <h1>Downtime</h1>
 </div>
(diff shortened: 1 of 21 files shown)


fofr:
:fireworks: :tada: :+1:

app/views/downtimes/index.html.erb
```diff
@@ -0,0 +1,53 @@
...
+  <tbody>
+    <% @transactions.each do |transaction| %>
+    <tr>
+      <td>
+        <h4 class="publication-table-title">
+          <%= link_to transaction.title, Downtime.for(transaction.artefact).present? ?
+                                          edit_edition_downtime_path(transaction) :
+                                          new_edition_downtime_path(transaction) %>
+        </h4>
+        <%= link_to "/#{transaction.slug}", "#{Plek.find('www')}/#{transaction.slug}", class: 'link-muted' %>
```
alext: This shouldn't be using `Plek.find('www')`.  Instead it should be using `Plek.new.website_root`.  The former will return something like "https://www.production.alphagov.co.uk" instead of "https://www.gov.uk"

PR #343, opened 2015-01-02
Make tabs on editions linkable, and work without js

fofr:
* Add routes for each edition tab
* Create a tab model for generating tab links, titles and for determining which tab is active
* Update the page URL when selecting tabs with Javascript
* Show the correct tab when going directly to a tab url, eg `/metadata`
* Toggle tabs when clicking on them whether using javascript or not
* Include Bootstrap recommended aria attributes

linkable-tabs


vinayvinay:
@fofr: i like this change. i've extracted the view logic from the `EditionsController` into a `TabsHelper` and moved the `View::Edition::Tab` into the `TabsHelper` because it is a view layer object, as opposed to a model object that represents something stored in the database. these changes are on a [commit on this branch](https://github.com/alphagov/publisher/compare/linkable-tabs...linkable-tabs-refactored), so if you like that change i can send you a diff, or your could cherry-pick that commit; whatever you're comfortable with.

fofr:
The view model is a pattern I've borrowed from Transition, as created by @rgarner and @jennyd. There it's used for filters, hits categories and time periods. These models begun life as helpers.

See:
* https://github.com/alphagov/transition/blob/master/app/models/view/hits/category.rb 
* https://github.com/alphagov/transition/blob/master/app/models/view/mappings/filter.rb

@rgarner Do you have an opinion on https://github.com/alphagov/publisher/blob/linkable-tabs/app/models/view/edition/tab.rb?

PR #350, opened 2015-01-15
Allow claiming of 2i when reviewer has been reset

fofr:
Resetting the 2i user through the edit edition form would set the user to an empty string, rather than nil.

* Update logic to match that used in reviewer.html.erb
* Include what was a failing test

Fixes https://govuk.zendesk.com/tickets/923214
https://www.agileplannerapp.com/boards/173808/cards/9155

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -93,7 +93,7 @@ class EditionsController < InheritedResources::Base
   end
 
   def review
-    if resource.reviewer
+    if resource.reviewer.present?
       flash[:danger] = "#{resource.reviewer} has already claimed this 2i"
       redirect_to edition_path(resource)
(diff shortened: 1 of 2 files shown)


app/controllers/editions_controller.rb
```diff
@@ -93,7 +93,7 @@ def update
   end
 
   def review
-    if resource.reviewer
+    if resource.reviewer and resource.reviewer.present?
```
alext: You don't need the `and` clause here, this line could just be `if resource.reviewer.present?`.

alext:
minor style point, otherwise looks good.

fofr:
@alext Thanks, I've updated.

alext:
Good to merge when green.

PR #353, opened 2015-01-15
New reports

elliotcm:
This adds two new reports to Publisher:
1. A report with one row per non-archived document providing the current status to be split and distributed to departments.
2. A report with one row per non-archived edition containing information to calculate churn.

These were requested by @lutgendorff and Nick Smith.


rboulton:
This looks excellent.

PR #354, opened 2015-01-16
Downtime messages begin at midnight the day before

fofr:
Update the messaging to reflect this.
cc @vinayvinay


alext:
Won't that lead to some confusion as people aren't necessarily clear on whether "midnight on Wednesday 31 December" means the beginning of Wednesday, or the end of Wednesday?

fofr:
Yes, perhaps. If that is the case we can easily move to 1am.

vinayvinay:
how about just: 'A downtime message will show from Wednesday 31 December'?

PR #356, opened 2015-01-20
Use an index for sorting on the org content report.

elliotcm:
Optimization to bring down the request time for the `organisation-content.csv` report.


elliotcm:
Will need to run `rake db:mongoid:create_indexes` when deploying.

bradleywright:
@elliotcm see alphagov-deployment#808

alext:
That's already in place for publisher - https://github.gds/gds/alphagov-deployment/blob/master/publisher/config/deploy.rb#L71

elliotcm:
Bonza.  Thanks both.

PR #357, opened 2015-01-22
Alter expand/collapse text when adding/removing parts

fofr:
Use the events emitted by the nested form javascript to trigger changes in the link text for collapsing or expanding groups. Also remove the “if valid” check on the collapse class. All parts are shown expanded at page load, and all should be collapsible.

For example:
* When a part is added and all parts were collapsed, change text to “expand”
* When a part is removed and it was the only open part, change text to “expand”


fofr:
Rebased with master so it's merge-able again.

fofr:
Intermittent test failures being addressed in https://github.com/alphagov/publisher/pull/362

vinayvinay:
Good for merge once build goes green.

PR #359, opened 2015-01-25
Improve auto generation of slugs

fofr:
When editing part titles on new editions of published guides, the slug was auto updating. This change of slug would go unnoticed and lead to those old URLs no longer working (they redirect to the guide's first page rather than 404ing). This breaks a lot of cross-linking between editions and can take a user out of the context of what they are doing.

Change the auto slug generation behaviour to:
* only update the slug if the slug is empty or has a data attribute "accepts-generated-value"
* use this data attribute to continue auto generating slug values on first editions
* if the slug started empty (eg a new part), then all subsequent changes before saving continue to generate slugs (eg typo correction in title)

This PR also moves a lot of the parts javascript into a module with tests. It includes:
* testing part re-ordering using `jquery.simulate.drag-sortable.js` as a helper
* testing part re-ordering after adding parts

https://www.agileplannerapp.com/boards/244464/cards/8236

### DIFF
--- a/test/integration/adding_parts_to_guides_test.rb
+++ b/test/integration/adding_parts_to_guides_test.rb
@@ -51,7 +51,7 @@ class AddingPartsToGuidesTest < JavascriptIntegrationTest
     assert page.has_content? random_name
   end
 
-  test "slug for parts should be automatically generated" do
+  test "slug for new parts should be automatically generated" do
     random_name = (0...8).map{65.+(rand(25)).chr}.join + " GUIDE"
 
     guide = FactoryGirl.create(:guide_edition, :title => random_name, :slug => 'test-guide')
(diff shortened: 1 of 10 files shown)


test/integration/adding_parts_to_guides_test.rb
```diff
@@ -67,6 +67,10 @@ class AddingPartsToGuidesTest < JavascriptIntegrationTest
       fill_in 'Title', :with => 'Part One'
       fill_in 'Body',  :with => 'Body text'
       assert_equal 'part-one', find(:css, ".slug").value
+
+      fill_in 'Title', :with => 'Part One changed'
+      fill_in 'Body',  :with => 'Body text'
+      assert_equal 'part-one-changed', find(:css, ".slug").value
     end
```
alext: Is it worth a corresponding test that a version 2 doesn't get the slug updated?

fofr: Yep, it was worth adding, done so in 9504c0d58d2cbfb61a81e018b3cdbc0f2db0991d

PR #360, opened 2015-01-29
Delete downtime once end_time has passed

vinayvinay:
https://www.agileplannerapp.com/boards/173808/cards/9233

we're using Sidekiq's `perform_at` to minimise the gap between downtime's `end_time` and the time that it gets deleted. this will ensure all information showed in Publisher related to downtimes is realtime. the other option is to use cron which will have to be run very frequently to achieve the same level of realtime. given there aren't many downtime objects running cron that frequently is not worth it.

### DIFF
--- a/app/workers/expired_downtime_cleaner.rb
+++ b/app/workers/expired_downtime_cleaner.rb
@@ -1,0 +1,21 @@
+class ExpiredDowntimeCleaner
+  include Sidekiq::Worker
+
+  def self.enqueue(downtime)
+    dequeue_existing_jobs(downtime)
+    perform_at(downtime.end_time, downtime.id.to_s)
+  end
+
+  def self.dequeue_existing_jobs(downtime)
+    Sidekiq::ScheduledSet.new.select do |job|
+      job['class'] == self.name && job.args.first == downtime.id.to_s
+    end.map(&:delete)
+  end
+
+  def perform(downtime_id)
+    downtime = Downtime.find(downtime_id)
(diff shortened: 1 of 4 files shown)


app/workers/expired_downtime_cleaner.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def self.dequeue_existing_jobs(downtime)
+    Sidekiq::ScheduledSet.new.select do |job|
+      job['class'] == self.name && job.args.first == downtime.id.to_s
+    end.map(&:delete)
+  end
+
+  def perform(downtime_id)
+    downtime = Downtime.find(downtime_id)
```
alext: It would be good if this gracefully handled the Downtime not existing.

app/workers/expired_downtime_cleaner.rb
```diff
@@ -0,0 +1,21 @@
+class ExpiredDowntimeCleaner
+  include Sidekiq::Worker
+
+  def self.enqueue(downtime)
+    dequeue_existing_jobs(downtime)
+    perform_at(downtime.end_time, downtime.id.to_s)
```
alext: Is it worth adding a small margin here?  I'm not sure how precise sidekiq is with its scheduling, but we don't want to end up with a race between this time and the `downtime.end_time.future?` call below.  Adding something like 30 seconds here would guarantee to avoid that.

vinayvinay:
@alext: fair points, addressed.

PR #365, opened 2015-02-06
Only save when the save button is visible and enabled

fofr:
Using keyboard shortcuts, the form could still be submitted when the save button is disabled or when it’s not visible (eg from a different tab). This behaviour was spotted via analytics events.

* Only allow keyboard shortcut to work when the save button is visible and enabled.


fatbusinessman:
This looks thoroughly sensible to me, and it all looks like it works, but I’ll hold off merging until someone who knows JavaScript more better than I do has seen it.

PR #366, opened 2015-02-06
Track flash messages in analytics

fofr:
Keep track of the success, error and warning messages displayed to
users by using events.

* Track the alert type as the action
* Track the alert message as the label
* The page URL is captured as the event category automatically.

This will allow tracking of edition saves that happen without ajax.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -35,8 +35,11 @@
         alert_class = k
       end
     %>
-    <div class="alert alert-<%= alert_class %>">
-      <%= flash[k] %>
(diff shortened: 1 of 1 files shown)


fatbusinessman:
data-module="auto-track-event"

What is this sorcery?

fofr:
The AutoTrackEvent module is within the admin template, and has mostly been used by the Transition app.

See: https://github.com/alphagov/govuk_admin_template/blob/master/app/assets/javascripts/govuk-admin-template/modules/auto_track_event.js

fatbusinessman:
Thanks for the explanation, @fofr. :+1:

PR #368, opened 2015-02-10
Fix downtime slug links in production

fofr:
The slugs were linking to:
https://www.production.alphagov.co.uk/apply-universal-credit

Instead of:
https://www.gov.uk/apply-universal-credit

* Use the public website root, ie the GOVUK_WEBSITE_ROOT environment variable, instead of `find(‘www’)`

https://www.pivotaltracker.com/story/show/87991876


alext:
Would be good if there was a test to cover this...

fofr:
Discussed adding a test with @alext on Slack and it's not trivial. Plek now has better documentation.

Tests are all passing now. Happy to merge @alext?

PR #371, opened 2015-03-16
Organ donor registration on Completed Transaction

vinayvinay:
https://trello.com/c/caDndyIX

Completed transaction pages need to optionally show organ donor registration promotion. Form fields are stored as `presentation_toggles` on `Edition`: https://github.com/alphagov/govuk_content_models/pull/292

screen shot 2015-03-18 at 14 00 27

Checking "Promote organ donor registration" shows the URL text input:
screen shot 2015-03-18 at 14 00 50

If you've chosen to promote, entering the URL is mandatory:
screen shot 2015-03-18 at 14 02 12

### DIFF
--- a/app/assets/javascripts/modules/checkbox_toggle.js
+++ b/app/assets/javascripts/modules/checkbox_toggle.js
@@ -1,0 +1,18 @@
+(function(Modules) {
+  "use strict";
+
+  Modules.CheckboxToggle = function() {
+
+    var that = this;
+
+    that.start = function(element) {
+      element.on('change', '.js-checkbox-toggle', toggle);
+
+      function toggle(event) {
+        element.find('.js-checkbox-toggle-target').toggle(element.find('.js-checkbox-toggle').is(':checked'));
+      }
+      toggle();
+    };
+  };
+
+})(window.GOVUKAdmin.Modules);
--- a/app/assets/javascripts/modules/checkbox_toggle.js
+++ b/app/assets/javascripts/modules/checkbox_toggle.js
@@ -9,7 +9,8 @@
       element.on('change', '.js-checkbox-toggle', toggle);
 
       function toggle(event) {
-        element.find('.js-checkbox-toggle-target').toggle(element.find('.js-checkbox-toggle').is(':checked'));
+        var isCheckboxChecked = element.find('.js-checkbox-toggle').is(':checked');
+        element.find('.js-checkbox-toggle-target').toggle(isCheckboxChecked);
       }
       toggle();
     };
--- a/app/views/completed_transactions/_fields.html.erb
+++ b/app/views/completed_transactions/_fields.html.erb
@@ -10,17 +10,17 @@
 
 <hr />
 <div class="row">
-  <div class="col-md-8 edition-tags" data-module="checkbox-toggle">
+  <div class="col-md-8" data-module="checkbox-toggle">
     <h3 class="remove-top-margin">Promote organ donor registration</h3>
     <p class="help-block add-bottom-margin">
-      Displays the organ donor registration promotion above the completed transaction satisfaction survey
+      Display a promotion above the satisfaction survey
     </p>
 
     <div class="form-group">
       <%= f.input :promote_organ_donor_registration,
         as: :boolean,
         input_html: { disabled: @resource.locked_for_edits?, class: "js-checkbox-toggle" },
-        wrapper_html: { class: "form-group input-md-8" } %>
+        wrapper_html: { class: "form-group input-md-8 emphasised-field" } %>
 
       <%= f.input :organ_donor_registration_url,
           required: false,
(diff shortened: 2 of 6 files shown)


jamiecobbett:
In general :+1: 

I think it would be great if @fofr could review this - I'm not sure about the placeholder URL. Unless you View Source (or similar), you can't copy the URL, so why have it? If it's the default, why don't we make it the default?

fofr:
I can't access the Trello card which may answer my question: Why does the URL need to be editable?

I agree with @jamiecobbett’s point about having a default value rather than a placeholder.

"Organ donor registration promotion" is hard to read. Using the active voice it can be made better:
"Promote organ donor registration". This could be the label for the checkbox, and checking it could toggle the display of the URL, if the URL _must_ be editable. These are mostly reckons, I still need to review the changes in-app.

vinayvinay:
@fofr: The URL needs to be editable because it contains tracking query parameters, which are unique for each completed transaction page.

About the labels, these have been suggested by Nayeema and @marksheldonGDS. I like your suggestions, would like them to confirm if they look OK.

The only reason I've not toggled the URL display based on the checkbox value is because I'm assuming that content editors might have access to the URL before they'd turn-on the promotion, and it'll be good to allow them to store those URLs without having to turn-on the promotion. Also, when they turn off the promotion, I've chosen to not blank out the URL, so that when they decide to turn it on again, they don't have to search for the URL, and also so that it remains as a record. So I've decided to keep those to fields independently editable. This is one assumption which I hoped to get tested during Product Review.

marksheldonGDS:
@vinayvinay The changes @fofr is suggesting for the labels are great, so use those. Also, as @fofr suggests we should show/hide the URL when the checkbox is toggled on/off.

vinayvinay:
@fofr: i've updated the labels and added the toggle functionality.

app/assets/javascripts/modules/checkbox_toggle.js
```diff
@@ -0,0 +1,18 @@
...
+
+    var that = this;
+
+    that.start = function(element) {
+      element.on('change', '.js-checkbox-toggle', toggle);
+
+      function toggle(event) {
+        element.find('.js-checkbox-toggle-target').toggle(element.find('.js-checkbox-toggle').is(':checked'));
+      }
+      toggle();
```
fofr: I'd move the bit inside the .toggle call into a variable above for readability. Something like:
```js
var isCheckboxChecked = element.find('.js-checkbox-toggle').is(':checked');
element.find('.js-checkbox-toggle-target').toggle(isCheckboxChecked);
```

app/views/completed_transactions/_fields.html.erb
```diff
@@ -7,4 +7,28 @@
 <% end %>
 
 <%= render partial: 'shared/common_edition_tags', locals: {f: f} %>
+
+<hr />
+<div class="row">
+  <div class="col-md-8 edition-tags" data-module="checkbox-toggle">
+    <h3 class="remove-top-margin">Promote organ donor registration</h3>
+    <p class="help-block add-bottom-margin">
+      Displays the organ donor registration promotion above the completed transaction satisfaction survey
```
fofr: This copy can be shortened too, the title already indicates the nature of the promotion so here it can just say "promotion":
> Displays a promotion above the satisfaction survey

app/views/completed_transactions/_fields.html.erb
```diff
@@ -7,4 +7,28 @@
 <% end %>
 
 <%= render partial: 'shared/common_edition_tags', locals: {f: f} %>
+
+<hr />
+<div class="row">
+  <div class="col-md-8 edition-tags" data-module="checkbox-toggle">
```
fofr: Is the `edition-tags` class necessary?

app/views/completed_transactions/_fields.html.erb
```diff
@@ -7,4 +7,28 @@
...
+
+    <div class="form-group">
+      <% f.object.presentation_toggles.each do |toggle_name, toggle_attributes| %>
+        <% toggle_attributes.each do |toggle_attribute_key, toggle_attribute_value|
+          boolean_field = toggle_attribute_value.is_a?(Boolean) %>
+          <%= f.input toggle_attribute_key,
+              as: boolean_field ? :boolean : :string,
+              required: false,
+              input_html: { disabled: @resource.locked_for_edits?, class: "#{'js-checkbox-toggle' if boolean_field}" },
+              wrapper_html: { class: "form-group input-md-8 #{'js-checkbox-toggle-target' if !boolean_field}" } %>
```
fofr: You can add the class `emphasised-field` to the wrapper so that the field matches the "Content beta" checkbox at the top.

fofr:
I've added a couple of notes, otherwise it looks good :smile:

vinayvinay:
@fofr: i've made the changes you suggested.
also, i've simplified the form a bit after you last looked: https://github.com/alphagov/publisher/commit/845f6afcc788941e989772dc91999f0cfe523d5b

fofr:
Looks good to me :+1:

fofr:
:gift_heart:

PR #373, opened 2015-04-02
Allow users to convert published Answer editions to Simple Smart Answer editions

benilovj:
The body of the previous edition is copied across.

https://www.pivotaltracker.com/n/projects/1261204/stories/89923420


jackscotti:
:+1:

PR #374, opened 2015-04-07
Fix timezone conversion issues

vinayvinay:
https://www.pivotaltracker.com/story/show/91500052

content designers noticed that scheduling an edition at 00:01 was showing-up as 01:01 in the publishing tool.

when the edition schedule time is submitted through the form, the `ApplicationController#squash_multiparameter_datetime_attributes` method squashes multi-parameter datetime into a single attribute before saving the field in the database. this logic was ignoring the local timezone, and made all field handling logic beyond this point think that `publish_at` field is in UTC.

using `Time.zone.local` preserves the timezone information, and fixes this issue.

also correcting the integration test that was masking this issue.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -27,7 +27,7 @@ class ApplicationController < ActionController::Base
       datetime_params = params.select { |k, v| k.include? attribute_name }.values.map(&:to_i)
       params.delete_if { |k, v| k.include? attribute_name }
 
-      params[attribute_name] = DateTime.new(*datetime_params) if datetime_params.present?
+      params[attribute_name] = Time.zone.local(*datetime_params) if datetime_params.present?
     end
     params
   end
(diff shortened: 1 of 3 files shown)


fofr:
Code looks good to me. Testing locally to confirm fix.

fofr:
It's confusing that the downtime feature didn't have the same problem, it also goes through the `squash_multiparameter_datetime_attributes` method, and uses `time_select` and `date_select`.

fofr:
Tested scheduled publishing and downtime locally, worked as expected.

vinayvinay:
@fofr: thanks for reviewing! the only reason i can think of why downtime timestamps didn't need to be updated is because we're doing [something extra for scheduled publishing](https://github.com/alphagov/govuk_content_models/blob/master/lib/govuk_content_models/action_processors/schedule_for_publishing_processor.rb#L6). that seems unnecessary, as rails converts from local timezone to UTC before saving time fields to database.

i'm testing to find-out if that can be removed, and will create a PR shortly.

PR #375, opened 2015-04-20
Validate placeholders against content schema

jamiecobbett:


danielroseman:
Looks like this is failing because we also need to update the base jenkins.sh script to check out the schema repo. Strangely there is also a jenkins_branches.sh, that could be replaced with a more flexible jenkins.sh as we do everywhere else.

jamiecobbett:
The tests are passing now.

> Strangely there is also a jenkins_branches.sh, that could be replaced with a more flexible jenkins.sh as we do everywhere else.

I'm not sure what you are getting at - it seems unrelated to these changes.

danielroseman:
Ah sorry, I was just confused by the existence of that script - I thought it was responsible for running the tests for the branch build, rather than handing off to `jenkins.sh`.

PR #379, opened 2015-05-27
Allow users to clear selects with include_blank

tijmenb:
Currently, even if a select dropdown has a blank option (`include_blank: true`), there is no way to clear the selection.

To enable this we have to specify `allowClear: true` in select2's config.

Fixes a particular problem whereby users cannot undo their "Primary topic" selection.

Note that when include_blank: false (the rails default), the clear button will not show up. Only selects where it was explicitly set will allow clearing after this commit.

## Before

screen shot 2015-05-27 at 13 25 20

## After

screen shot 2015-05-27 at 13 25 11

### Trello

> Once you've assigned a primary sub-topic to a piece of mainstream content, there's no way of getting rid of it. Where we've selected something from the drop-down in error or later realised it's not quite right, and there's no suitable alternative, we're having to assign a dummy topic that doesn't appear on the front end as a workaround. eg, on a document like https://publisher.preview.alphagov.co.uk/editions/5551c270ed915d2ec80002d1, if you add an entry to "Primary topic", there's no way to unset it.

https://trello.com/c/Y21ZBE9R/105-no-way-to-delete-primary-sub-topic-in-mainstream-publisher

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -11,7 +11,7 @@
 $(function () {
   $('a.preview').attr("target","_blank");
   $('form.preview').attr("target","_blank");
-  $(".select2").select2();
+  $(".select2").select2({ allowClear: true });
 })
 
 // System wide library functions


app/assets/javascripts/application.js
```diff
@@ -11,7 +11,7 @@
 $(function () {
   $('a.preview').attr("target","_blank");
   $('form.preview').attr("target","_blank");
-  $(".select2").select2();
+  $(".select2").select2({ allowClear: true });
```
fofr: This will affect all `.select2` uses in the app – is this the only one, or are there some other that need checking?
(There's a 2i assignee select2 field if I remember correctly but I think that uses a different technique)

KushalP:
If, like me, you were trying to look at the two images and weren't able to pick out the difference immediately... there's a tiny 'x' in the _Primary Topic_ field.

app/assets/javascripts/application.js
tijmenb: Correct - that assignee one uses `allowClear` as well ([assignee_select.js](https://github.com/alphagov/publisher/blob/50d8ba6ebfb66c8d2bfe2415910cdfbf351ed568/app/assets/javascripts/modules/assignee_select.js#L5
)).

The only other select2-enabled components I could find (by searching for `select2` in `app/views`) are in that same template. Those are multi-selects and therefore already have a clear button.

(Even if there were any other affected selects, that would be okay since it only applies the button to dropdowns where the original developer explicitly set `include_blank: true`.)

fofr: :+1: This is a nice improvement :smile:

PR #382, opened 2015-06-08
Associate Editions with mainstream browse pages

jackscotti:
This work is required by this [ticket](https://trello.com/c/MfDxYdta/125-tag-all-content-in-publisher-with-the-childcare-tags-using-a-data-migration). 

We need to update all content in publisher which should be tagged to childcare mainstream browse pages to have those tags.  Doing this through the publisher interface would be slow, and interfere with any existing edits in progress.

To avoid this, we will write a rake task to take a list of content and the mainstream browse pages they should be tagged to, perhaps as a CSV, and apply that to the editions in publisher.

This PR adds a rake task that associates Editions with mainstream browse pages from a CSV file.

(Tested locally with a sample CSV file)

### DIFF
--- a/lib/edition_tagger.rb
+++ b/lib/edition_tagger.rb
@@ -1,0 +1,14 @@
+class EditionTagger
+
+  def initialize(edition_tag_associations)
+    @edition_tag_associations = edition_tag_associations
+  end
+
+  def run
+    @edition_tag_associations.each do |association|
+      edition = Edition.where(slug: association[:slug]).first
+      edition.browse_pages << association[:tag]
+      edition.save!
+    end
+  end
+end
--- a/lib/edition_tagger.rb
+++ b/lib/edition_tagger.rb
@@ -1,14 +1,58 @@
 class EditionTagger
 
-  def initialize(edition_tag_associations)
+  def initialize(edition_tag_associations, logger)
     @edition_tag_associations = edition_tag_associations
+    @logger = logger
   end
 
   def run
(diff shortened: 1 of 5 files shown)


lib/edition_tagger.rb
```diff
@@ -0,0 +1,27 @@
...
+      edition.subsequent_siblings.where(state: :draft).each do |sibling|
+        association_tag_with_edition(sibling, association[:tag])
+      end
+    end
+  end
+
+  private
+
+  def association_tag_with_edition(edition, tag)
+    edition.browse_pages << tag
```
alext: It would be good to check the tag isn't already present here to avoid duplicates.

jackscotti: Good spot. Going to do it!

rboulton:
This looks pretty good, thanks.

I think it could do with some more documentation for the rake task; in particular, describing the format of the CSV it accepts.  This could be done using a multi-line "desc" for the task, similar to https://github.com/alphagov/publisher/blob/master/lib/tasks/update_mainstream_slug.rake#L1-L13 , so it can be shown by `rake -D`.

It would be great to copy the couple of paragraphs from the Trello ticket into the description of this pull request (as well as linking to the trello ticket) - that way, if the trello instance goes away for some reason, the history here still makes sense.

jackscotti:
Thank you for the feedback Richard, will do.

jackscotti:
@rboulton @alext the PR is ready for review now.

PR #386, opened 2015-06-18
Fix the organisation content report

jamiecobbett:
This report was generating an error when it encountered an artefact without any
editions.

This commit doesn't address the fact that the report takes a very long time to
generate and so users aren't able to access them because nginx times out first. The solution to that is probably to generate the reports off-line.


boffbowsh:
Re slowness, there's a story here https://www.pivotaltracker.com/story/show/93451218

jamiecobbett:
Build now passing.

issyl0:
Looks sensible to me... :partly_sunny:

PR #387, opened 2015-06-26
Offline CSV Report Generation

boffbowsh:
Story: https://www.pivotaltracker.com/story/show/93451218

This PR changes the flow for these reports so that they are no longer generated on demand, but once per hour as a cron job, and stored in `/data/uploads/publisher/reports` on the NFS mount.

When the controller serves the stored file, it will add the generation time to the filename it sends to the browser.

There's a fair bit of refactoring in this too, the reports are now a lot more standardised around the `CSVPresenter` class.

This PR is best reviewed commit-by-commit

TODO:

- [x] race condition lock
- [x] deployment change: symlink NFS to folder in `Rails.root`
- [x] cronjob to run the rake task every hour

Requisite deployment change: https://github.gds/gds/alphagov-deployment/pull/968


boffbowsh:
Related puppet change https://github.gds/gds/puppet/pull/2985

/cc @benilovj @jamiecobbett

lib/csv_report_generator.rb
```diff
@@ -0,0 +1,42 @@
+class CsvReportGenerator
+  CSV_PATH = "/data/uploads/publisher/reports"
```
alext: It would be better for this to be a directory under the `Rails.root` that can then be symlinked to the NFS mount at deploy time.  See asset-manager for prior art on this.

boffbowsh: Good point. Will do.

benilovj: @alext fixed in 4c70577.

fofr:
@benilovj Your changes look good to me :+1:

benilovj:
The Puppet and Deployment changes have been merged, I've reviewed @boffbowsh's changes and @fofr has reviewed my changes. Merging.

boffbowsh:
FWIW: :+1:

PR #391, opened 2015-07-06
Fix offline generation of Publisher reports

boffbowsh:
Addresses the comments in https://trello.com/c/QqiFmXZ3/10-reports-in-publisher-shouldn-t-time-out-medium:

- The "weird redirect" was caused by Rails trying to send e.g. `/data/vhost/publisher.preview.alphagov.co.uk/releases/20150706083530/re
ports/organisation_content.csv` because the filename was based on `Rails.root`. Nginx is only configured to server `/var/apps/publisher/reports`, so use that as the CSV path.
- Ensure "Generated" times are displayed as Local time. This seems to be the only way to achieve this - mtime is reported as being in local time, but with a UTC offset of 0. So we have to explicitly convert this to local time, which currently means being in BST. 
- Fixes a randomly-failing test. The sort order needs to be explicitly specified in the test.

There is another randomly failing test in the integration section. Fixing that seems like it doesn't belong in this PR, will fix it as part of https://trello.com/c/jI6B7tDa/19-upgrade-apps-to-supported-ruby-versions-publisher-large

### DIFF
--- a/lib/csv_report_generator.rb
+++ b/lib/csv_report_generator.rb
@@ -2,7 +2,7 @@ require 'redis'
 require 'redis-lock'
 
 class CsvReportGenerator
-  CSV_PATH = "#{Rails.root}/reports"
+  CSV_PATH = "#{ENV['GOVUK_APP_ROOT'] || Rails.root}/reports"
 
   def run!
     redis.lock("publisher:#{Rails.env}:report_generation_lock", :life => 15.minutes) do
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -34,7 +34,7 @@ class ReportsController < ApplicationController
   helper_method :report_last_updated
 
   def mtime_for(report)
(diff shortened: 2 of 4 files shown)


lib/csv_report_generator.rb
```diff
@@ -2,7 +2,7 @@
 require 'redis-lock'
 
 class CsvReportGenerator
-  CSV_PATH = "#{Rails.root}/reports"
+  CSV_PATH = "/var/apps/publisher/reports"
```
alext: It would be better to use `#{ENV['GOVUK_APP_ROOT'] || Rails.root}` here.  That way it will work in a non-deployment environment.  It'll also be resilient to changes in the filesystem layout on the servers.

boffbowsh: Nice, thanks. Didn't know about that ENV var

boffbowsh:
Updated, @alext

PR #392, opened 2015-07-07
Upgrade to Ruby 2.1.6

boffbowsh:
- Remove the debugger gem as it's not compatible with Ruby 2+
- Use StringIO instead of TempIO, as `rubygems/test_utilities.rb` no longer exists
- Fix a race condition in the Smart Answers test


benilovj:
We also have a `.travis` file - do we need to update this? Do we still even need travis?

alext:
> Do we still even need travis?

AFAIK, this doesn't use Travis at all, so I think we should remove it to avoid any confusion.

PR #393, opened 2015-07-07
Fix random test failures

boffbowsh:
Using `assert_selector` makes Capybara wait for the element to appear. This should fix the not-so-random test failures


jamiecobbett:
:clap:

PR #396, opened 2015-07-13
Support ordering browse_page tags when importing

rboulton:
Add a third column to the CSV being used to import mainstream browse
tags to indicate whether the tag should be applied as a "primary" tag
(ie, at the start of the list of browse pages, so used for breadcrumbs
on the page, or at the end of the list).

Complain if we get any value other than "TRUE", "FALSE" or unspecified
from this field, so we don't get confused when converting from the value
in the CSV to a boolean.

Needed for https://trello.com/c/QAEn1zX6/230-tag-mainstream-browse-pages-for-childcare

### DIFF
--- a/lib/edition_tagger.rb
+++ b/lib/edition_tagger.rb
@@ -7,7 +7,10 @@ class EditionTagger
 
   def run
     @edition_tag_associations.each do |association|
-      add_mainstream_browse_tag(association[:slug], association[:tag])
+      unless ["TRUE", "FALSE", nil].include? association[:primary]
+        raise "Invalid value for `primary`: #{association[:primary]}"
+      end
+      add_mainstream_browse_tag(association[:slug], association[:tag], association[:primary] == 'TRUE')
     end
     logger.info "Retagging complete"
   end
@@ -16,7 +19,7 @@ private
 
   attr_reader :logger
 
-  def add_mainstream_browse_tag(slug, tag)
+  def add_mainstream_browse_tag(slug, tag, primary)
     non_archived_editions = Edition.where(:slug => slug, :state.ne => 'archived')
 
     unless non_archived_editions.any?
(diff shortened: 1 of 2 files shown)


lib/edition_tagger.rb
```diff
@@ -33,25 +36,37 @@ def add_mainstream_browse_tag(slug, tag)
...
+    if primary
+      new_value.unshift(tag)
+    else
+      new_value << tag
+    end
+
+    if old_value == new_value
+      logger.info(
+        "Unchanged edition '#{edition.slug}' version #{edition.version_number}"
+      )
```
alext: Did you mean to return here to avoid an unnecessary save below?

lib/edition_tagger.rb
```diff
@@ -7,7 +7,10 @@ def initialize(edition_tag_associations, logger)
 
   def run
     @edition_tag_associations.each do |association|
-      add_mainstream_browse_tag(association[:slug], association[:tag])
+      unless ["TRUE", "FALSE", nil].include? association[:primary]
+        raise "Invalid value for `primary`: #{association[:primary]}"
```
KushalP: Why do you expect an invalid value here? Can you add a test to show this intent?

rboulton: I expect a risk of invalid values because this is expected to be fed by a spreadsheet created manually, and humans make typos.  A test is a good idea, though: added https://github.com/alphagov/publisher/commit/faaf043aa167e302de9982ec441a0863ab367cf6#diff-af472a3dbebbe1ff56ff17d442aeedeaR85

lib/edition_tagger.rb
rboulton: I was mainly concerned about making it obvious when the script hadn't changed anything, but that's a good idea too.  Now implemented - I've used an `else` clause instead of an early return, though: https://github.com/alphagov/publisher/commit/faaf043aa167e302de9982ec441a0863ab367cf6#diff-e1918ccfb7ea58c9de088874023288daR65

PR #397, opened 2015-07-15
Let publishers self serve cross-domain analytics tracking

fofr:
Add a "Department analytics profile" field to transaction editions so that publishers can self serve, and migrate existing cross domain analytics (from https://github.com/alphagov/frontend/blob/d940889b5bb779ca72266c03c7458bb296e010af/app/helpers/cross_domain_analytics_helper.rb).

> As a government editor, I need to be able to add cross-domain tracking to my service start page, so that I can measure the entire user journey for my service, and measure the success of any promotional efforts for my service.

screen shot 2015-07-15 at 12 12 20

https://trello.com/c/QyHwb1Zu/

Related: https://github.com/alphagov/govuk_content_models/pull/307

### DIFF
--- a/db/migrate/20150715113301_add_analytics_profiles.rb
+++ b/db/migrate/20150715113301_add_analytics_profiles.rb
@@ -1,0 +1,20 @@
+class AddAnalyticsProfiles < Mongoid::Migration
+  def self.up
+    profiles = {
+      'register-to-vote' => 'UA-23066786-5',
+      'accelerated-possession-eviction' => 'UA-37377084-12',
+      'renewtaxcredits' => 'UA-43414424-1',
+      'registered-traveller' => 'UA-47583357-4',
+    }
+
+    profiles.each do |slug, profile|
+      editions = TransactionEdition.where(slug: slug, :state.ne => 'archived')
+      editions.each do |edition|
+        edition.set(:department_analytics_profile, profile)
+      end
+    end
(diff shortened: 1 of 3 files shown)


boffbowsh:
LGTM but I don't know Publisher that well :+1:.

db/migrate/20150715113301_add_analytics_profiles.rb
```diff
@@ -0,0 +1,20 @@
...
+      'register-to-vote' => 'UA-23066786-5',
+      'accelerated-possession-eviction' => 'UA-37377084-12',
+      'renewtaxcredits' => 'UA-43414424-1',
+      'registered-traveller' => 'UA-47583357-4',
+    }
+
+    profiles.each do |slug, profile|
+      editions = TransactionEdition.where(slug: slug, :state.ne => 'archived')
+      editions.each do |edition|
+        edition.set(:department_analytics_profile, profile)
```
jamiecobbett: Any reason for set over [update_attribute](http://www.rubydoc.info/gems/mongoid/2.6.0/Mongoid/Persistence#update_attribute-instance_method)?

benilovj:
It feels like there ought to be an integration test for this.

fofr:
@benilovj I've added an integration test for Transaction Editions which was previously missing, and then a separate test for the analytics profile.

PR #398, opened 2015-07-15
Add a rake task to generate a CSV of Local Transactions

issyl0:
- Generates a CSV of the form "slug","lgsl","lgil","title".
- This might need to be done again in future, and across environments,
  hence the rake task.

Trello: https://trello.com/c/IycgidgH/52-export-list-of-local-transaction-formats-from-publisher

### DIFF
--- a/lib/tasks/export_local_transactions.rake
+++ b/lib/tasks/export_local_transactions.rake
@@ -1,0 +1,15 @@
+desc "Export Local Transactions' slugs, names, LGSL and optional LGIL codes as CSV"
+
+task :export_local_transactions => :environment do
+  require "csv"
+
+  csv_string = CSV.generate do |csv|
+    csv << ["slug","lgsl","lgil","title"]
+
+    Edition.where(_type: "LocalTransactionEdition").each do |lte|
+      csv << [lte.slug,lte.lgsl_code,lte.lgil_override,lte.title]
+    end
+  end
+
+  puts csv_string
+end


lib/tasks/export_local_transactions.rake
```diff
@@ -0,0 +1,15 @@
+desc "Export Local Transactions' slugs, names, LGSL and optional LGIL codes as CSV"
+
+task :export_local_transactions => :environment do
+  require "csv"
+
+  csv_string = CSV.generate do |csv|
+    csv << ["slug","lgsl","lgil","title"]
+
+    Edition.where(_type: "LocalTransactionEdition").each do |lte|
+      csv << [lte.slug,lte.lgsl_code,lte.lgil_override,lte.title]
```
alext: It's worth noting that `lgil_override` is unset for most local transactions.  If not set, there's [some logic](https://github.com/alphagov/govuk_content_models/blob/master/app/models/local_authority.rb#L41-L53) to pick a LGIL from the available ones. It picks the lowest lgil that's not 8 (information), but falls back to 8 if there are no others.

issyl0: Is that anything to worry about? I was surprised everything didn't have a `lgil_code`, and settled for the sometimes present `lgil_override` as per the Publishing admin interface fields.

jamiecobbett: @alext does this block merging this?

alext: I can't answer that as I don't know what the intended use of this CSV is.  I was just pointing out how that field works in case it might cause problems.

issyl0: This is mainly for Mark Sheldon and the Custom team's use in our work around local DirectGov and transactions, as far as I know. I only made it into a rake task in case we needed to run it more than once easily.

issyl0: Mark has :+1:ed the resulting CSV from this script, if that helps.

issyl0:
:sunny:

lib/tasks/export_local_transactions.rake
```diff
@@ -0,0 +1,15 @@
+desc "Export Local Transactions' slugs, names, LGSL and optional LGIL codes as CSV"
+
+task :export_local_transactions => :environment do
+  require "csv"
+
+  csv_string = CSV.generate do |csv|
+    csv << ["slug","lgsl","lgil","title"]
+
+    Edition.where(_type: "LocalTransactionEdition").each do |lte|
```
jennyd: This is exporting all editions, not only the currently published one, so services are duplicated in the output. We'll need to update this to include only the currently published edition for each service.

marksheldonGDS: It may be better to simply export the state of the Edition. Then we'll have a complete CSV that can easily be de-duped or filtered based on state.

PR #400, opened 2015-07-15
Arrange mainstream browse pages in prefered order

bishboria:
The list of tags is not orderable, but is shown during editing in the order that the tags were added. However, when the draft is saved and then re-displayed, the tags appear in alphabetical order. This has caused some confusion in the past amongst editors.

Similar solution to: https://github.com/alphagov/policy-publisher/pull/66

untitled


fofr:
Some screenshots/video of this new functionality would be awesome.

PR #401, opened 2015-07-17
Make integration tests more reliable

bishboria:
Attempt to improve the reliability of two integration tests (`campaign_edit_test` and `video_edition_create_edit_test`) as they fail regularly on CI.

These changes seem to make the test suite a lot more stable - Famous Last Words.

### DIFF
--- a/test/integration_test_helper.rb
+++ b/test/integration_test_helper.rb
@@ -1,4 +1,4 @@
-# encoding: UTF-8
+# encoding: utf-8
 
 require 'test_helper'
 require 'capybara/rails'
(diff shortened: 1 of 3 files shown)


test/integration_test_helper.rb
```diff
@@ -194,9 +195,22 @@ def save_edition_and_assert_success
...
+                      page.has_selector?("input[type=submit]#save-edition:enabled"))
+    assert save_attempted, "Failed to attempt saving the edition"
+
+    saved = (page.has_content?("edition was successfully updated.") ||
+             page.has_css?('.workflow-message', text: 'Saved'))
+    assert saved, "Failed to save the edition"
+  end
+
+  def save_edition_and_assert_success_without_ajax
+    with_javascript = false
```
boffbowsh: I think it'd be more readable without this variable

boffbowsh:
Tiny comment, but otherwise heroic effort :clap:

test/integration_test_helper.rb
rboulton: Wouldn't that make it just say `save_edition(false)` on the next line?  I think that would be quite hard to read.

rboulton:
I'm going to merge this, since I really want master to have a successful build.  Feel free to discuss better ways to make the "save_edition(...)" bit readable though, and follow up in a subsequent PR if needed.

fofr:
A little late to this, but I'm not sure what `save_edition_and_assert_success_slow` means and when to use it.

PR #406, opened 2015-07-21
Add an Edition's state to the local transaction export CSV

issyl0:
- Duplicates were appearing in the output of this because it was
  checking all editions and not differentiating between published and
  archived editions. As per
  https://github.com/alphagov/publisher/pull/398#discussion_r34978263,
  display the `state` of an edition to guard against this confusion and
  provide opportunities for filtering.

### DIFF
--- a/lib/tasks/export_local_transactions.rake
+++ b/lib/tasks/export_local_transactions.rake
@@ -4,10 +4,10 @@ task :export_local_transactions => :environment do
   require "csv"
 
   csv_string = CSV.generate do |csv|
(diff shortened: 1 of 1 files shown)


boffbowsh:
Could this utilise the Reports generation code? https://github.com/alphagov/publisher/pull/387

issyl0:
@boffbowsh This was meant to be a one-off thing, but I added a rake task just in case it needed to be done in different environments - I don't think it needs that much work. :confused: See #398.

PR #408, opened 2015-08-20
Make Fact Check email fetching more robust

boffbowsh:
This is an attempt to make the fact check email fetcher more robust and
more transparent. It seems like errors extending the lock time in Redis
may be causing emails to not be processed correctly and exit the loop
early, so we now rescue from the after_message hook and report errors
to Errbit.

Also added a few extra instances of Errbit logging in order to find
other errors in production.

Re: https://govuk.zendesk.com/agent/tickets/1121762


jamiecobbett:
I note that only Alex Tomlins is subscribed to Publisher errors in production. I've just added myself, but a wider audience receiving those would be good.

PR #409, opened 2015-08-26
Department analytics profile => Service analytics profile

jamiecobbett:
This label more closely matches reality. Not all services relate to a
department, and even if they do, they might be a separate profile from the
department-wide one.

I did contemplate renaming the field in the database, but it would be a lot of
work across several repos for little benefit.


boffbowsh:
I like :+1:

fofr:
Très bon!

PR #410, opened 2015-09-02
Replace old 'script/console' by 'rails console'

aureliojargas:
Since now Publisher is Rails 3.2.22 and `script/` is gone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ with other apps, here's how.
 
 ### Create a user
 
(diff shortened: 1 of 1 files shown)


tommyp:
Good catch. Thanks for contributing! 👍

jamiecobbett:
There's still a file there: https://github.com/alphagov/publisher/blob/master/script/rails

PR #412, opened 2015-09-03
Send content id to content store

jackscotti:
We want to use content IDs instead of slugs to represent items in curated lists, so that things like slug changes don't break the linkage. In order to do this, we need to ensure that all content in publisher has a content-id, and these content-ids are published to the publishing API.

https://trello.com/c/rmBoBPm9/290-ensure-that-all-content-in-publisher-has-a-content-id


jamiecobbett:
:fireworks:

PR #414, opened 2015-09-14
Gss codes for business support editions

mikejustdoit:
We're working on removing BusinessSupportEdition's dependence on 'area slugs' in favour of referring to areas by their (more stable) GSS codes (http://www.ons.gov.uk/ons/guide-method/geography/products/names--codes-and-look-ups/standard-names-and-codes/index.html).

Since the migration involves multiple applications and gems, we've decided to first add the GSS Codes to the BusinessSupportEditions, then think about removing the current area slugs after.

I've explained the reasoning behind these changes in the commit messages.

Changes of particular interest are in areas.js.erb and the EditionsController. I'm not 100% happy with how using Area slugs and GSS codes side-by-side is being handled, but as a temporary measure, it works and is tested.

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -204,7 +204,24 @@ class EditionsController < InheritedResources::Base
 
     def coerce_business_support_params
       if (params[:edition][:areas] && !params[:edition][:areas].kind_of?(Array))
-        params[:edition][:areas] = params[:edition][:areas].split(',').map(&:strip)
+        area_slugs = []
+        area_gss_codes = []
+
+        business_support_area_identifier_pairs.each do |area_slug, area_gss_code|
+          area_slugs << area_slug
+          area_gss_codes << area_gss_code
+        end
+
+        params[:edition][:area_gss_codes] = area_gss_codes.compact
+        params[:edition][:areas] = area_slugs.compact
       end
+    end
+
+    def business_support_area_identifier_pairs
+        params[:edition][:areas]
+          .split(',')
+          .map { |slug_and_gss_code|
+            slug_and_gss_code.split(';').map(&:strip)
+          }
     end
 end
--- a/app/views/editions/areas.js.erb
+++ b/app/views/editions/areas.js.erb
@@ -1,3 +1,3 @@
 var areas = <%= raw @areas.map { |a|
-  { id: a.slug, text: a.name, country: a.country_name, type: a.type }
+  { id: "#{a.slug};#{a.codes["gss"]}", text: a.name, country: a.country_name, type: a.type }
 }.to_json %>;
--- a/db/migrate/20150828153331_change_business_support_edition_areas_to_gss_codes.rb
+++ b/db/migrate/20150828153331_change_business_support_edition_areas_to_gss_codes.rb
@@ -1,0 +1,23 @@
+class ChangeBusinessSupportEditionAreasToGssCodes < Mongoid::Migration
+  def self.up
+    # Editing of an edition with an Archived artefact is not allowed
+    Edition.skip_callback(:save, :before, :check_for_archived_artefact)
+
+    BusinessSupportEdition.all.each do |edition|
+      edition.area_gss_codes = Area.areas_for_edition(edition).map { |area|
+        area.codes["gss"]
+      }.compact
+
+      edition.save!(validate: false) # Published editions can't be edited.
+    end
+  end
+
+  def self.down
+    # If this needs rolling back permanently, the area_gss_codes field should
+    # probably just be removed from GOV.UK Content Models.
+
+    BusinessSupportEdition.all.each do |edition|
+      edition.unset(:area_gss_codes)
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ end
 if ENV['CONTENT_MODELS_DEV']
   gem "govuk_content_models", :path => '../govuk_content_models'
 else
-  gem "govuk_content_models", '31.0.0'
+  gem "govuk_content_models", '31.1.0'
 end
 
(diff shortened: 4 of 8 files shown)


mikejustdoit:
Tried to shuffle the commits around and ended up with GitHub showing things in the wrong order. The branch is fine, it isn't affecting the outcome of the pull request, but be aware when reviewing that "Tidy up ImminenceAreasTestHelper a bit" is in the wrong place in history :clock1030:

db/migrate/20150828153331_change_business_support_edition_areas_to_gss_codes.rb
```diff
@@ -0,0 +1,20 @@
+class ChangeBusinessSupportEditionAreasToGssCodes < Mongoid::Migration
+  def self.up
+    # Editing of an edition with an Archived artefact is not allowed
+    Edition.skip_callback(:save, :before, :check_for_archived_artefact)
+
+    BusinessSupportEdition.all.each do |edition|
+      edition.area_gss_codes = Area.areas_for_edition(edition).map(&:gss).compact
```
jennyd: Since we decided to [nest the GSS codes](https://github.com/alphagov/imminence/blob/master/app/presenters/areas_presenter.rb#L39-L41) we actually need `codes["gss"]` here.

mikejustdoit: Thanks, @jennyd, I've fixed that now

app/views/editions/areas.js.erb
```diff
@@ -1,3 +1,3 @@
 var areas = <%= raw @areas.map { |a|
-  { id: a.slug, text: a.name, country: a.country_name, type: a.type }
+  { id: "#{a.slug};#{a.codes["gss"]}", text: a.name, country: a.country_name, type: a.type }
```
jennyd: I agree that putting both values in the form isn't pretty, but since it's only going to be here temporarily while we migrate to using GSS codes I think it's OK. I'm happy with doing it in this template for the same reason.

app/controllers/editions_controller.rb
```diff
@@ -204,7 +204,24 @@ def report_state_counts
...
+        end
+
+        params[:edition][:area_gss_codes] = area_gss_codes.compact
+        params[:edition][:areas] = area_slugs.compact
       end
     end
+
+    def business_support_area_identifier_pairs
+        params[:edition][:areas]
+          .split(',').map(&:strip)
```
jennyd: Minor point: the first `.map(&:strip)` isn't needed here - the second one will do everything that the first one does.

db/migrate/20150828153331_change_business_support_edition_areas_to_gss_codes.rb
```diff
@@ -0,0 +1,22 @@
...
+      edition.save!(validate: false) # Published editions can't be edited.
+    end
+  end
+
+  def self.down
+    # If this needs rolling back perminantly, the area_gss_codes field should
+    # probably just be removed from GOV.UK Content Models.
+    #
+    # .up is only adding data, not changing existing data, so I assume
+    # a temporary rollback doesn't really require any action.
```
jennyd: I think we should remove the `area_gss_codes` field in the `down` method so that we can keep the data clean, rather than relying on ignoring extra fields in code if we need to roll back. We should be able to migrate the database back to its original state by rolling back this migration, and when we've switched over to using the GSS codes we should remove the `areas` field in another migration.

mikejustdoit: @jennyd, you'll have to show me how to do that in the MongoDB/Rails world. All I could think of was doing something like `edition.area_gss_codes = []` to empty the field

jennyd: It looks like we've used [`unset`](http://mongoid.github.io/en/mongoid/docs/persistence.html) for this [in the past](https://github.com/alphagov/publisher/search?utf8=%E2%9C%93&q=unset) (I didn't know how to do that either).

mikejustdoit: Thanks @jennyd, I've added an `unset` and it worked fine for a `db:rollback` and `db:migrate` on my dev VM

jennyd:
This doesn't appear to be preserving all the GSS codes which were populated by the migration when the areas are edited subsequently :( I can't see exactly how to fix it - steps to reproduce:

- before running the migration, pick a business support scheme to test with and create a new edition to edit if it doesn't have one already
- inspect the latest edition for the scheme in the console: its `areas_gss_codes` will be an empty array
- run the migration
- inspect the latest edition again (reloading first): at this point it will have its `area_gss_codes` field populated with the GSS codes for all of its `areas`
- add an area to the related areas for the scheme in the UI and save it
- reload and inspect the latest edition again: at this point `area_gss_codes` has only the GSS code for the area that was added, and all the others have been lost

I thought at first that the codes might be being lost when a new edition is created, but that isn't the case - all the GSS codes which were set by the migration are still there on a newly created edition before it's edited.

db/migrate/20150828153331_change_business_support_edition_areas_to_gss_codes.rb
jennyd: The migration seems to do the right thing :+1:

mikejustdoit:
Thanks @jennyd, I'll look into that this morning :scream_cat:

mikejustdoit:
I've addressed all of @jennyd's comments except for the last one (where it was made clear that I'd overlooked some of the existing implementation details of the "related areas" selection interface).

I'm gonna close this for now and open a separate pull request doing some necessary work in preparation for this change to work properly.

PR #418, opened 2015-10-01
Use <select multiple="multiple"> for "related areas"

mikejustdoit:
Editors of Business Support Schemes can specify geographic areas (regions, counties, local authorities) related to a particular scheme.

Until now, this was implemented using a `<textarea>` progressively enhanced with select2. Without JavaScript enabled, an editor would need to enter a comma-separated list of the related areas' slugified names.

Using a "multiple" `<select>` instead doesn't change the JavaScript-on behaviour, but means that a user with no JavaScript can select from a list of area names.

### DIFF
--- a/app/assets/javascripts/views/business_support/areas_relator.js
+++ b/app/assets/javascripts/views/business_support/areas_relator.js
@@ -1,52 +1,34 @@
 $(document).ready(function() {
   "use strict";
 
-  var $relatedAreasTextArea = $("#edition_areas"),
+  var $relatedAreasSelect = $("#edition_areas"),
       bindAreasSelection = function (selector, filter) {
         $(selector).change(function () {
           var self = this;
           $('.areas-chkbx').each(function (i,e) {
             if (e !== self) $(e).prop('checked', false);
           });
-          $relatedAreasTextArea.select2("data", []);
+          $relatedAreasSelect.select2("data", []);
           if ($(this).is(':checked')) {
-            $relatedAreasTextArea.select2("data", $.map(areas, filter));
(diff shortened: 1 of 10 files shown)


mikejustdoit:
#### With JavaScript enabled (this view isn't changing)
related-areas-with-js-enabled

#### Without JavaScript before this change
original-related-areas-with-no-js

#### Without JavaScript after this change
new-related-areas-with-no-js

app/assets/javascripts/views/business_support/areas_relator.js
```diff
@@ -1,52 +1,50 @@
...
+        $relatedAreasSelect.select2("data", $.map(areas(), filter));
+      }
+    });
+  };
 
   // http://ivaynberg.github.io/select2/select-2.1.html
-  $relatedAreasTextArea.select2({
+  $relatedAreasSelect.select2({
     width: "75%",
-    multiple: true,
```
fofr: Presumably this option is now implied by using a select with the attr multiple?

fofr:
I :heart: this. Great simplification. Works better without JS, code is better, simpler and clearer all round. You've cleaned up one of the gnarliest parts of Publisher.

app/assets/javascripts/views/business_support/areas_relator.js
mikejustdoit: Yes! Actually, select2 complained that it would rather infer this from the <select> element's attributes. I tried to describe this in the commit message (with "- Remove a lot of custom select2 configuration"), but I can make that a bit clearer still if you'd like?

jamiecobbett:
:heart:

PR #419, opened 2015-10-02
Add rake task for fast republishing

tijmenb:
We have added a `content_id` for all artefacts owned by this application in https://github.com/alphagov/panopticon/pull/298

We'd like to send these documents to the content-store, to make sure the new `content_id` is in there as well (for example, https://www.gov.uk/api/content/check-register-rents).

This PR adds a rake task - `rake publishing_api:republish_content` to republish all published editions quickly. 

There's [a rake task named `reregister:all` that does the same but also for panopticon](https://github.com/alphagov/publisher/blob/master/lib/tasks/register.rake#L6). @rboulton and I decided against using that because we don't really know what the effect of the republishing will be on panopticon, and because it is very slow.

Trello: https://trello.com/c/rmBoBPm9/290-ensure-that-all-content-in-publisher-has-a-content-id

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -1,0 +1,11 @@
+namespace :publishing_api do
+  task :republish_content => [:environment] do
+    puts "Scheduling republishing of #{Edition.published.count} editions"
+
+    Edition.published.each do |edition|
+      PublishingAPINotifier.perform_async(edition.id, "republish")
+    end
+
+    puts "Scheduling finished"
+  end
+end
--- a/app/workers/publishing_api_notifier.rb
+++ b/app/workers/publishing_api_notifier.rb
@@ -1,20 +1,12 @@
-require "gds_api/publishing_api"
-
 class PublishingAPINotifier
   include Sidekiq::Worker
 
-  def perform(edition_id)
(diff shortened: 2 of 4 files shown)


lib/tasks/publishing_api.rake
```diff
@@ -0,0 +1,11 @@
+namespace :publishing_api do
+  task :republish_content => [:environment] do
+    puts "Scheduling republishing of #{Edition.published.count} editions"
+
+    Edition.published.each do |edition|
```
benlovell: If (and only if) the result set is large, and given you only need the ID perhaps you could just `Edition.published.pluck(:id).each(&blk)` to avoid having to hydrate the full `Edition`?

benlovell: OK, ignore this. It's not an arel relation :see_no_evil:

tijmenb: Mongo :disappointed: 

Mongoid 3.X has pluck, but we're on 2.5. https://github.com/mongodb/mongoid/blob/master/CHANGELOG.md#310

benlovell: :sob:

mobaig:
LGTM.

PR #420, opened 2015-10-06
GSS codes for Business Support Editions

mikejustdoit:
We're working on removing BusinessSupportEdition's dependence on 'area slugs' in favour of referring to areas by their (more stable) GSS codes (http://www.ons.gov.uk/ons/guide-method/geography/products/names--codes-and-look-ups/standard-names-and-codes/index.html).

This is the new, improved attempt at adding GSS codes to these Editions. @jennyd's review of the first attempt (https://github.com/alphagov/publisher/pull/414) highlighted the need for more preparatory work, which has now been done and merged (https://github.com/alphagov/publisher/pull/418).

Since the migration involves multiple applications and gems, we've decided to first add the GSS Codes to the BusinessSupportEditions, then think about removing the current area slugs after.

I've explained the reasoning behind each change in their commit messages.

Changes of particular interest are in AreasHelper and the EditionsController. I'm not 100% happy with how using Area slugs and GSS codes side-by-side is being handled, but as a temporary measure, it works and is tested.

The story is https://trello.com/c/XsykRbHp/45-publisher-and-business-support-api-should-use-gss-codes.

### DIFF
--- a/test/functional/editions_controller_test.rb
+++ b/test/functional/editions_controller_test.rb
@@ -215,6 +215,33 @@ class EditionsControllerTest < ActionController::TestCase
     assert_equal ["oil-and-gas/wells"], edition.additional_topics
   end
 
+  context "with Business Support areas" do
+    setup do
+      artefact = FactoryGirl.create(:artefact)
+
+      @business_support_edition = FactoryGirl.create(
+        :business_support_edition,
+        panopticon_id: artefact.id,
+      )
+
+      update_params = {
+        # select2 produces an array beginning with an empty string
+        "areas" => ["", "northern-ireland", "yorkshire-and-the-humber"],
+      }
+
+      post :update,
+        :id => @business_support_edition.id,
+        :edition => update_params
+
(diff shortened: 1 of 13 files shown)


test/functional/editions_controller_test.rb
```diff
@@ -215,6 +215,42 @@ class EditionsControllerTest < ActionController::TestCase
...
+
+      post :update,
+        :id => @business_support_edition.id,
+        :edition => update_params
+
+      @business_support_edition.reload
+    end
+
+    should "update area slugs" do
+      assert_equal ["northern-ireland", "yorkshire-and-the-humber"],
```
tommyp: Representin' 👍👍👍👍

tommyp:
This looks fine to me, but I think someone else with more knowledge of Publisher should hit merge.

jennyd:
:briefcase: :snail: :no_entry_sign: :shell: :gem: :maple_leaf: :it:

mikejustdoit:
:joy_cat:

PR #426, opened 2015-10-26
Drop business support area slugs

tommyp:
Most of this work was done by my main man @mikejustdoit. I'm just picking it up now that he's departed this world.

As we want to move away from using slugs for areas, this PR removes the slugs from business support areas. There's not a lot of interesting code changes in here as it's removing support and we've already added GSS codes in #424. [Ticket](https://trello.com/c/XsykRbHp/45-publisher-and-business-support-api-should-use-gss-codes).


app/controllers/editions_controller.rb
```diff
@@ -196,23 +196,8 @@ def report_state_counts
...
-        area_gss_codes = []
-
-        business_support_area_identifier_pairs.each do |area_slug, area_gss_code|
-          area_slugs << area_slug
-          area_gss_codes << area_gss_code
-        end
-
-        params[:edition][:area_gss_codes] = area_gss_codes.compact.reject(&:empty?)
-        params[:edition][:areas] = area_slugs.compact.reject(&:empty?)
+      if (params[:edition][:area_gss_codes])
```
benlovell: It's idiomatic to drop the parens unless you need to force precedence.

tommyp: Good catch. Fixed up the commit.

h-lame:
Seems legit.  

Do we need a PR on govuk_content_models to removes the `areas` field from the model too?  Or do we need to sweep all apps that use these models for any `areas` usages first?

tommyp:
@h-lame I'm honestly not sure. @jennyd might know?

jennyd:
Yes, we should remove `areas` from content models. I don't think areas are used by any other apps, but it's worth checking first.

h-lame:
A manual search via github reveals that only this repo and govuk_content_models and govuk_content_api make direct references to `BusinessSupportEdition`.  Of course other repos could handle these objects implicitly via `Edition`, but a search for references to `areas` suggests that nothing cares about the areas field in a way that relates back to this type of `Edition`.

The only places that do refer to areas in a way that involves `BusinessSupportEditions` appear to be migrations in this repo and an importer class.  We can drop the importer as it's only purpose is to populate the areas field and we're about to drop it.  The migrations are another matter - they will become un-runnable in future when we drop the field from the model as there's a mismatch between the code at the time the migration was written (the past when `BusinessSupportEdition`s have an `areas` field) and the code at the time the migration might be run (the present or future when `BusinessSupportEdition`s no longer have an `areas` field).  This shouldn't be a huge problem as they will already have been run in production (at which point there's an argument for them being deletable - as the db is the point of truth, not the text files that got us there).  We could just remove the offending lines and have empty migrations (perhaps with a comment explaining why).

h-lame:
New commit added to remove the importer.  Decided to leave the migrations as-is (there's a migration that tries to invoke a long gone rake task, so I assume this is BAU).  Raised alphagov/govuk_content_models#351 to remove the field from the model.

benilovj:
LGTM :+1:

PR #428, opened 2015-10-29
Convert any format to any format

binaryberry:
This enables converting any format to any format, except for local transactions which were a special case  (they require a lgsl code, so they are out of the scope of the story).

When reviewing, it'd be good to check I didn't lose any relevant test coverage by deleting the ad-hoc conversion tests and replacing them with the automated/generic version.

For tests to pass, it will require a govuk_content_models gem bump to take into account [these changes](https://github.com/alphagov/govuk_content_models/pull/350).

cc @boffbowsh 

screen shot 2015-11-02 at 16 53 21

[Trello](https://trello.com/c/p7wzEHo0/79-self-serve-changing-a-format-to-almost-any-format-in-mainstream-publisher-medium)

PR #429, opened 2015-11-05
Bump govuk_content_models to 32.0.0

benilovj:
This involves removing references to the `BusinessSupportEdition#areas`.

/cc @jennyd @tommyp @h-lame


jamiecobbett:
There's nothing in the commits that says why you're doing this. I think its to include [these changes](https://github.com/alphagov/govuk_content_models/pull/350)?

benilovj:
@jamiecobbett I've added a comment to the gem bump commit. This is predominantly to finish the work started in #424 and #426, which we need to merge before we can finish #428.

PR #430, opened 2015-11-05
Fix sorting of GSS codes for regions for business support editions

jennyd:
Imminence still returns a [fake 'England' region](https://github.com/alphagov/imminence/blob/1295ff3c5b6caeb2a16f24d34165c369b237ef03/lib/mapit_api.rb#L48-L50), which doesn't have a GSS
code. That means that there will be a `nil` in an array of GSS codes for all
areas returned from Imminence, which needs to be removed by `compact` before
the array can be sorted; otherwise we get `ArgumentError: comparison of
NilClass with String failed`.

### DIFF
--- a/app/helpers/areas_helper.rb
+++ b/app/helpers/areas_helper.rb
@@ -20,12 +20,12 @@ module AreasHelper
   def all_regions?(edition)
     Area.regions.map { |area|
       area.codes["gss"]
-    }.sort.compact == edition.area_gss_codes.sort
+    }.compact.sort == edition.area_gss_codes.sort
   end
 
   def english_regions?(edition)
     Area.english_regions.map { |area|
       area.codes["gss"]
-    }.sort.compact == edition.area_gss_codes.sort
+    }.compact.sort == edition.area_gss_codes.sort
   end
 end


benilovj:
Is it worth adding a test which captures the scenario you described, @jennyd, to prevent breakage in future?

h-lame:
I might be reading this wrong, but I think that according to the data in https://geoportal.statistics.gov.uk/geoportal/catalog/search/resource/details.page?uuid=%7B3EEE064E-DD8A-4193-B5C2-3AB4F14AADD8%7D England does have a GSS code: it's `E92000001`.  Maybe we should put that in?

jennyd:
The England region doesn't exist in MapIt so will never be returned for a postcode lookup, so it's meaningless to have any schemes tagged with it. I'd rather stop Imminence adding the extra region in at all rather than appear to support it more fully, since it doesn't work as intended in the finder.

In the event of our MapIt data missing a GSS code for any other area, though, it's better for Publisher to handle that gracefully rather than prevent anyone editing business support schemes, so I'll do this more explicitly as @jamiecobbett suggested in :meat_on_bone: :space_invader:.

jennyd:
Closing in favour of https://github.com/alphagov/publisher/pull/431.

PR #431, opened 2015-11-06
Exclude areas without GSS codes

jennyd:
At the moment on master visiting the edit page for any business support edition [errors with `ArgumentError: comparison of NilClass with String failed`](https://errbit.preview.alphagov.co.uk/apps/5304e0470da1152df9000004/problems/563b6a1d65786306a12e0500), which is caused by trying to sort an array of GSS codes which contains a `nil` as well as strings. This was introduced by https://github.com/alphagov/publisher/commit/86b5a5ee2a82a71bbe4c4d253c28987fb219b8ec.

The `nil` is for a [fake `England` region](https://github.com/alphagov/imminence/blob/1295ff3c5b6caeb2a16f24d34165c369b237ef03/lib/mapit_api.rb#L48-L50) which Imminence adds to the regions returned by MapIt, and no GSS code is included in that region's data. That region already doesn't work as intended (it isn't in MapIt so it's never returned to the business support API for a postcode lookup, so is never used to find schemes via the content API even if some were tagged with it) so we should remove it. In the meantime we should handle it safely here.

This PR supercedes https://github.com/alphagov/publisher/pull/430 and follows suggestions from @benilovj and @jamiecobbett.

### DIFF
--- a/app/helpers/areas_helper.rb
+++ b/app/helpers/areas_helper.rb
@@ -20,12 +20,12 @@ module AreasHelper
   def all_regions?(edition)
     Area.regions.map { |area|
       area.codes["gss"]
-    }.sort.compact == edition.area_gss_codes.sort
+    }.sort == edition.area_gss_codes.sort
   end
 
   def english_regions?(edition)
     Area.english_regions.map { |area|
       area.codes["gss"]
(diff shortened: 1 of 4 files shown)


jamiecobbett:
:heart:

PR #435, opened 2015-11-11
Use Rails form helper for conversion form

binaryberry:
This will allow the authentication token to be passed through and
therefore fix the form.

Instead of Edition.subclasses, I am now using the list of formats provided in govuk_content_models, so that all formats are available and loaded when needed by the select tag in the view.

cc @benilovj

[Trello](https://trello.com/c/p7wzEHo0/79-self-serve-changing-a-format-to-any-format-in-mainstream-publisher-medium)

### DIFF
--- a/app/helpers/editions_helper.rb
+++ b/app/helpers/editions_helper.rb
@@ -36,6 +36,11 @@ module EditionsHelper
     end
   end
 
+  def format_conversion_select_options(edition)
+    possible_target_formats = Edition.convertible_formats - [edition.artefact.kind]
+    possible_target_formats.map{|format_name| [format_name.humanize, format_name]}
+  end
+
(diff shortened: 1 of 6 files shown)


benilovj:
LGTM :+1: merge when green

app/helpers/editions_helper.rb
```diff
@@ -36,6 +36,11 @@ def browse_options_for_select(grouped_collections)
     end
   end
 
+  def format_conversion_select_options(edition)
+    possible_target_formats = Edition.convertible_formats - [edition.artefact.kind]
+    possible_target_formats.map{|format_name| [format_name.humanize, format_name]}
```
jamiecobbett: Its a bit late now, but our styleguide says we should have spaces before and after the curl braces.

PR #436, opened 2015-12-07
Add rake task for detecting "ghost" interactions

h-lame:
The CSV of interactions from local.directgov includes a complete list of all interactions in their database.  Those that have been deleted are marked with an "X" instead of a URL.  Our importer handles this to add/remove/update our DB so it mirrors this file.  However, we've noticed lots of "ghost" interactions where we have a record of an interaction that does not appear in the CSV file.  This shouldn't happen, but it does.  While we get to the root of the problem this rake task allows us to generate a CSV of all interactions that are in our DB but not in the CSV.

We iterate over all the local interactions stored against all the local authorities in our DB and so-doing we can detect 3 types of ghost:

 * interaction_in_input_to_be_deleted - any interaction that still lives in our DB but is listed as an "X" in the CSV.  Assumption here is that we simply haven't run the importer yet and it will be cleared on next import.
 * interaction_not_in_input - an interaction in our DB where the authority exists in the CSV but the interaction does not
 * authority_not_in_input - an interaction in our DB where the authority does not exist in the CSV

### DIFF
--- a/lib/tasks/check-for-ghosts.rake
+++ b/lib/tasks/check-for-ghosts.rake
@@ -1,0 +1,37 @@
+require 'local_authority_interaction_ghost_detector'
+
+namespace :check_for_ghosts do
+  desc "Compare local authority interactions in our DB against the local.directgov CSV export to detect \"ghosts\"."
+  task :run => :environment do
+    File.open(Rails.root.join('local_authority_service_details.csv'), 'r:windows-1252:UTF-8') do |input|
+      detector = LocalAuthorityInteractionGhostDetector.new(input)
+
+      output_filename = Rails.root.join("ghosts_in_local_authority_interactions_#{Time.zone.today.iso8601}.csv")
+      CSV.open(output_filename, 'w') do |output|
+        output.puts ['LA Name', 'LA SNAC', 'LGSL', 'LGIL', 'URL']
+
+        current_la = nil
+        total_count = ghosts_count = current_ghost_count = 0
+
+        detector.detect_ghosts do |la, lai, ghost_status|
+          if current_la != la
+            if current_la.present?
+              puts " - #{current_ghost_count} ghosts"
+            end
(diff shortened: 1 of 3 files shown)


h-lame:
New commits add a task for destroying ghosts.

lib/tasks/check-for-ghosts.rake
```diff
@@ -0,0 +1,74 @@
...
+require 'local_interaction_importer'
+require 'local_authority_interaction_ghost_detector'
+
+namespace :check_for_ghosts do
+  def detector
+    input = LocalInteractionImporter.fetch_data()
+    LocalAuthorityInteractionGhostDetector.new(input)
+  end
+
+  task :run => :environment do
```
jennyd: Could you add a `desc` for this so that it appears in the list of tasks?

h-lame: Oh, yes - added in a fixup that I've rebased in.

lib/tasks/check-for-ghosts.rake
```diff
@@ -0,0 +1,110 @@
...
+        total_count = removed_count = ghosts_count = 0
+        output_filename = with_output_file 'removed_ghosts_in_local_authority_interactions' do |output|
+          to_keep = []
+          to_remove = []
+          iterate_over_interactions(
+            on_new_authority: ->(current_la, new_la) do
+              if current_la.present?
+                current_la.local_interactions.substitute(to_keep)
+                to_remove.each do |(lai, ghost_status)|
+                  output.puts [current_la.name, current_la.snac, lai.lgsl_code, lai.lgil_code, lai.url]
```
jennyd: This output line is missing the status at the end.

h-lame: So it's not.  It is now that I've rebased.

jennyd:
Nice work :ghost:

PR #437, opened 2015-12-17
Created new rake task to import local transactions and remove ghosts

sihugh:
The current process to import local transactions from [http://local.direct.gov.uk/Data/](http://local.direct.gov.uk/Data/local_authority_service_details.csv) does not check to see if existing entries in our database have been removed from the CSV.  This has resulted in a number of "ghost" entries which could still be visible in GOV.UK.

I have added the `local_transactions:fetch_and_clean` task to perform the import as before, but then
subsequently run the `check_for_ghosts:remove` task to do the clean up.

The `check_for_ghosts:remove` task also now aborts if it fails checks.  This is to ensure that a run failure results in a non-zero return code that will be picked up in the monitoring software.

@jennyd has changed `schedule.rb` in deployment to call the new rake task.

https://trello.com/c/bIG472Xz/233-automate-removal-of-ghost-local-interactions

### DIFF
--- a/lib/tasks/local_transactions.rake
+++ b/lib/tasks/local_transactions.rake
@@ -5,6 +5,12 @@ namespace :local_transactions do
     LocalAuthorityDataImporter.update_all
   end
 
+  desc "Download and import services, interactions and contacts, and subsequently remove 'ghost' interactions"
+  task :fetch_and_clean => :environment do
+    Rake::Task["local_transactions:fetch"].invoke
+    Rake::Task["check_for_ghosts:remove"].invoke
+  end
+
   desc "Dowload the latest contact list CSV from Local Directgov and import"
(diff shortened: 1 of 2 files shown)


lib/tasks/local_transactions.rake
```diff
@@ -5,6 +5,12 @@ namespace :local_transactions do
     LocalAuthorityDataImporter.update_all
   end
 
+  desc "Download and import services, interactions and contacts, and subsequently remove 'ghost' interactions"
+  task :fetch_and_clean => :environment do
+    Rake::Task["local_transactions:fetch"].invoke
+    Rake::Task["check_for_ghosts:remove"].invoke
```
jamiecobbett: :ghost: :bust_in_silhouette:

PR #438, opened 2015-12-18
Remove `therubyracer` dependency

benlovell:
This dependency will attempt to build `libv8` thus contributing to
global warming. Instead, we can rely on `ExecJS` finding `nodejs` as
your chosen JavaScript runtime as everyone has this installed already.


floehopper:
Am I right in thinking this relies on `node.js` being installed on the server in order for asset pre-compilation to work? Is that a valid assumption?

benlovell:
@floehopper According to puppet, the ruby app servers have the `nodejs` class, so yes.

PR #440, opened 2015-12-31
Bump admin gem to 4.0.1

fofr:
* Allow environment_style to be set to integration. Add custom favicon.
* Use admin template config
  * Use standard sign out block in navbar
  * Set app title in config


benlovell:
Needs a rebase but looks good otherwise.

PR #443, opened 2016-01-28
Import Home page URL from local contact csv

h-lame:
This is part two of https://trello.com/c/td7NzjPG/261-use-local-authority-home-page-url-when-we-don-t-have-a-matching-transaction

There is a PR on frontend that uses this newly imported attribute in the "we matched your postcode to an LA but don't have a service endpoint for it" case (see: https://github.com/alphagov/frontend/pull/914).

### DIFF
--- a/lib/local_contact_importer.rb
+++ b/lib/local_contact_importer.rb
@@ -17,18 +17,24 @@ class LocalContactImporter < LocalAuthorityDataImporter
     authority.name = decode_broken_entities( row['Name'] )
     authority.contact_address = parse_address(row)
     authority.contact_phone = decode_broken_entities( row['Telephone Number 1'] )
-    url = row['Contact page URL']
-    unless url.blank? || url.start_with?("http://") || url.start_with?("https://")
-      url = "http://" + url
-    end
-    authority.contact_url = url
+    authority.contact_url = parse_url( row['Contact page URL'] )
     authority.contact_email = row['Main Contact Email']
+    authority.homepage_url = parse_url( row['Home page URL'] )
     authority.save!
   end
 
   def parse_address(row)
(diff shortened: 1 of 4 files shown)


lib/local_contact_importer.rb
```diff
@@ -17,12 +17,9 @@ def process_row(row)
...
     authority.contact_address = parse_address(row)
     authority.contact_phone = decode_broken_entities( row['Telephone Number 1'] )
-    url = row['Contact page URL']
-    unless url.blank? || url.start_with?("http://") || url.start_with?("https://")
-      url = "http://" + url
-    end
-    authority.contact_url = url
+    authority.contact_url = parse_url( row['Contact page URL'] )
     authority.contact_email = row['Main Contact Email']
+    authority.homepage_url = parse_url( row['Home page URL'] )
```
issyl0: Human linter: could you remove the spaces between the brackets as per the style guide?

jamiecobbett: Maybe it's [time](https://github.com/alphagov/government-frontend/pull/84)

issyl0: Yes, it's on my list.

h-lame: Ah, well, while I agree with the human (or robot) linter here, I'm following the pattern that's established in the file.  I think consistency within the file trumps linting (until we auto lint and correct in one fell swoop).

issyl0: Uuuh. We should add linting and fix it all in one go then, at a later date.

h-lame:
With https://github.com/alphagov/govuk_content_models/pull/356 merged and released, this PR is ready to be reviewed and merged now.  The tests are green!

PR #444, opened 2016-01-28
Notify publishing api

mgrassotti:
- migrate editions publishing notifications to use the publishing api v2
- create a new sidekiq worker “publishing_api_update_notifier” to notify the publishing-api on editions updates.


mgrassotti:
https://trello.com/c/HQ0MBmWv/447-publisher-put-content-item-when-creating-editing-drafts

app/workers/publishing_api_notifier.rb
```diff
@@ -5,8 +5,8 @@ def perform(edition_id, update_type = "normal")
     edition = Edition.find(edition_id)
```
danielroseman: This class probably needs to be renamed as it is now only responsible for calling the publish endpoint. Maybe PublishingAPIPublisher?

app/workers/publishing_api_notifier.rb
```diff
@@ -5,8 +5,8 @@ def perform(edition_id, update_type = "normal")
...
     presenter = PublishedEditionPresenter.new(edition)
     document_for_publishing_api = presenter.render_for_publishing_api(republish: update_type == "republish")
```
danielroseman: We don't need the full presenter here, just the content_id which is available from `edition.artefact`.

lib/enhancements/edition.rb
```diff
@@ -47,6 +47,8 @@ def self.state_names
     all_of(for_user(user).selector, internal_search(term).selector)
   }
 
+  after_save :notify_udpate_publishing_api
```
danielroseman: typo: `update`.

app/workers/publishing_api_notifier.rb
danielroseman: Also the update_type should not default to "normal", as that is not a valid value. Probably should be "minor".

test/unit/edition_test.rb
```diff
@@ -68,6 +68,16 @@ class EditionTest < ActiveSupport::TestCase
     publish(user, edition)
   end
 
+  should "notify the content store when updated" do
+    edition = FactoryGirl.create(:guide_edition, state: "ready")
+    PublishingAPIUpdateNotifier.expects(:perform_async).with(edition.id.to_s)
+    GdsApi::Panopticon::Registerer.any_instance.stubs(:register)
+
+    user = FactoryGirl.create(:user)
```
danielroseman: Why do we need to create a user here?

app/workers/publishing_api_notifier.rb
mgrassotti: Right, could it be "PublishingApiPublishNotifier" to be similar to the other one?

danielroseman: I'm not really a fan of the name "notifier". Both of these classes are doing more than notifying, they're updating content.

mgrassotti: :+1:

app/workers/publishing_api_notifier.rb
mgrassotti: :+1:

lib/republish_content.rb
```diff
@@ -1,7 +1,7 @@
 class RepublishContent
   def self.schedule_republishing
     Edition.published.each do |edition|
-      PublishingAPINotifier.perform_async(edition.id.to_s, "republish")
+      PublishingApiPublisher.perform_async(edition.id.to_s, "republish")
```
danielroseman: This no longer really does what is expected by "republish" - that is, re-send the content to publishing-api. We need to call both put_content and publish; I think unfortunately this means we will need a third worker class, PublishingAPIRepublisher, which does both actions, as calling both existing workers individually will lead to out-of-order problems.

elliotcm: I've now updated this.

elliotcm:
While adding the extra worker I noticed that the two existing ones were capitalising "API" differently.  I prefer API but as long as it's consistent.

test/unit/publishing_api_updater_test.rb
```diff
@@ -0,0 +1,28 @@
+require "test_helper"
+require "gds_api/test_helpers/publishing_api_v2"
+
+class PublishingAPIUpdaterTest < ActiveSupport::TestCase
+  include GdsApi::TestHelpers::PublishingApiV2
+
+  context ".perform(edition_id)" do
+    setup do
+      Edition.after_save.clear #avoids automated calls to PublishingAPIUpdater
```
tuzz: This is dangerous because the test compromises the implementation.

We should take an approach more like this:
https://github.com/alphagov/publisher/commit/a377f52b2a7c1f706818eefed36b0fc0430b7a14#diff-63598632bcbe0c01faf4cedced45e944R18

PR #445, opened 2016-02-03
Fix publishing API presenter for unpublished items

danielroseman:
Use updated_at value if public_updated_at is nil.

cc @jamiecobbett

### DIFF
--- a/test/unit/published_edition_presenter_test.rb
+++ b/test/unit/published_edition_presenter_test.rb
@@ -3,7 +3,7 @@ require "test_helper"
 class PublishedEditionPresenterTest < ActiveSupport::TestCase
   include GovukContentSchemaTestHelpers::TestUnit
 
-  context ".render_for_publishing_api" do
+  context ".render_for_publishing_api with a published document" do
     setup do
       artefact = FactoryGirl.create(:artefact,
         content_id: "fd4b7ea6-5e95-489e-ac73-0d8710e894d8",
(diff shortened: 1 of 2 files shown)


test/unit/published_edition_presenter_test.rb
```diff
@@ -74,5 +74,13 @@ class PublishedEditionPresenterTest < ActiveSupport::TestCase
...
       assert_equal 'major', output[:update_type]
     end
+
+    should 'use updated_at value if public_updated_at is nil' do
+      updated_at = 1.minute.ago
+      edition = FactoryGirl.create(:edition, state: "draft", updated_at: updated_at)
+      assert_nil edition.public_updated_at
+      output = PublishedEditionPresenter.new(edition).render_for_publishing_api
+      assert_equal edition.updated_at, output[:public_updated_at]
+    end
```
jamiecobbett: The schema specifies that `public_updated_at` must be populated. Maybe it's worth validating draft editions against the schema too?

jamiecobbett:
One minor comment, otherwise :+1:

test/unit/published_edition_presenter_test.rb
jamiecobbett: To clarify what I meant: it seems worth adding a line to validate a draft edition against the schema, because there may be other things that are invalid about draft editions. We're already doing it for published editions.

danielroseman:
Added validation against schema, and some logic and tests to fix the issues this revealed with empty topic tags.

PR #446, opened 2016-02-10
Ensure that Services is loaded.

danielroseman:
Speculative fix for `NameError: uninitialized constant PublishingAPIUpdater::Services`

### DIFF
--- a/app/workers/publishing_api_publisher.rb
+++ b/app/workers/publishing_api_publisher.rb
@@ -1,3 +1,5 @@
+require 'services'
+
 class PublishingAPIPublisher
   include Sidekiq::Worker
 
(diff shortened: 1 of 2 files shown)


bradwright:
@danielroseman shall we deploy this to Integration and see if it works?

tijmenb:
I believe I solved it like this previously: 

https://github.com/alphagov/collections-publisher/pull/149

danielroseman:
@tijmenb thanks, do you think that's the same issue though? Your PR there talks about an explicit circular dependency error message, we're just getting uninitialized constant.

tijmenb:
I think so. I could be mistaken but I think the circular dependency error is actually a masked uninitialized constant error. At the time I thought it was caused because collections-publisher has a `/app/services` directory.

PR #447, opened 2016-02-22
Discard publishing api draft when artefact is destroyed...

steventux:
https://trello.com/c/5SKdEz0Q/609-publisher-discard-draft-for-unpublished-editions

In the event of an artefact being destroyed the publishing-api needs to
be notified to discard the draft content item as the artefact slug may
be used again, eg. to draft a different type of edition.
The artefact would have a different content_id from the publishing api as the draft is not currently removed.
The above scenario had happened in production and had to be remedied with a data fix so this PR attempts to ensure drafts are discarded in these cases in future.

### DIFF
--- a/lib/enhancements/artefact.rb
+++ b/lib/enhancements/artefact.rb
@@ -1,0 +1,11 @@
+require "artefact"
+
+class Artefact
+  before_destroy :discard_publishing_api_draft
+
+private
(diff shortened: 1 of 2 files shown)


jamiecobbett:
Tested successfully in dev :+1:

PR #448, opened 2016-02-25
Use Plek to find Mapit

alexmuller:
This will make the app use the environment specific Mapit meaning that we don't have to override this file in alphagov-deployment.

It will also change Publisher so that it uses Mapit over HTTPS rather than HTTP.


jennyd:
This change will mean that we'll have to have mapit running in dev when we're working with the local authorities import. I think that's fine, but it's worth being explicit about it.

alexmuller:
Is that a change which this pull request will cause? The vhost for `mapit.integration.publishing.service.gov.uk` only allows connections from inside the environment so I'm not sure `mapit.preview.alphagov.co.uk` has worked any time recently.

PR #450, opened 2016-02-29
Avoid leap day failures in integration specs

h-lame:
On leap day `1.year.ago(Date.today)` and `1.year.since(Date.today)` land
on the 28th, not the 29th.  This means the tests that change the year
on the form to `Date.today.year` fail because they expect `Date.today`,
but actually get `Date.yesterday`.  We avoid this by recording a new
value for "today" based on calling `1.year.since` on the
`1.year.ago(Date.today)` value and then using that "today" to get the
`1.year.since` for the future value.


h-lame:
If we don't merge this today, we can probably leave it until 2020.

davidbasalla:
Makes sense to me! :+1:

PR #452, opened 2016-03-08
Test against v2 content schemas

tijmenb:
Currently this app tests against the v1 content schemas. v2 does not require a content_id to be set.

Part of: https://trello.com/c/nGta782z


tijmenb:
Note that `update_type` is also not needed, bit still allowed in the content schemas. I'm making a PR now to disallow including this, so that this can be test-driven removed.

PR #456, opened 2016-03-15
Update govuk_content_models & gds-api-adapters

tijmenb:
We are going to need the most recent versions of these gems in upcoming pull requests.

cc @MatMoore @jackscotti


jackscotti:
:+1:  thanks!

PR #457, opened 2016-03-15
Source topics & browse pages from publishing-api

tijmenb:
This app currently uses the content-api to show the list of tags (browse pages & topics) in the edit form. This changes the source for that information to be the publishing-api.

This change brings this application's display of the dropdown more in line with the content-tagger.

## Before

The dropdown is grouped and the elements are in the form of `Parent: Child`.

screen shot 2016-03-15 at 14 29 32
screen shot 2016-03-15 at 14 29 36

## After

Elements are in the form of `Parent / Child`.

screen shot 2016-03-17 at 09 44 43
screen shot 2016-03-17 at 09 44 45

https://trello.com/c/PZrBY8g9


tijmenb:
After a talk with Holly, we concluded that we should keep the grouping for the topics & browse pages in the select boxes. This to prevent confusion about selecting tags, because tagging to a top-level topic won't have an effect.

tijmenb:
I've updated the PR to 1) remove the top-level topics from the select box and 2) group them by top-level topic. It looks like a hack, which I've tried to justify by excessive commenting.

I've also renamed the class to `Linkables`. The idea of that class is that we could use the same interface in content-tagger, whitehall and other apps that do tagging (like [policy-publisher](https://github.com/alphagov/policy-publisher/blob/af14dd89239bc63771b754f589da43c812c1dfc0/app/helpers/data_for_select_helper.rb#L22-L26)). What do you think @alphagov/team-finding-things ?

MatMoore:
:+1:

Only thing that seems strange is using the name to work out what level a tag is at... is this functionality that would be useful to have in the publishing api long term? (E.g. give me all the tags with no parent or with X as a parent)

Is there anything stopping us from extracting the Linkables functionality into a gem now? Seems like it would be immediately useful for other migrations.

jackscotti:
looks good, but @MatMoore has raised a really good point. What's your view on it, Tijmen?

tijmenb:
@MatMoore 

Yes, very good point - it's very strange. 

The argument against adding this to the publishing-api right now is that the new taxonomy can be arbitrarily nested. Adding this to the publishing-api now would perhaps not be valuable long-term. The risk is that we build something into publishing-api that won't work with the new taxonomy. 

The thinking is that having a temporary localised hack here is better than having a possibly wrong API endpoint. What do you think?

I suspect that it's too early now to extract thing into a gem, because we don't precisely know what the new tagging interface will look like and what to support. Also, the hacky stuff with the `base_path` is only applicable to this app until we've finished https://trello.com/c/PwmEKOgN. I'd rather wait for all apps to have implemented this and then extract it than creating the wrong abstraction.

MatMoore:
Fair enough, as long as temporary :)

PR #458, opened 2016-03-16
Save links to publishing-api on publish

jackscotti:
Ticket: https://trello.com/c/Cx3K3y8I/544-publisher-migration-publisher-should-send-links-to-publishing-api-on-publish
Part of: https://trello.com/c/nGta782z/525-epic-migrate-publisher-to-the-new-tagging-infrastructure

Paired on this with @MatMoore.

cc: @tijmenb @mobaig @Davidslv @bishboria

### DIFF
--- a/app/presenters/edition_links_presenter.rb
+++ b/app/presenters/edition_links_presenter.rb
@@ -1,0 +1,39 @@
+class EditionLinksPresenter
+  def initialize(edition)
+    @edition = edition
+  end
+
+  def payload
+    all_paths = browse_pages + topics
+    if all_paths.empty?
+      content_ids_by_path = {}
+    else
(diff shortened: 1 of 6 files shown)


MatMoore:
This code assumes there is no need for a separate primary topic and additional topics. This will matter when we do the next step of the migration, i.e. making the frontend get this information from the publishing api. I'm going to do some investigation to see if this is actually the case.

tijmenb:
So far, we haven't found anything that depends on the `primary_topic`. @MatMoore has contacted Mark Hurrel, who may know more about `primary_topic`'s pedigree and intention.

In the meantime I'm happy to merge this PR, since we can always go back and add the `primary_topic` to the links hash.

PR #459, opened 2016-03-21
Send parent mainstream browse to publishing api

MatMoore:
When we started sending links to the publishing api, we missed
the parent link. This is used for the breadcrumb in mainstream browse and is
always the first tagged browse page.

https://trello.com/c/B8J0eOkm/561-send-parent-mainstream-browse-page-from-publisher


Davidslv:
:+1:

PR #461, opened 2016-03-22
Remove rake task to republish all tags

tijmenb:
We introduced this task to migrate the tags to the publishing-api.

However, the [rake task in panopticon](https://github.com/alphagov/panopticon/blob/ae5426fcebe457ad7d6fa6f9eb74666c1762ab94/lib/tasks/migrate_taggings.rake) is actually more suited for this task because it also sends the `organisation` taggings. Panopticon and Publisher share a database, so we can use the task. 

cc @MatMoore


bishboria:
:+1:

PR #462, opened 2016-03-23
Content Summary and Workflow History report

ikennaokpala:
This PR is aimed at producing a Content Summary and Workflow history report that will offer Content Delivery Managers  better insights into the flow of work through Publisher, 

This report will be listed and accessed via this [url](https://publisher.publishing.service.gov.uk/reports)

This CSV report comprises of the following columns:

- Content title
- Content slug
- Content URL
- Current status
- Format
- Current assignee
- *** Timestamps for each request type during the workflow ***

### DIFF
--- a/app/presenters/content_workflow_presenter.rb
+++ b/app/presenters/content_workflow_presenter.rb
@@ -1,0 +1,35 @@
+class ContentWorkflowPresenter < CSVPresenter
+  def initialize(scope = Edition.published)
+    super(scope)
+    self.column_headings = [
+      :content_title,
+      :content_slug,
+      :content_url,
+      :current_status,
+      :stage,
+      :format,
+      :current_assignee,
+      :created_at,
+    ]
+  end
+
+private
+
+  def build_csv(csv)
+    csv << column_headings.collect { |ch| ch.to_s.humanize }
+    scope.each do |item|
+      item.actions.each do |action|
+        csv << [
+          item.title,
+          item.slug,
+          "#{Plek.current.website_root}/#{item.slug}",
+          item.state,
+          action.request_type,
+          item.format,
+          item.assignee,
+          action.created_at.to_s(:db),
+        ]
+      end
+    end
+  end
+end
--- a/test/unit/content_workflow_presenter_test.rb
+++ b/test/unit/content_workflow_presenter_test.rb
@@ -1,0 +1,71 @@
+# encoding: utf-8
+require 'test_helper'
+
+class ContentWorkflowPresenterTest < ActiveSupport::TestCase
+  should "provide a CSV export of content workflow" do
+    action_1 = Action.new(
+      request_type: "create",
+      created_at: "2016-01-07 17:41:57"
+    )
+
+    action_2 = Action.new(
+      request_type: "request_review",
+      created_at: "2016-01-11 10:21:00"
+    )
+
+    action_3 = Action.new(
+      request_type: "send_fact_check",
+      created_at: "2016-01-17 12:11:33"
+    )
+
+    action_4 = Action.new(
+      request_type: "request_amendments",
+      created_at: "2016-01-20 17:41:28"
+    )
+
+    action_5 = Action.new(
+      request_type: "send_fact_check",
+      created_at: "2016-03-07 17:31:44"
+    )
+
+    transaction_edition = TransactionEdition.new(
+      title: "Register to vote (armed forces)",
+      slug: "register-to-vote-armed-forces",
+      state: "published",
+      assignee: "Ray Khan",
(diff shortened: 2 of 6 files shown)


app/presenters/content_workflow_presenter.rb
```diff
@@ -0,0 +1,35 @@
+class ContentWorkflowPresenter < CSVPresenter
+  def initialize(scope = Edition.not_in(state: ["archived"]).all)
+    super(scope)
```
tuzz: You don't technically need to pass `scope` in here as the default behaviour for `super` will pass all method arguments to the superclass.

app/presenters/content_workflow_presenter.rb
```diff
@@ -0,0 +1,35 @@
...
+      :current_assignee,
+      :created_at,
+    ]
+  end
+
+private
+
+  def build_csv(csv)
+    csv << column_headings.collect { |ch| ch.to_s.humanize }
+    scope.each do |item|
```
tuzz: You probably want to use `#find_each` as I'd imagine these scopes can be quite large.

lib/enhancements/edition.rb
```diff
@@ -43,6 +43,8 @@ def self.state_names
     )
   }
 
+  scope :published, -> { where(state: "published").desc(:created_at) }
```
tuzz: It looks like there's already a `published` scope in Edition. I wouldn't recommend redefining this with an unexpected ordering applied.

https://github.com/alphagov/govuk_content_models/blob/master/app/models/edition.rb#L38-L40

It'd probably be better to explicitly order in the report generator or add a scope with a different name.

test/unit/content_workflow_presenter_test.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    guide_edition = GuideEdition.new(
+      title: "The Queen's Awards for Enterprise",
+      slug: "queens-awards-for-enterprise",
+      state: "published",
+      assignee: "Constance Cerf",
+      actions: [action_1, action_2, action_3, action_4, action_5]
+    )
+
+    Edition.stubs(:published).returns([transaction_edition, guide_edition])
```
tuzz: Why do you need to stub this out here? Doesn't the data setup above cause `Edition.published` to return the right things?

tuzz:
There are an awful lot of commits for a relatively small set of changes. It'd be nicer to squash a lot of the ones where you changed method names that you'd previously introduced in a previous change. Otherwise, it's just unnecessary noise in `git log`.

Personally, I don't work on the Publisher app so it'd be best to have this reviewed by people who do (cc @boffbowsh). With my naive understanding of what you're trying to achieve, it looks to be along the right lines.

lib/enhancements/edition.rb
ikennaokpala: @tuzz the ordering comes as part of the requirement..

ikennaokpala:
@tuzz your review is much appreciated.. I am inclined towards more micro commits as it affords me the opportunity to define my thought process.. yes this appears small.. I believe micor commits opens up a lot more detail probably not in the short term

app/presenters/content_workflow_presenter.rb
ikennaokpala: @tuzz At the moment we are only searching for published content.. not too sure about the size.. however, it would be useful to batch things up.

test/unit/content_workflow_presenter_test.rb
ikennaokpala: @tuzz my intention here is to try and setup just what I need.. based on my chat with @steventux I will go through the govuk_content_models

lib/enhancements/edition.rb
steventux: I think what might avoid adding a duplicate scope to the `Edition` model might be something like:
```ruby
Edition.published.order(created_at: :desc)
```
as @tuzz says, use the existing `published` scope and modify it by chaining your own order clause to it.

test/unit/content_workflow_presenter_test.rb
steventux: I'm not sure any modification to the `Edition` model is needed, the scope as I understand it finds published editions and orders them, this can be achieved by chaining existing scopes and order clauses as mentioned above. This could be done here: https://github.com/alphagov/publisher/pull/462/files/78f0ea9a7805855f91c30245973be7598a34bb88#diff-ef36a8aacde58db76a4c63c23399c589R36
with
```ruby
Edition.published.order(created_at: :desc)
```
couldn't it?

ikennaokpala: @steventux I have removed the duplicate scope that I defined in the enhancements for `Edition`

PR #464, opened 2016-04-01
Create a report for all mainstream content that was published between 2 dates

mgrassotti:
To create the report:
```
publisher$ rake export_dates_report[<start_date> (format: YYYY-MM-DD), <end_date> (format: YYYY-MM-DD)]
```
The report is produced inside `/reports`

### DIFF
--- a/app/presenters/dates_report_presenter.rb
+++ b/app/presenters/dates_report_presenter.rb
@@ -1,0 +1,40 @@
+class DatesReportPresenter < CSVPresenter
+  def initialize(start_date, end_date)
+    start_date = start_date.beginning_of_day
+    end_date = end_date.end_of_day
+
+    scope = Edition.any_of({ state: "published" }, state: "archived")
+      .where(:updated_at.gte => start_date)
+      .where(:created_at.lte => end_date)
+      .order_by(created_at: 'asc')
+      .select { |item|
+        item.actions.select { |a|
+          a.request_type == "publish" &&
+            a.created_at >= start_date &&
+            a.created_at <= end_date
+        }.any?
+      }
+    super(scope)
+
+    self.column_headings = [
+      :created_at,
+      :title,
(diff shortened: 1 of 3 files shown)


fofr:
Hi @mgrassotti, could you update the commit to have a summary no longer than 50 chars? This will match our styleguide for commits and avoid the little `...` button appearing in Github. More details are here: https://github.com/alphagov/styleguides/blob/master/git.md#structure

boffbowsh:
The scope you're using doesn't capture all the content we're looking for. I've only just had chance to look at the data model properly, I think what we want is:

- All Editions
- That are archived or published
- That were created before the end date and updated after the start date
- That have an embedded action with a `request_type` of `publish` between the dates

There's no index into these actions, so the script will have to iterate the Edition scope and find all content with an action matching that criteria. Sorry I wasn't able to define this more clearly up front, hopefully it'll be a relatively simple change.

app/presenters/dates_report_presenter.rb
```diff
@@ -0,0 +1,40 @@
+class DatesReportPresenter < CSVPresenter
+  def initialize(start_date, end_date)
+    start_date = Date.parse(start_date).beginning_of_day
```
boffbowsh: You're double-parsing here - the rake task already parses a string into a DateTime so when you run the rake task you get 
```
TypeError: no implicit conversion of Date into String
/var/govuk/publisher/app/presenters/dates_report_presenter.rb:3:in `parse'
```

boffbowsh:
One small comment. Once you've fixed that, could you run an export and send it to @lutgendorff for her to review please?

app/presenters/dates_report_presenter.rb
mgrassotti: @boffbowsh sorry, fixed

PR #467, opened 2016-04-13
Fix a bug with setting the transaction done page promos

issyl0:
- It was reported that new done pages failed to save when adding the
  organ donation promo to them.

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -236,7 +236,13 @@ protected
         :video_summary,
         :caption_file,
       ]
-    else # answer_edition, completed_transaction_edition, help_page_edition
+    when :completed_transaction_edition
+      [
+        :body,
+        :promote_organ_donor_registration,
+        :organ_donor_registration_url,
+      ]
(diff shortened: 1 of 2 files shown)


benilovj:
Is it worth adding a test to cover the case that was broken?

issyl0:
Interestingly, there's already a test for whether you can fill in the boxes and it saves: https://github.com/alphagov/publisher/blob/master/test/integration/completed_transactions_create_edit_test.rb#L69, but it hasn't ever failed else we'd have noticed sooner, so I don't know what went wrong there. I'll look into it.

h-lame:
Looks like the test doesn't fail because that's checking the update action and I can't see that the update action uses the `permitted_params` (or does any strong parameters style work at all TBH - but that seems unlikely so I must not be tracing it through), whereas the create action does, but there's no similar test for promos on that.

issyl0:
Right, I've added an explicit page refresh to the existing integration test (linked above) so that it does actually check that, once saved and the page refreshed, the selections are still there. The AJAX was never reloading the page, so the test was continually passing.

PR #475, opened 2016-05-05
Fix GSS codes for some councils in existing Business Support editions

issyl0:
- Following on from frontend PR 948, we had to edit existing Business
  Support Editions so that they would refer to the correct GSS codes
  from the new data in MapIt, not the old ones that no longer worked.
  This migration does just that, replacing the old GSS codes in the
  edition's `area_gss_codes` array with the new ones. The councils
  edited are: Northumberland, Gateshead, East Hertfordshire and
  Stevenage.
- We had to skip callbacks that checked for archived Artefacts in
  Panopticon, and skip validations that said that already published
  editions couldn't be edited.


sihugh:
Looks good

PR #478, opened 2016-06-06
Use the format name as `document_type`

tijmenb:
Currently all documents from publisher are sent to the publishing-api with format `placeholder`. 

`format` is being deprecated and replaced by `document_type` and `schema_name` ([RFC 41](https://gov-uk.atlassian.net/wiki/display/GOVUK/RFC+41%3A+Separate+document+type+from+format)). This PR will make publisher use `Artefact#kind` (the format name in publisher) for the `document_type`. This allows us to determine the type of content from the content item.

Possible values for `document_type` / `kind` in publisher are: 

- answer
- business_support
- campaign
- completed_transaction
- guide
- help_page
- licence
- local_transaction
- place
- programme
- simple_smart_answer
- transaction
- video

https://github.com/alphagov/govuk_content_models/blob/master/app/models/artefact.rb#L67-L79

I've checked that these don't clash with existing formats in the publishing-api:

```
[3] pry(main)> ContentItem.where(document_type: %w[answer business_support campaign completed_transaction guide help_page licence local_transaction place programme simple_smart_answer transaction video]).count
=> 0
```


boffbowsh:
This LGTM, why is it marked as Don't Merge?

danielroseman:
👍

tijmenb:
@boffbowsh I wanted some extra eyes on this to make sure this was the right thing to do! Happy for this to get merged now.

PR #479, opened 2016-06-14
Simplify the task to send documents to Panopticon

tijmenb:
This app "registers" its documents with panopticon.

This commit removes the publishing-api interaction from the class, so that `PublishedSlugRegisterer` does only one thing (you can use `rake publishing_api:republish_content` to republish to the publishing-api more efficiently). The publishing-api sending had stopped working, because the client uses the V1 endpoint and does not use a bearer token.

Also removes a bunch of rake tasks that are seldom/never used to remove cognitive overload.


MatMoore:
👍  to both of these things 

Feel free to merge when green

PR #482, opened 2016-06-21
Reduce local services to only those in use

h-lame:
For: https://trello.com/c/bXJG1RU8/401-investigate-and-fix-data-discrepancies-3

We want to use the local_services.csv as an indicator of which LGSL codes are actually used by GOV.UK documents.  To this end we remove all rows from it that are not currently used by a local transaction and provide rake tasks to clean up the now "orphaned" instances that would be left in the DB.

There's also a db migration which should be run before the rake task to fix the LGSL code used in an archived edition.

See the individual commits for a more verbose description of why we're doing this.

### DIFF
--- a/lib/local_service_cleaner.rb
+++ b/lib/local_service_cleaner.rb
@@ -1,0 +1,46 @@
+require 'set'
+
+class LocalServiceCleaner
+  def initialize(input_data = LocalServiceImporter.fetch_data)
+    @input_data = input_data
+  end
+
+  def run
+    LocalService.all.each do |local_service|
+      print "Looking at #{local_service.lgsl_code} -"
+      if in_csv? local_service
+        puts " in csv"
+      elsif in_local_transaction? local_service
+        puts " in local transaction"
+      else
+        puts " removed"
+        local_service.destroy!
+      end
+    end
+  end
+
+  def lgsls_from_csv
+    @_lgsls_from_csv ||= fetch_lgsls_from_csv
+  end
+
+private
+
+  def in_csv?(local_service)
+    lgsls_from_csv.include? local_service.lgsl_code
+  end
+
+  def in_local_transaction?(local_service)
+    LocalTransactionEdition.where(lgsl_code: local_service.lgsl_code).any?
+  end
+
+  def fetch_lgsls_from_csv
+    output = Set.new
+    CSV.new(@input_data, headers: true).
+      each.with_object(output) do |row, lgsls|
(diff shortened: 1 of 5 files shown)


test/unit/lib/local_service_cleaner_test.rb
```diff
@@ -0,0 +1,49 @@
+require_relative '../../test_helper'
+
+class LocalServiceCleanerTest < ActiveSupport::TestCase
+
+  context "removing services we don't need anymore" do
+    setup do
+      data = "LGSL,Description,Providing Tier\n"
+      data += "1234,Finding Joy,all\n"
+      data += "5678,Escaping Sadness,county/unitary\n"
```
issyl0: 👏

lib/local_service_cleaner.rb
```diff
@@ -0,0 +1,48 @@
+require 'set'
+
+class LocalServiceCleaner
+  def initialize(input_data = LocalServiceImporter.fetch_data)
+    @input_data = input_data
+  end
+
+  DetectionResult = Struct.new(:local_service, :"in_csv?", :"in_local_transaction?")
```
issyl0: What does this do? 💭

h-lame: Prove that I'm a messy developer who doesn't tidy up vestigial tails of alternative versions I decided against showing to the world.  Will delete.

test/unit/lib/local_service_cleaner_test.rb
jennyd: Districts can't escape sadness? 😢

h-lame: Handled at the county level, innit?  More efficient at that scale.

lib/tasks/local_transactions.rake
```diff
@@ -9,6 +9,7 @@ namespace :local_transactions do
   task :fetch_and_clean => :environment do
     Rake::Task["local_transactions:fetch"].invoke
     Rake::Task["check_for_ghosts:remove"].invoke
+    Rake::Task["local_transactions:remove_old_services"]
```
davidbasalla: Should this be `Rake::Task["local_transactions:remove_old_services"].invoke` ?

h-lame: ugh, yes.  Good spot!

h-lame: Resolved via rebase.  Sorry!

PR #486, opened 2016-06-28
Use publishing api for tagging

Davidslv:
- Fixed tests to handle the new tagging tab ui
- Fixed bug where the success message was not displayed for tag update
- Split mainstream browse pages ui into parent/breadcrumb and browse
  pages.
- Coalesce parent and topics into topics.
- Don't call patch_links with content from the shared db any more
- Fix tagging to linkables for new template
- Added tests to see that we preserve links when we mutate
- Use copy from content-tagger

This has been a team effort from: @benhyland @tijmenb and @davidslv
Trello: https://trello.com/c/iafUZESJ/681-make-mainstream-publisher-read-and-write-from-publishing-api


Davidslv:
The reason of failure is because of us using class variables, and the ruby styleguide advices against it, although we have no intent of inheriting this class in the future, and we shouldn't be using instance variables just because they work, when in reality we want the behaviour of a class variable. https://github.com/bbatsov/ruby-style-guide#no-class-vars

lib/tagging/linkables.rb
```diff
@@ -0,0 +1,43 @@
+# Linkables
```
tijmenb: Perhaps it's best to copy the new version of `Linkables` from content-tagger so we don't have to use the class variables:

https://github.com/alphagov/content-tagger/blob/master/lib/linkables.rb

app/controllers/editions_controller.rb
```diff
@@ -8,6 +8,10 @@ class EditionsController < InheritedResources::Base
   after_filter :report_state_counts, :only => [:create, :duplicate, :progress, :destroy]
   before_filter :remove_blank_collections, only: [:create, :update]
 
+  def tagging_update_form
```
tijmenb: Is this better `private`?

lib/tagging/tagging_update_form.rb
```diff
@@ -0,0 +1,46 @@
+# Tagging update form
+#
+# This class is partially duplicated from content-tagger. Once we've added the
```
tijmenb: This comment is not true anymore because we won't be changing the Whitehall interface in this mission.

tijmenb:
Looking good! 👍 

Two points:

- When doing `patch_links` we're sending a `previous_version`. This will cause publishing-api to return an error when somebody else has updated the links in the between you opening the edit form and pressing save ([scenario in content-tagger](https://github.com/alphagov/content-tagger/blob/4788ac547fe1a3bcff43234697845623bec334f7/spec/features/tagging_content_spec.rb#L23-L32)). Could we make sure the user sees a nice error message?
- Would it be good to combine `stub_linkables` and `stub_no_links_for_all_content_ids`? Not sure, but it seems like they're always together.

app/views/shared/_tagging.html.erb
```diff
@@ -0,0 +1,104 @@
...
+
+<script type="text/javascript">
+  $(function () {
+    var $primaryMainstream = $('.js-primary-mainstream');
+    var $mainstreamAlertWarning = $('.mainstream.alert-warning');
+
+    $('.js-primary-mainstream, .js-mainstreams').on('change', function() {
+      $(this).addClass('interacted');
+
+      var mainstreams = $('.js-mainstreams :selected').map( function() {
```
tijmenb: Is it possible to use `.val()` here instead of `map()`?

app/views/shared/_tagging.html.erb
```diff
@@ -0,0 +1,104 @@
...
+        </div>
+      </div>
+    </div>
+
+    <hr/>
+    <%= f.submit 'Update tags' %>
+<% end %>
+
+
+<script type="text/javascript">
```
tijmenb: Perhaps we could put this into a separate JS file?

app/views/shared/_tagging.html.erb
Davidslv: That's not the same thing, `.val()` is just returning one value instead of all.
I want all of the selected values so I can check on them.

screen shot 2016-07-04 at 14 01 13

app/views/shared/_tagging.html.erb
Davidslv: We could. But all this logic is only for this section of the form.

lib/tagging/linkables.rb
Davidslv: Addressed.

app/views/shared/_tagging.html.erb
tijmenb: Will `$('.js-mainstreams').val()` return an array?

Davidslv: No

Davidslv:
@tijmenb Comments have been addressed.

app/views/shared/_tagging.html.erb
```diff
@@ -0,0 +1,100 @@
...
+    <div class="row">
+      <div class="col-md-7 edition-tags">
+
+        <div class="form-group" id="edition_browse_pages_input">
+          <%= f.label :mainstream_browse_pages, "Mainstream browse pages", class: 'control-label' %>
+          <%= f.select :mainstream_browse_pages,
+                       @linkables.mainstream_browse_pages,
+                       {},
+                       { multiple: true,
+                         class: 'js-mainstreams select2',
```
tijmenb: We don't need `js-mainstreams` and `js-primary-mainstream` anymore.

lib/tagging/linkables.rb
MatMoore: If Linkables is reusable can we move it to gds-api-adapters or somewhere shared?

lib/tagging/tagging_update_form.rb
MatMoore: Can we extract the form now? Is the only thing that differs the available link_types?

app/controllers/editions_controller.rb
```diff
@@ -109,6 +109,11 @@ def tagging
   def update_tagging
     Tagging::TaggingUpdateForm.new(params[:tagging_tagging_update_form]).publish!
     redirect_to :back, flash: { success: "Tags have been updated!" }
+  rescue GdsApi::HTTPConflict
+    redirect_to :back,
+    flash: {
+      danger: "Somebody changed the tags before you could. Your changes have not been saved."
```
MatMoore: Is `redirect_to :back` reliable? Shouldn't we send them to a fixed page?

Also I'm wondering if we can make this more friendly to the user - it's almost like this error is taunting them... what should the user do next if this happens?

lib/tagging/linkables.rb
tijmenb: What advantages does that have?

lib/tagging/linkables.rb
```diff
@@ -2,42 +2,66 @@
 #
 # Note that the functionality to fetch tags from the publishing-api and
 # transform them into a structure for a dropdown in a form is duplicated
-# across apps like content-tagger and Whitehall. Once we've migrated all apps
+# across apps like content-tagger and Whitehall.
+# Example: https://github.com/alphagov/content-tagger/blob/master/lib/linkables.rb
+# Once we've migrated all apps
 # to use `content_id`, we may move this functionality into a gem.
```
tijmenb: This comment is not accurate anymore.

app/views/shared/_tagging.html.erb
Davidslv: Indeed!

app/views/shared/_tagging.html.erb
jackscotti: Agree with Tijmen

app/views/shared/_tagging.html.erb
```diff
@@ -0,0 +1,100 @@
...
+            <a href="https://www.gov.uk/topic/animal-welfare/pets">Animal Welfare / Pets</a>.
+            Tagging a page with a topic will send out an
+            <a href="https://www.gov.uk/topic/business-tax/vat/email-signup">email alert to subscribers</a>.
+            Formerly called <i>specialist sectors</i>.
+          </p>
+        </div>
+      </div>
+    </div>
+
+    <hr/>
```
jackscotti: This can be just `<hr>` I believe: https://developer.mozilla.org/en/docs/Web/HTML/Element/hr

Davidslv: The slash indicates that the Element has no closing tag, it's a common practice. Just like you should do <br /> instead of <br>

app/views/shared/_tagging.html.erb
Davidslv: Content Tagger and Publisher are meant to have the same logic in the tagging form. This form has been heavily "inspired" on the way content tagger has been done. It's easier to come to this one file and know anything that is directly related, I'm not sure when someone will make this a shared resource between projects but I feel this should be directly in this file for the sake of the next person that picks this up.
I believe what you suggesting is creating another file and but bear in mind that application.js requires all the js files in the directory, that basically is loading files that are not needed to be loaded all the time.

app/views/shared/_tagging.html.erb
MatMoore: It's gone out of fashion with html5. The slashes are only necessary in XML and XHTML where everything needs to be balanced.

lib/tagging/linkables.rb
MatMoore: Reuse. ♻️  I'm just concerned we have a lot of repeated code in the various publishing apps that will never get cleaned up, and it all adds maintenance overhead.

tijmenb:
🎉 let's start testing this on integration!

PR #490, opened 2016-07-15
Make the local_services.csv only contain transactions we use

issyl0:
- `LocalTransactionEdition.where(state: "published").to_a.map(&:lgsl_code).sort`
  and
  https://github.com/jennyd/ldg-mappings/blob/master/source_data/govuk-local-transactions.csv
  helped us find Local Transactions on GOV.UK that we actually still
  use. We should tidy up and remove the ones we don't use from this
  file, so they don't get reimported into Local Links Manager as they
  don't need to be.

Trello: https://trello.com/c/G8GfEDcH/424-disable-services-in-local-links-manager-that-are-actually-licences.


sihugh:
Matches the referenced sources :+1:

PR #491, opened 2016-07-15
Add LGSL 107 as first step of replacing 230

sihugh:
We currently use LGSL 230 to refer to sheltered housing, however this is from the Scottish list and is not supported in England and Wales.  107 seems like a better bet for us, also because it refers to sheltered housing more generally, as opposed to just for older people.

As a first step, we'll add 107 as an enabled service.

There is other work that needs to be done as part of this, and this is coordinated via [this Trello card](https://trello.com/c/Zb42F6Jq/417-switch-local-transaction-for-apply-for-sheltered-housing-from-one-lgsl-to-another-3).  Once that work is done, we can retire 230.

### DIFF
--- a/data/local_services.csv
+++ b/data/local_services.csv
@@ -23,6 +23,7 @@ LGSL,Description,Providing Tier
 90,Find out about appealing against  a rehousing decision,district/unitary
 92,Find out about applying for a council home,district/unitary
 101,Find out about squatters and unauthorised occupants in council property,district/unitary
+107,Find out about sheltered and supported housing,all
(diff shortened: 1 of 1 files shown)


jennyd:
Is it safe to remove 230 at the same time as adding 107? or do we need to add 107, switch the local transaction over to use it, and then remove 230 afterwards?

sihugh:
Yep, agree with you as discussed.  I'll remove the LGSL 230 entry in a subsequent PR when we're ready

brenetic:
👍

PR #492, opened 2016-07-19
Retire LGSL 230 (sheltered housing) in favour of LGSL 107

sihugh:
LGSL 230 is being retired in favour of LGSL 107 which was added in a previous commit.

We're waiting until we've done the rest of the shuffling steps in https://trello.com/c/Zb42F6Jq/417-switch-local-transaction-for-apply-for-sheltered-housing-from-one-lgsl-to-another-3 before removing 230 from the CSV.  This means that we will no longer import URLs for LGSL 230 from Local Direct Gov. 

We'll then run the migration to change the LGSL code on the LocalTransactions from 230 to 107 (which we are still importing)

We'll then run the `local_transactions:remove_old_services` task to clean up


brenetic:
🍔

PR #493, opened 2016-08-05
Stop importing from local directgov admin

h-lame:
For: https://trello.com/c/opAHOMr4/443-4-of-4-delete-old-ldg-import-into-publisher-and-related-code-3

This removes the imports for local authorities and local interactions as the data for these comes from LDG Admin and we have a new source of truth in [local links manager](https://github.com/alphagov/local-links-manager).   We also include the migrations that remove local authorities and local interactions entirely from the db (see: alphagov/govuk_content_models#389 and alphagov/govuk_content_models#391).


lib/tasks/local_transactions.rake
```diff
@@ -1,27 +1,11 @@
 namespace :local_transactions do
 
-  desc "Download and import services, interactions and contacts"
-  task :fetch => :environment do
-    LocalAuthorityDataImporter.update_all
-  end
-
   desc "Download and import services, interactions and contacts, and subsequently remove 'ghost' interactions"
```
sihugh: Might need to remove ", interactions and contacts, and subsequently remove 'ghost' interactions"

h-lame: Good spot.  Done and rebased.

sihugh:
Deploying Publisher should stop the import schedule as we've now removed it from [alphagov-deployment](https://github.gds/gds/alphagov-deployment/blob/master/publisher/to_upload/config/production/schedule.rb), which is fine.  We'll have to remember to run it manually if we change the services.csv.

sihugh:
Looks good to me, merge when ready 👍

h-lame:
Removed Discuss flag as we've talked about it and decided we can also remove local authorities.  Flagged as DNM as we want to bump govuk_content_models before merging + deploying.

h-lame:
Removed DNM flag as we've bumped the dependencies now that 39.0.0 of content models is released.

PR #498, opened 2016-08-22
Stop relying on slugs for areas in imminence API responses

h-lame:
For: https://trello.com/c/PKAohBzb/468-stop-adding-slugs-to-areas-returned-from-mapit-by-imminence-2

Imminence will shortly stop providing slugs for areas in it's api
responses.  Happily we don't rely on them, but we update our test
fixtures so that it doesn't look to future developers that slugs are
valid bits of data.

See https://github.com/alphagov/imminence/pull/136 for the imminence change


emmabeynon:
👍

PR #499, opened 2016-08-24
Add 'kind' column to generated report

kalleth:
[Trello card](https://trello.com/c/MFIzBBFZ/23-add-new-fields-to-publisher-export)

Titled as 'format' because that's what's shown in panopticon. We want to make this information available in the publisher reports so that it's still available when panopticon is retired.

TODO:

- [x] Deploy this branch to integration and run the rake task to generate these reports. I haven't been able to do this locally due to the hilarious number of app dependencies involved.


NeilvB:
Looks good 👍

PR #500, opened 2016-09-01
Remove unused collections in shared database

sihugh:
These collections are old and redundant so we'll clean them up.

They last had documents added in Feb 2012 at the latest and we can't find any reference of them being used now.

[Trello card](https://trello.com/c/2NFVxE0F/472-clean-up-old-unused-collections-in-shared-database-3-2)


brenetic:
👍

PR #501, opened 2016-09-13
Filter out Linkables with nil titles (eg. redirects)

steventux:
https://errbit.publishing.service.gov.uk/apps/5304e0520da11511b0000011/problems/57d7ed6c6578636cf0a40100

Publisher is erroring due to a redirected browse page which now has a nil title. So filter out items with nil titles.


test/support/tag_test_helpers.rb
```diff
@@ -11,6 +11,7 @@ def stub_linkables
       { base_path: '/browse/tax/vat', internal_name: 'Tax / VAT', publication_state: 'published', content_id: 'CONTENT-ID-VAT' },
       { base_path: '/browse/tax/capital-gains', internal_name: 'Tax / Capital Gains Tax', publication_state: 'published', content_id: 'CONTENT-ID-CAPITAL' },
       { base_path: '/browse/tax/rti', internal_name: 'Tax / RTI', publication_state: 'draft', content_id: 'CONTENT-ID-RTI' },
+      { base_path: '/browse/tax/nil', internal_name: nil, publication_state: 'draft', content_id: 'CONTENT-ID-RTI' },
```
kevindew: probably should switch this content id to be different to line 13

PR #502, opened 2016-09-16
Add functionality to manage external links

jackscotti:
We are planning to move this functionality from Panopticon to Publisher given
that only items published by this application can have related external links.

How it will look like:
screen shot 2016-09-16 at 14 57 03

https://trello.com/c/BgdAzNQl/118-set-related-external-links-in-mainstream-publisher


app/assets/stylesheets/forms.scss
```diff
@@ -20,3 +20,34 @@
...
+  input[type=text] {
+    margin-bottom: 0px;
+  }
+
+  .fields:last-of-type {
+    margin-bottom: 10px;
+  }
+}
+
+.nested-item {
```
MatMoore: I'm not sure how much publisher shares styles with other apps in general, but is there a shared admin component we can put these styles into?

app/controllers/editions_controller.rb
```diff
@@ -116,6 +119,20 @@ def update_tagging
...
+  def update_related_external_links
+    artefact = resource.artefact
+    external_links = params.require(:artefact).permit(external_links_attributes: [:title, :url, :id, :_destroy])
+
+    artefact.external_links_attributes = external_links["external_links_attributes"]
+
+    if artefact.save
+      flash[:success] = "External links have been saved."
+    else
+      flash[:danger] = "Not saved. Links should begin with 'http'."
```
MatMoore: Should the flash be displaying something from the `errors` of the artefact here? There could be other reasons why the artefact isn't valid.

app/helpers/tabs_helper.rb
```diff
@@ -32,7 +33,14 @@ def self.[](name)
       end
 
       def title
-        name == 'history' ? 'History and notes' : name.capitalize
+        case name
+        when 'history'
+          return 'History and notes'
```
MatMoore: I think you can drop the "return" here

app/controllers/editions_controller.rb
jackscotti: Gave a look into this and right now, independently from the error, the error message would be something like "Artefact external link is invalid". 

I could try to return a customised error here:
https://github.com/alphagov/govuk_content_models/blob/master/app/models/artefact_external_link.rb

What do you suggest?

jackscotti: I decided to return the original messages and stop testing for their content as the error messages body should only be tested on the model in govuk_content_models.

app/assets/stylesheets/forms.scss
jackscotti: I don't know, is it?

jackscotti:
@MatMoore addressed all your comments but the one about shared stylesheet as I know nothing about it.

MatMoore:
Me neither unfortunately, might be worth checking in with a front end dev but 👍  from me.

rubenarakelyan:
👍

PR #503, opened 2016-09-21
Send external related links to publishing api

jackscotti:
The final objective is storing and retrieving them from the content-store.

This PR needs to be merged first: https://github.com/alphagov/govuk-content-schemas/pull/392

Ticket: https://trello.com/c/rDKC4sLm/133-send-related-external-links-to-pub-api-from-publisher
Part of: https://trello.com/c/mOeDK914/107-epic-related-links-and-panopticon

### DIFF
--- a/app/presenters/published_edition_presenter.rb
+++ b/app/presenters/published_edition_presenter.rb
@@ -28,12 +28,22 @@ class PublishedEditionPresenter
           additional_topics: @edition.additional_topics,
           topics: (primary_topic + @edition.additional_topics)
         },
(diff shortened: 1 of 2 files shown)


tijmenb:
> The final objective is storing and retrieving them from the content-store.

We only have to send it to the publishing-api (like you're doing here). There's no need to read from it because other apps won't edit the details hash of these items.

When we start rendering this in frontend we will need need to add a note to the new tab that explains the user needs to republish the document to make the changes visible.

PR #504, opened 2016-09-23
Send documents to Rummager directly

mobaig:
Previously this app would send documents to Panopticon in order to have
them added to the search index. This flow is now being simplified so
that all apps send documents to Rummager directly.

This commit adds:

- a presenter to form the payload required by Rummager. The payload now
  includes the following extra attributes: publishing_app, rendering_app
  and content_id.
- an indexer service object to handle the actual sending.
- updates across codebase and test suite wherever interaction with
  panopticon takes place.

search-admin result before:

screen shot 2016-09-22 at 15 46 33

search-admin result after:

screen shot 2016-09-22 at 15 47 48

### DIFF
--- a/app/presenters/search_payload_presenter.rb
+++ b/app/presenters/search_payload_presenter.rb
@@ -1,0 +1,33 @@
+class SearchPayloadPresenter
+  attr_reader :registerable_edition
+  delegate :slug,
+           :title,
+           :description,
+           :indexable_content,
+           :public_timestamp,
+           :artefact,
+           :format,
+           to: :registerable_edition
+
+  def initialize(registerable_edition)
+    @registerable_edition = registerable_edition
+  end
+
+  def self.present(registerable_edition)
+    new(registerable_edition).present
+  end
+
+  def present
+    {
+      content_id: artefact.content_id,
+      rendering_app: "publisher",
+      publishing_app: "publisher",
+      format: format.underscore,
+      title: title,
+      description: description,
+      indexable_content: indexable_content,
+      link: "/#{slug}",
+      public_timestamp: public_timestamp,
+    }
+  end
+end
--- a/lib/published_slug_registerer.rb
+++ b/lib/published_slug_registerer.rb
@@ -1,7 +1,7 @@
 require 'gds_api/panopticon'
+require 'gds_api/rummager'
 
-# Registers slugs for published editions with Panopticon (which in turn updates
-# rummager/search).
+# Registers slugs for published editions with Panopticon
 class PublishedSlugRegisterer
   attr_reader :logger
 
@@ -10,7 +10,7 @@ class PublishedSlugRegisterer
     @slugs = slugs.sort.uniq
   end
 
-  def run
+  def do_panopticon
     @success_slugs = []
     @not_found_slugs = []
     @errored_slugs = []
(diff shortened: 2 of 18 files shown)


app/presenters/search_payload_presenter.rb
```diff
@@ -0,0 +1,36 @@
...
+      content_id: artefact.content_id,
+      rendering_app: "publisher",
+      publishing_app: "publisher",
+      format: format.underscore,
+      title: title,
+      description: description,
+      indexable_content: indexable_content,
+      link: "/#{slug}",
+      public_timestamp: public_timestamp,
+      specialist_sectors: registerable_edition.specialist_sectors.compact,
```
tijmenb: We don't have to send `specialist_sectors`, as rummager will fetch them from `publishing-api` when indexing.

lib/published_slug_registerer.rb
```diff
@@ -72,10 +87,11 @@ def register(slug)
     end
   end
 
-  def register_with_panopticon(edition)
+  def register_externally(edition, service_name:)
     retry_count = 0
+    registration_method = "register_with_#{service_name}"
     begin
-      edition.register_with_panopticon
```
tijmenb: Would it be better to duplicate this method to avoid the metaprogramming?

test/test_helper.rb
```diff
@@ -67,6 +67,7 @@ def without_metadata_denormalisation(*klasses, &block)
 
   def stub_register_published_content
     WebMock.stub_request(:put, %r{\A#{PANOPTICON_ENDPOINT}/artefacts/})
+    WebMock.stub_request(:post, %r{search.dev.gov.uk/documents})
```
tijmenb: This is weird, I think Plek should make us use `search.test.gov.uk` in tests. Are we setting it to `dev.gov.uk` somewhere?

app/presenters/search_payload_presenter.rb
```diff
@@ -0,0 +1,36 @@
...
+      sections: sections,
```
tijmenb: This is interesting. `sections` are mainstream browse pages, so we should've started indexing these from publishing-api just like `topics` and `specialist_sectors`. 

But we don't: https://github.com/alphagov/rummager/blob/master/lib/indexer/links_lookup.rb

Where did you find `sections` before?

test/unit/search_indexer_test.rb
```diff
@@ -0,0 +1,35 @@
...
+    Services.rummager.expects(:add_document).with(
+      'edition',
+      "/#{edition.slug}",
+      content_id: "content-id",
+      rendering_app: "publisher",
+      publishing_app: "publisher",
+      format: "guide",
+      title: "A title",
+      description: "An overview",
+      indexable_content: "",
```
tijmenb: Can you add a test that shows we send something sensible to `indexable_content`?

app/presenters/search_payload_presenter.rb
mobaig: As defined by https://github.com/alphagov/publisher/blob/master/lib/registerable_edition.rb#L40, and picked up as an optional param here: https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/panopticon/registerer.rb#L38.

This basically doesn't need to be sent, for the same reason as the specialist sectors - rummager will fetch the links for a published doc.

lib/published_slug_registerer.rb
mobaig: have reworked this class with some duplication and blocks. see f1dea54. WDYT?

tijmenb: Excellent. Code duplication ❤️

test/test_helper.rb
mobaig: In hindsight this does seem strange, but this changelog entry would suggest the 'test.gov.uk' behaviour is no longer present: https://github.com/alphagov/plek/blob/e57c986c8163257267197590ec0211f9ddb1da46/CHANGELOG.md#to-upgrade

test/unit/search_indexer_test.rb
mobaig: yep, oversight - will add.

tijmenb:
Two more things: 

- panopticon does not index pages with format `business_support`  and `completed_transaction` ([here](https://github.com/alphagov/panopticon/blob/a753b31fa620da6fe255e7b0530dac5f5123160c/app/models/rummageable_artefact.rb#L3)). Unfortunately there are three business support schemes that are "exceptional" and are indexed ([code](https://github.com/alphagov/panopticon/blob/a753b31fa620da6fe255e7b0530dac5f5123160c/app/models/rummageable_artefact.rb#L8-L12), https://github.com/alphagov/panopticon/pull/80). But I'm thinking that it would be good to not migrate this code, because exceptions are horrible.

- We need a way to remove things from search. I think this is currently done when the editor withdraws an artefact, which is done in panopticon. Maybe we should add it there temporarily (we'd have to re-add parts of https://github.com/alphagov/panopticon/pull/398).

mobaig:
@tijmenb see latest sha and https://github.com/alphagov/panopticon/pull/399

carvil:
@tijmenb I've merged this one

PR #507, opened 2016-09-29
Allow filtering by format in the edition list index

h-lame:
While doing discovery on migrating publisher formats to the new
world of publishing-api it became clear it'd be nice to filter the
index to show only one specific format.  Mostly for when we want
to find an example of a format.  This might be useful for others.

Only poked this a bit on dev, not sure if it needs to be more robust in the face of other controllers that share this functionality.

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -30,13 +30,22 @@ class RootController < ApplicationController
 
 private
 
+  def format_filter
+    Artefact::FORMATS_BY_DEFAULT_OWNING_APP["publisher"].include?(params[:format_filter]) ? params[:format_filter] : 'edition'
+  end
+
+  def filtered_editions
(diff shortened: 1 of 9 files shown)


davidbasalla:
I like it! 👍  What do you mean with that it might have to be more robust? Which controllers are you concerned about?

h-lame:
Eh. I dunno TBH, I'll check if the helpers and views used to build the sidebar are used elsewhere or not.  I don't think they are, but I should be sure.

binaryberry:
Feel free to merge this when you're happy @h-lame ? We've been deploying this branch to integration, it's definitely useful.

h-lame:
Just adding it to the (probably soon to be deprecated) user search page.  I'll request another eyeballing and then merge.  Glad to hear it's useful though!

PR #508, opened 2016-09-30
Remove database.yml and schema.rb

davidbasalla:
Publisher does not appear to be using sqlite (and ActiveRecord) as its own local database so we remove the relevant files. These files were last edited in 2011/2012.


Davidslv:
Indeed ✏️ 👍

PR #510, opened 2016-10-04
Bugfix: do not crash when trying to save an unmodified item

jackscotti:
This makes impossible to save without having added new external related links. 

Different approach the one proposed that was hiding/showing the save button based on the addition of external links (with JS).

screen shot 2016-10-05 at 16 32 11

This works updates the copy of the buttons too. Wasn't very clear.


rubenarakelyan:
👍

PR #512, opened 2016-10-11
Really use GdsApi.config.always_raise_for_not_found

sihugh:
The gds_api.rb initializer is overwritten on deploy by [files from alphagov-deployment](https://github.gds/gds/alphagov-deployment/blob/master/publisher/initializers_by_organisation/integration/gds_api.rb)

These files in alphagov-deployment don't have the `always_raise_for_not_found`
configuration set.  This means that it's only used in dev.  I recently tried
to create a new `Artefact` in Panopticon in dev (which worked), but I was
subsequently unable to edit the created `Edition` in Publisher as Publishing-api
returned a 404 for `Services.publishing_api.get_links` on it.

Production environments are currently unaffected, but if we fix the
alphagov-deployment files to also have this configuration before this fix is
merged we'll be unable to edit `Editions` on newly created `Artefacts`.

The production initializers also try to use environment variables, but this is
not backed up by Puppet or the deployment repo, so this is something that we'll
fix in the near future.

[Trello story](https://trello.com/c/v7pdphVZ/83-publisher-api-adapters-upgrade-is-incomplete)

### DIFF
--- a/config/initializers/gds_api.rb
+++ b/config/initializers/gds_api.rb
@@ -1,3 +1,6 @@
+# This file is overwritten on deploy (see alphagov-deployment), so be careful
+# about relying on this content
+
 require 'gds_api/base'
 require 'gds_api/asset_manager'
 require 'attachable'
@@ -16,5 +19,3 @@ PANOPTICON_API_CREDENTIALS = {
 Attachable.asset_api_client = GdsApi::AssetManager.new(Plek.current.find('asset-manager'), {
   bearer_token: ENV['PUBLISHER_ASSET_MANAGER_CLIENT_BEARER_TOKEN'] || 'also_set_at_deploy_time',
 })
-
-GdsApi.config.always_raise_for_not_found = true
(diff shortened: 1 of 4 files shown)


config/initializers/gds_api.rb
```diff
@@ -16,5 +16,3 @@
 Attachable.asset_api_client = GdsApi::AssetManager.new(Plek.current.find('asset-manager'), {
   bearer_token: ENV['PUBLISHER_ASSET_MANAGER_CLIENT_BEARER_TOKEN'] || 'also_set_at_deploy_time',
 })
```
mobaig: Think it might be worth making it clearer that this file is overwritten on deploy. Missed this detail from the existing comment here. Perhaps a separate, more specific comment right at the top.

mobaig:
Thanks for the fix 👍 . Looks good, barring a comment about a comment.

config/initializers/gds_api.rb
sihugh: Ta, we're going to fix this, but will add comment. :+1:

sihugh: I've added the comment to the top now

PR #514, opened 2016-10-20
Skip review button should be available permitted authors of content

steventux:
https://trello.com/c/r7YfHVzG/282-add-force-publish-to-mainstream-medium
The original story for this feature mentioned placing the button within
the review buttons visible when viewing another editor's work when in
review. The skip review feature should mean that an author with the correct
permissions can skip the review stage of their own work which means the
relevant button needs to appear when the user is viewing their own work.

screenshot from 2016-10-20 12 10 51

screenshot from 2016-10-20 12 11 47


keithiopia:
👍

PR #516, opened 2016-11-03
Remove old tagging code

tijmenb:
This commit removes all code related to the old way of tagging. We no longer need tags and related links in the content-api, so we can remove everything from the content models, panopticon and this application.

Tagging is now 100% done via the new publishing platform.

- [x] Only to be merged with a bumped version of https://github.com/alphagov/govuk_content_models/pull/402.

https://trello.com/c/KfgeIRtP


mobaig:
🔥 🔥 🔥

PR #519, opened 2016-12-09
Fix the needs Rake task authenticating with the Publishing API

issyl0:
- We were instantitating a new one without a token and it was working
  fine in dev, but it failed on Integration, so use the already set up
  `Services.publishing_api` for it, adjusting the timeout option as before.


tijmenb:
You'll need to add the new link type to the schemas too. Links are actually validated in the publishing-api, but just raise errors to errbit at the moment.

issyl0:
@tijmenb Thanks. Done in https://github.com/alphagov/govuk-content-schemas/pull/457.

PR #520, opened 2016-12-15
Send the right routes to the publishing API

brenetic:
- Send the right routes to the publishing api.

[Trello card](https://trello.com/c/sZyeX2Ul/570-preparatory-work-fix-content-item-routes-for-publisher-formats-3)


sihugh:
👍

PR #521, opened 2016-12-19
Set locale correctly in content item

brenetic:
Language is set on the artefact by Panopticon, and we should use it for
the content item instead of saying everything is English.

https://trello.com/c/KoJ833G9/591-fix-locale-welsh-publisher-content-in-publishing-api


davidbasalla:
Seems fair!

PR #525, opened 2017-01-06
Clarify when external related links are published

tijmenb:
When the related external links were first moved from panopticon into this application, the sidebars on the site were populated from the content-api. This has been changed to the content store.

This means that the "save" action on the related external links tab will no longer have an immediate effect. Instead, the user has to republish the edition to make it visible on the site.

This adds two notes to the UI to make this clear.

https://trello.com/c/71hfUPZA


rubenarakelyan:
👍

PR #526, opened 2017-01-09
Set locale correctly in content item

davidbasalla:
Language is set on the artefact by Panopticon, and we should use it for
the content item instead of saying everything is English.

https://trello.com/c/KoJ833G9/591-fix-locale-welsh-publisher-content-in-publishing-api

Coincides with https://github.com/alphagov/publishing-api/pull/717


whoojemaflip:
seems legit. :1st_place_medal:

PR #527, opened 2017-01-12
Set locale when publishing with PublishingAPI

davidbasalla:
This is necessary when publishing Welsh content (ie with locale 'cy') via the PublishingAPI. The publish command defaults to using the 'en' locale so we set it here explicitely

Tested successfully on Integration with Welsh and English content


emmabeynon:
:sheep:

PR #528, opened 2017-01-12
Add Unpublish tab to the edition editor

emmabeynon:
This adds unpublish (fomerly know as 'withdraw') functionality to Publisher.  This functionality is copied from Panopticon, where it is soon to be deleted.

It's 'old world' publishing platform, in that it expects to be managing routes and
Rummager content itself, rather than letting the publishing-api manage them.

When we move the mainstream formats to a new Schema (soon!) we will need to revisit
this to stop it from speaking to the router-api and rummager directly.

Trello: https://trello.com/c/xM4HRRYt/574-preparatory-work-moving-withdrawing-functionality-into-publisher-5

image

Paired with @whoojemaflip

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -26,6 +26,7 @@ class EditionsController < InheritedResources::Base
   alias_method :metadata, :show
   alias_method :history, :show
   alias_method :admin, :show
+  alias_method :unpublish, :show
 
   def new
     @publication = build_resource
@@ -181,6 +182,22 @@ class EditionsController < InheritedResources::Base
     @comparison = @resource.previous_siblings.last
   end
 
+  def process_unpublish
+    edition = Edition.find(params[:id])
+    artefact = edition.artefact
+    success = UnpublishService.call(artefact, current_user, redirect_url)
+
+    if success
+      notice = "Content unpublished"
+      notice << " and redirected" if redirect_url.present?
+      flash[:notice] = notice
+      redirect_to root_path
+    else
+      flash[:alert] = artefact.errors.full_messages.join("\n")
+      redirect_to unpublish_edition_path(edition)
+    end
+  end
+
 protected
 
   def permitted_params(subtype: nil)
@@ -330,6 +347,14 @@ protected
   end
 
 private
+
+  def redirect_url
+    make_govuk_url_relative params["redirect_url"]
+  end
+
+  def make_govuk_url_relative(url = "")
+    url.sub(%r{^https?://(www\.)?gov\.uk/}, "/")
+  end
 
   def tagging_update_form
     Tagging::TaggingUpdateForm.build_from_publishing_api(@resource.artefact.content_id)
--- a/app/models/routable_artefact.rb
+++ b/app/models/routable_artefact.rb
@@ -1,0 +1,87 @@
+require 'plek'
+require 'gds_api/router'
+
+class RoutableArtefact
+  attr_reader :artefact
+
+  def initialize(artefact)
+    @artefact = artefact
+  end
+
+  def logger
+    Rails.logger
+  end
+
+  def router_api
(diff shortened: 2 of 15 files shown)


davidbasalla:
What happens when the user tries to unpublish an already unpublished artefact? Is there a warning message of some sort (sorry Integration is busy at the moment, can't test it myself yet). Should we add an integration test for that?

app/models/routable_artefact.rb
```diff
@@ -0,0 +1,87 @@
+require 'plek'
+require 'gds_api/router'
+
+class RoutableArtefact
```
davidbasalla: This class currently does a lot more than it needs to. But I guess it's a straight lift from Panopticon and will be used for the creation of content as well, so probably fine!

emmabeynon: That's right - it's straight from Panopticon, and we're expecting to remove it at a later date.

emmabeynon:
@davidbasalla Re: unpublishing an already unpublished artefact, we do cover that in lines 38-43 of `unpublish_test.rb`.  Do you think it would be clearer to split it out into another test?

davidbasalla:
@emmabeynon I guess I was more thinking about what happens when unpublishing fails? Looks like a flash message gets set -> https://github.com/alphagov/publisher/pull/528/files#diff-3d0014943f4437aab728020d24b993ebR195 We could add an integration test for that?

app/controllers/editions_controller.rb
```diff
@@ -331,6 +347,10 @@ def description(r)
 
 private
 
+  def redirect_url
+    params.fetch("redirect_url", "").sub(%r{^https?://(www\.)?gov\.uk/}, "/")
```
davidbasalla: Apols if it's obvious, but what does this do? Still haven't memorised my regex patterns :/ Would there be a benefit of putting this in an intention revealing named method?

app/controllers/editions_controller.rb
```diff
@@ -181,6 +182,21 @@ def diff
     @comparison = @resource.previous_siblings.last
   end
 
+  def process_unpublish
+    edition = Edition.find(params[:id])
+    success = UnpublishService.call(edition.artefact, current_user, redirect_url)
```
davidbasalla: do we need to cater for the fact that edition could be `nil`?

app/models/routable_artefact.rb
davidbasalla: Okay! I'd probably be in favour of only copying over what's strictly needed as it would make it easier to read/review and remove any unnecessary code. But no reason to hold up this PR!

app/services/unpublish_service.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    def add_gone_route_in_router_api(artefact)
+      RoutableArtefact.new(artefact).submit
+    end
+
+    def unpublish_in_publishing_api(artefact)
+      Services.publishing_api.unpublish(
+        artefact.content_id,
+        type: 'gone',
```
davidbasalla: shouldn't this be `redirect` for when we do a redirect?

emmabeynon:
@davidbasalla Now added a test in for redirecting to an external link.  This involved a change to setting the flash alert in the editions controller, which should also address your point about what happens if unpublishing fails.

app/services/unpublish_service.rb
emmabeynon: Because this is 'old world' rather than 'new world', Publishing API won't deal with redirects due to the 'placeholder' type we're working with.

app/controllers/editions_controller.rb
emmabeynon: This has now been broken out in to two methods, which hopefully should be a bit clearer 👍

app/services/unpublish_service.rb
davidbasalla: As discussed IRL, might be better to entirely remove this or do both Gone and Redirects

PR #529, opened 2017-01-13
Improve UI for edition metadata

Davidslv:
The current user journey to add or edit related links was sending users
through Panopticon and then they would see the functionality had moved
to Content Tagger.

This commit aims to be easier for users to identify where they need to
go and send them directly to the place they need.

## How it looks
screen shot 2017-01-13 at 15 28 35

Trello: https://trello.com/c/1PBuV84e/418-link-to-content-tagger-from-mainstream-publisher

### DIFF
--- a/app/views/shared/_metadata.html.erb
+++ b/app/views/shared/_metadata.html.erb
@@ -6,4 +6,9 @@
     <% end %>
   </div>
 </div>
-<%= link_to "Edit in Panopticon", panopticon_edit_url(publication), :class => "btn btn-primary" %>
+
+<ul class="list-unstyled">
+  <li><%= link_to "Add or edit related GOV.UK links", content_tagger_url(publication)%></li>
+  <li><%= link_to "Add user need or mark Welsh content", panopticon_edit_url(publication)%></li>
+  <li><%= link_to "Withdraw and redirect content", panopticon_withdraw_url(publication)%></li>
+</ul>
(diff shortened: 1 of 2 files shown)


Davidslv:
Commit has been updated to reflect decisions made on the text to be used.

fofr:
Do they need to be buttons?

Davidslv:
@fofr I can change them to be links in an unordered list
What do you prefer?

screen shot 2017-01-13 at 15 14 00

fofr:
I think the list of links is clearer :+1:

I don't think the bullets are necessary, you can use the class `list-unstyled` to remove them if you prefer.

Davidslv:
Thank you @fofr, I've addressed your comment

app/views/shared/_metadata.html.erb
```diff
@@ -6,4 +6,9 @@
     <% end %>
   </div>
 </div>
-<%= link_to "Edit in Panopticon", panopticon_edit_url(publication), :class => "btn btn-primary" %>
+
+<ul class="list-unstyled">
+  <li><%= link_to "Add or edit related GOV.UK links", content_tagger_url(publication)%></li>
+  <li><%= link_to "Add Maslow need or mark Welsh content", panopticon_edit_url(publication)%></li>
```
tijmenb: I think Andy suggested "Add user need" for this.

PR #530, opened 2017-01-13
Add the ability to create and edit Artefacts

whoojemaflip:
This is functionality that is being deprecated from Panopticon.

It adds a new task to the main menu 'Add Artefact', as well as providing
a view to edit (draft) artefact parameters.

screen shot 2017-01-19 at 14 48 35

screen shot 2017-01-13 at 15 46 42

screen shot 2017-01-13 at 15 46 53

### DIFF
--- a/app/controllers/artefacts_controller.rb
+++ b/app/controllers/artefacts_controller.rb
@@ -1,0 +1,46 @@
+class ArtefactsController < ApplicationController
+  def new
+    @artefact = Artefact.new(content_id: SecureRandom.uuid)
+  end
+
+  def create
+    @artefact = Artefact.new
+    if @artefact.update_attributes_as(current_user, creatable_params)
+      redirect_to publication_path(@artefact)
+    else
+      render "new"
+    end
+  end
+
+  def update
+    artefact = Artefact.find(updatable_params[:id])
+    if artefact.update_attributes_as(current_user, updatable_params)
+      PublishingAPIUpdater.perform_async(artefact.latest_edition_id)
+      flash[:notice] = "Metadata updated"
+    else
+      flash[:danger] = artefact.errors.full_messages.join("\n")
+    end
+    redirect_to metadata_artefact_path(artefact)
+  end
+
+  helper_method :formats
+
+private
+
+  def formats
+    Artefact::FORMATS_BY_DEFAULT_OWNING_APP['publisher']
+  end
+
+  def metadata_artefact_path(artefact)
+    edition = Edition.where(panopticon_id: artefact.id).order_by(version_number: :desc).first
+    metadata_edition_path(edition)
+  end
+
+  def creatable_params
+    params.require(:artefact).permit(:content_id, :name, :slug, :kind, :owning_app, :language)
+  end
+
+  def updatable_params
+    params.require(:artefact).permit(:id, :slug)
+  end
+end
--- a/test/integration/add_artefact_test.rb
+++ b/test/integration/add_artefact_test.rb
@@ -1,0 +1,31 @@
+require "integration_test_helper"
+
+class AddArtefactTest < ActionDispatch::IntegrationTest
+  setup do
+    setup_users
+    stub_linkables
+  end
+
+  should "create a new artefact" do
+    visit root_path
+    click_link "Add artefact"
+
+    fill_in "Name", with: "Thingy McThingface"
+    fill_in "Slug", with: "thingy-mc-thingface"
+    select "Help page", from: "Kind"
+
+    click_button "Save and go to item"
+
+    assert page.has_content?("Help page slugs must have a help/ prefix")
+
+    fill_in "Slug", with: "help/thingy"
+
+    click_button "Save and go to item"
+
+    assert %r{^\/editions\/[a-f0-9]*$} =~ page.current_path
+
+    edition = HelpPageEdition.last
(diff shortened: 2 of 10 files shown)


app/controllers/artefacts_controller.rb
```diff
@@ -0,0 +1,55 @@
+class ArtefactsController < ApplicationController
+  def new
+    @artefact = Artefact.new(content_id: SecureRandom.uuid)
+  end
+
+  def create
+    @artefact = Artefact.new
+    if @artefact.update_attributes_as(current_user, creatable_params)
+      edition = create_edition(@artefact)
```
davidbasalla: Some thoughts on this. Previously in Panopticon, after having created the artefact the controller would redirect to "/admin/publications/#{artefact.id}" in Publisher. The [corresponding controller method](https://github.com/alphagov/publisher/blob/master/app/controllers/publications_controller.rb#L2-L10) has some logic which does a `find_or_create` on the Edition, so the creation of the edition was handled in the PublicationsController. Do you have any strong opinions on keeping that behaviour? (mostly asking because it touches our other PR).

test/integration/add_artefact_test.rb
```diff
@@ -0,0 +1,21 @@
...
+    visit root_path
+    click_link "Add artefact"
+
+    fill_in "Name", with: "Thingy McThingface"
+    fill_in "Slug", with: "help/thingy-mc-thingface"
+    select "Help page", from: "Kind"
+
+    click_button "Save and go to item"
+
+    assert page.has_content?("Thingy McThingface")
```
davidbasalla: Nice. Could we check that the slug and kind have been set correctly too or is that overkill?

test/integration/edit_artefact_test.rb
```diff
@@ -0,0 +1,40 @@
+require "integration_test_helper"
+
+class EditArtefactTest < ActionDispatch::IntegrationTest
+  setup do
+    setup_users
+    stub_linkables
+  end
+
+  should "create a new artefact" do
```
davidbasalla: Should this be removed as it has a separate test file?

test/integration/edit_artefact_test.rb
```diff
@@ -0,0 +1,40 @@
...
+  should "edit a draft artefact" do
+    edition = FactoryGirl.create(:edition)
+    visit metadata_edition_path(edition)
+
+    fill_in "Slug", with: "thingy-mc-thingface"
+    select "Welsh", from: "Language"
+
+    click_button "Update metadata"
+
+    assert page.has_content?("Metadata updated")
```
davidbasalla: Could we check here that the changes have persisted?

davidbasalla:
Looks good, have added a few comments. Also the other controllers have functional tests, would it be worth writing some for the new ArtefactsController?

whoojemaflip:
There's coverage in the Integration tests, and I'm not sure that writing lower level Functional tests for the controller will add much value. [Relevant materials in support of this claim.](https://github.com/alphagov/publisher/blob/master/test/functional/transactions_controller_test.rb)

test/integration/edit_artefact_test.rb
whoojemaflip: see below...

fofr:
Do users understand what an artefact is?

fofr:
I think this UI could be improved.

* Creating a new document should be provided in the context that users would want to perform that action, eg it might not be the top nav
* The term "Artefacts" is an artefact of the system, use terms like "document" or "article"
* Please test this with users

davidbasalla:
@whoojemaflip I think if we're not adding functional/controller tests then we should add some more edge case tests to the integration tests. For example what happens when an artefact can not be created or updated, to exercise more of the code branches

whoojemaflip:
@fofr We've shown this functionality to a group of content designers, who didn't seem particularly put out by the naming. It's directly lifted from Panopticon you see. While I might agree that it's poorly named, renaming the function at this stage is probably going to be a worse user experience.

As to the positioning of the link in the top-nav: do you have any suggestions regarding where else it might exist?

whoojemaflip:
oh and @fofr do feel free to come over and discuss these concerns with us directly. We're friendly :)

whoojemaflip:
@davidbasalla we've updated the integration tests to ensure coverage of all controller logic branches.

app/controllers/artefacts_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+  def formats
+    Artefact::FORMATS_BY_DEFAULT_OWNING_APP['publisher']
+  end
+
+  def metadata_artefact_path(artefact)
+    edition = Edition.where(panopticon_id: artefact.id).order_by(version_number: :desc).first
+    metadata_edition_path(edition)
+  end
+
+  def create_edition(artefact)
```
davidbasalla: we don't need this method anymore

fofr:
@whoojemaflip Regarding:
> While I might agree that it's poorly named, renaming the function at this stage is probably going to be a worse user experience.

I don't understand why a better name would be a worse experience?

> We've shown this functionality to a group of content designers

Have you done user research though? Usually best to do some research with people relatively new to the application.

I'm happy for these changes to go through if there is a plan to iterate.

davidbasalla:
Looks good 👍

PR #531, opened 2017-01-17
Move publishing api calls to controllers

davidbasalla:
This commit aims to solve the following problems:
- Publisher creating a draft content item in PublishingAPI after making a 'publish' request to PublishingAPI
- Publisher sending the wrong content to PublishingAPI before a 'publish' request
- Publisher sending multiple redundant 'put_content' requests as part of its Edition-related workflow

Before this commit, the Edition model had an after_save hook that would make a 'put_content' request to PublishingAPI each time an Edition was saved. This would happen multiple times per state change for an edition (see workflow.rb) due to the state machine and extra calls to record users (see denormalize users). Since in the PublishingAPI we do not yet model the Publisher workflow, doing a put_content for these state changes is redundant, however it is possible to change edition content while also changing state (e.g when requesting 2i without clicking save directly ) so we cannot remove all the calls.

Also when an edition is published, each previous edition for that artefact would be set to 'archived', thereby triggering the save method and put_content calls to PublishingAPI. This led to the wrong content being in the PublishingAPI when 'publish' was finally called. Also, after the publish, the published edition in Publisher would be saved again, causing another put_content which would create a new draft content_item in the PublishingAPI.

To fix the issues described above, we remove the after_save hook from the Edition Model and call the put_content commands from the relevant controller methods instead. This makes it more obvious at which point Publisher communicates with the PublishingAPI and resolves the issues described above.

We refactor the functional tests for EditionController and PublicationsController to make add tests for calls to the PublishingAPI simpler.

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -39,6 +39,8 @@ class EditionsController < InheritedResources::Base
     @publication = current_user.create_edition(class_identifier, create_params[:edition])
 
     if @publication.persisted?
+      notify_update_publishing_api(@publication)
+
       flash[:success] = "#{description(@publication)} successfully created"
       redirect_to edition_path(@publication)
       return
@@ -56,7 +58,10 @@ class EditionsController < InheritedResources::Base
       flash[:warning] = 'Another person has created a newer edition'
       redirect_to edition_path(resource)
     elsif command.duplicate(target_edition_class_name, new_assignee)
-      return_to = params[:return_to] || edition_path(command.new_edition)
+      new_edition = command.new_edition
+      notify_update_publishing_api(new_edition)
+
+      return_to = params[:return_to] || edition_path(new_edition)
       flash[:success] = 'New edition created'
       redirect_to return_to
     else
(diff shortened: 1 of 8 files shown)


app/controllers/editions_controller.rb
```diff
@@ -56,7 +58,10 @@ def duplicate
       flash[:warning] = 'Another person has created a newer edition'
       redirect_to edition_path(resource)
     elsif command.duplicate(target_edition_class_name, new_assignee)
-      return_to = params[:return_to] || edition_path(command.new_edition)
+      new_edition = command.new_edition
+      PublishingAPIUpdater.perform_async(new_edition.id.to_s)
```
whoojemaflip: It would be good to get these duplicate calls into a method.

whoojemaflip:
LGTM :+1:

davidbasalla:
While we're sort of touching this code, do you mind having a look to see if we need to set `republish` in https://github.com/alphagov/publisher/blob/5ad1a9a914e472f09d620a9ef252683fc25c1c64/app/workers/publishing_api_updater.rb#L9? It's something I meant to have a look at. I think PublishingPlatform reckoned that it was redundant...

PR #532, opened 2017-01-18
Move to new ci

mgrassotti:
Trello: https://trello.com/c/UCcCp13D/576-move-publisher-to-new-ci-infrastructure-1

The old CI infrastructure is in the process of being wound down so we need to move our apps to the new one.


brenetic:
Hi @mgrassotti, does this also deal with the `govuk_content_schemas` kicking off a `govuk_publisher_schema_tests` build or do we need to hang on to the `jenkins-schema.sh` for the near future?

mgrassotti:
I don't think it launches any schema test itself. @steventux do you think the same? Thanks

suzannehamilton:
If you want to want `govuk_content_schemas` to kick off a `publisher` build to test the schema, you can do it like this:

- Enable builds of the `deployed-to-production` branch by editing the `govuk-puppet` Jenkins config. Example: alphagov/govuk-puppet#5364
- Add `publisher` to the `dependentApplications` list in the `govuk-content-schemas` Jenkinsfile: https://github.com/alphagov/govuk-content-schemas/blob/master/Jenkinsfile

Once that's merged, you can delete `jenkins-schema.sh` and remove the build from the `govuk_content_schemas` downstream builds on Jenkins 1.

mgrassotti:
depends on https://github.com/alphagov/govuk-content-schemas/pull/491 and https://github.com/alphagov/govuk-puppet/pull/5368

mgrassotti:
hello @suzannehamilton sorry, I tried to configure as you suggested. 
Now `govuk-puppet` was merged but `govuk-content-schemas` is saying
`ERROR: No parameterized job named publisher/deployed-to-production found`. Do you have any suggestions? Many thanks

suzannehamilton:
Hi @mgrassotti. The issue is that the Jenkinsfile is not on the publisher `deployed-to-production` branch yet. You'll have to merge this PR and deploy it before you can use publisher as a schema test.

It looks like the govuk-puppet config is fine - I ran the Jenkins [branch indexing job](https://ci.integration.publishing.service.gov.uk/job/publisher/indexing/), and the [logs](https://ci.integration.publishing.service.gov.uk/job/publisher/indexing/console) show that it is now including the `deployed-to-production` branch in the branches that it scans. But it won't actually _build_ `deployed-to-production` in the schema tests until the Jenkinsfile exists on the branch.

mgrassotti:
@suzannehamilton many thanks

Rakefile
```diff
@@ -4,6 +4,7 @@
 require File.expand_path('../config/application', __FILE__)
 require 'rake'
 require 'ci/reporter/rake/minitest' if Rails.env.test?
+require 'ci/reporter/rake/rspec'
```
h-lame: Because the ci-reporter gem is only available in the test gem group, you need to add the same guard clause to this new line as above.  Otherwise you block any rake tasks run on deployed environments where we don't install the test gem group.

mgrassotti: @h-lame Sorry for that, amended in https://github.com/alphagov/publisher/pull/541

h-lame: No worries, only reason I know is because I was burned by trying to deploy a couple of apps that had already had this change made.

PR #535, opened 2017-01-23
Register content store document type with Rummager

alecgibson:
This change will also publish the Content Store Document Type to Rummager, so that we can filter by Document Type in searches.

Trello: https://trello.com/c/M1cXENeB/341-add-content-store-document-type-to-all-guidance-content-in-rummager

### DIFF
--- a/app/presenters/search_payload_presenter.rb
+++ b/app/presenters/search_payload_presenter.rb
@@ -28,6 +28,7 @@ class SearchPayloadPresenter
       indexable_content: indexable_content,
       link: "/#{slug}",
       public_timestamp: public_timestamp,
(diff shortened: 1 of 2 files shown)


tijmenb:
Can you add a explicit test that what's being sent to the content store is exactly the same as what's being sent to rummager?

alecgibson:
@tijmenb I've added a test - do you want to check if it's sufficient?

alecgibson:
I should mention that I've also tested this locally by running the `rummager:index` Rake task (on a single entry, not on all of them), and verifying the presence of the field in `search-admin`.

tijmenb:
👍 nice

PR #536, opened 2017-01-23
Change the schema of Help Pages so the routes are managed by Publishing-API

whoojemaflip:
The first of the Publisher owned formats to have their routes managed by
content-store. This changes the published schema of the help pages from
'placeholder' to 'generic_with_external_related_links'.

In addition, we make it so that Publisher does not attempt to manage
routes or rummager for this content.

### DIFF
--- a/app/presenters/published_edition_presenter.rb
+++ b/app/presenters/published_edition_presenter.rb
@@ -5,6 +5,38 @@ class PublishedEditionPresenter
   end
 
   def render_for_publishing_api(options={})
+    if @artefact.generic_schema?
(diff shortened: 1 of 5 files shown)


tijmenb:
> In addition, we make it so that Publisher does not attempt to manage routes or rummager for this content.

Why no rummager registration anymore? Don't we still need it in search?

whoojemaflip:
@tijmenb will content-store not be managing registration with Rummager?

whoojemaflip:
Thanks for the heads up.

PR #537, opened 2017-01-23
Update all publisher schemas to generic_with_external_related_links

whoojemaflip:
Remove the guard clauses, this PR enables us to republish all Publisher content with the new schema, and in doing so allows publishing-api to manage the routes for these content items.

https://trello.com/c/VV0MJM5u/616-update-publisher-to-manage-routes-with-publishingapi-5


test/unit/edition_test.rb
```diff
@@ -9,12 +9,9 @@ class EditionTest < ActiveSupport::TestCase
       edition = FactoryGirl.create(:guide_edition, :state => "ready", panopticon_id: artefact.id)
 
       registerable = mock("registerable_edition")
-      routable = stub
-      routable.expects(:submit)
       PublishingAPIPublisher.expects(:perform_async)
       RegisterableEdition.expects(:new).with(edition).returns(registerable)
       SearchIndexer.expects(:call).with(registerable)
-      RoutableArtefact.expects(:new).with(artefact).returns(routable)
```
davidbasalla: Why is this being removed here?

davidbasalla:
Small style niggle, the first commit message doesn't follow the convention of starting with active verb etc. It might also be helpful to add a small explanation for that commit

davidbasalla:
Looks good!

davidbasalla:
I think https://github.com/alphagov/publisher/blob/master/lib/licence_content_importer.rb#L16 can be removed?

PR #538, opened 2017-01-24
Fix Rummager search bug

brenetic:
We discovered this bug in Errbit
(https://errbit.publishing.service.gov.uk/apps/5304e0520da11511b0000011/problems/58872a52657863786d413b00)
where when a piece of content is unpublished, the item is not removed from Rummager.

This was to do with a deprecated `delete_content!` method. We also
removed the functionality that retried deletion.


davidbasalla:
Nice! What was the thinking of removing the retries, was that originally put there for a specific reason? Also does that mean we now don't have any logging around failure of talking to Rummager? Or is that caught by some other error handling?

whoojemaflip:
The retry functionality depended on the exception being thrown in the now-deprecated `!` method.

I think it's safe to say that we should be both logging failures and handling retries in a consistent way across all external services. Maybe we should consider this in our backlog?

whoojemaflip:
The retry code was actually copy/pasted from Panopticon and it literally only needed to work for 2 days! (It clearly didn't even manage that).

In terms of failure logging, it is now the same as every other call to an external service than Publisher makes.

davidbasalla:
Nice! Merging

PR #540, opened 2017-01-25
Add router-data compatible export tool for archived multipart content

whoojemaflip:
There's a bug in the publishing platform that means all multi-part documents are incorrectly redirected upon archival.

As a short-term workaround, here we have a tool to create a router-data compatible CSV file with the redirects of all archived multi-part content. 

The resultant CSV will need to be integrated into router-data, independently.

Trello:
https://trello.com/c/jWVgYcxQ/618-investigate-unpublishing-parted-formats-with-redirects


davidbasalla:
Looks good!

PR #542, opened 2017-01-30
Create presenter for new Help schema

brenetic:
As part of mainstream format migration, we've created a content-schema
for help pages.

To support this, and future content schemas, we've created a small
number of classes:

A Format::HelpPagePresenter which converts a HelpPageEdition into the
correct format for Publishing-API
An EditionPresenterFactory which chooses and configures the correct
Format Presenter class for the type of content.

Note: The tests will only pass once https://github.com/alphagov/govuk-content-schemas/pull/500 has been merged (and deployed!)


app/presenters/formats/edition_format_presenter.rb
```diff
@@ -0,0 +1,83 @@
...
+          url: link["url"],
+          title: link["title"]
+        }
+      end
+    end
+
+    def routes
+      [
+        { path: "#{base_path}", type: path_type },
+        { path: "#{json_path}", type: "exact" }
```
boffbowsh: is the json path necessary? What serves it? https://www.gov.uk/help/cookies.json 404s, for example.

app/services/edition_presenter_factory.rb
```diff
@@ -0,0 +1,16 @@
+module EditionPresenterFactory
```
boffbowsh: What's the reasoning behind this being a module rather than a class with class methods? I'd not seen `module_function` before, it's a little confusing.

app/presenters/formats/edition_format_presenter.rb
whoojemaflip: Good catch @boffbowsh - https://github.com/alphagov/frontend/pull/1113 should fix that!

app/services/edition_presenter_factory.rb
whoojemaflip: It's maybe a pattern associated with a more functional style of Ruby programming. We decided that these factory functions could work without any side-effects or dependencies.

No harm in changing it to be a `Class` with `self.methods` if that;'s the house style.

boffbowsh: I think it'd be better in this instance to keep things consistent. I believe even `extend self` might be preferable to `module_function` here.

app/presenters/formats/edition_format_presenter.rb
boffbowsh: What is it supposed to present? Does anything actually use it? Would it be better served by `/api/content/help/cookies` ?

davidbasalla: These json requests redirect to the content api in [api_redirectable.rb](https://github.com/alphagov/frontend/blob/master/app/controllers/concerns/api_redirectable.rb) which is included in all the format controllers. But yeah, probably makes sense to redirect to the content store api instead... we'll update Frontend to redirect to that as we migrate I reckon

boffbowsh: If it's a redirect, I don't think we should continue to serve it. It's an unsupported API redirecting to another unsupported API. There were no requests to `/help/*.json` at all yesterday, for example. I haven't looked further back because I can't quite get the Kibana syntax right. Command to run on `cache-*.router.production`:
```bash
grep -E ' /help/.*json' /var/log/nginx/lb-access.log.1
```

boffbowsh: One for @timblair, can we retire `/help/foo.json`, which is currently just a straight redirect to `/api/help/foo.json`, which is also deprecated?

brenetic:
@boffbowsh new content schema is deployed so we are now green on ci.integration.

@timblair Any thoughts on @boffbowsh's comment about the json api?

app/presenters/formats/edition_format_presenter.rb
timblair: Yup, bin it.

davidbasalla: Great, we'll remove it from this presenter. We'll put a card into our Team Board for removing the functionality from Frontend as well (and remove the relevant routes)

sihugh:
Given that we'll catch up on the JSON routes work subsequently, I think this is great. 👍

PR #543, opened 2017-01-31
Add help page sync checker rake task

brenetic:
In order to check that we have migrated the help pages correctly this rake task will check certain fields have not changed after the format has been republished using the new schema.

### DIFF
--- a/lib/sync_checker.rb
+++ b/lib/sync_checker.rb
@@ -1,0 +1,122 @@
+require 'gds_api/content_store'
+
+class SyncChecker
+  class Success
+    attr_reader :base_path
+
+    def initialize(base_path:, content_store:)
+      @base_path = base_path
+      @content_store = content_store
+    end
+
+    def to_s
+      "✅  #{@base_path} in #{@content_store}"
+    end
+  end
+
+  class NotFoundFailure
+    attr_reader :base_path, :content_store
+
+    def initialize(base_path:, content_store:)
+      @base_path = base_path
+      @content_store = content_store
+    end
+
+    def to_s
+      "❌  Failed path: #{base_path} in #{content_store}, failed expectations: Not found\n\n"
+    end
+  end
+
+  class Failure
+    attr_reader :edition, :content_item, :base_path, :failed_expectations, :content_store
+
+    def initialize(edition:, content_item:, base_path:, failed_expectations:, content_store:)
+      @edition = edition
+      @content_item = content_item
+      @base_path = base_path
+      @failed_expectations = failed_expectations
+      @content_store = content_store
+    end
+
+    def to_s
+      str = "😡  Failed path: #{base_path} in #{content_store}, failed expectations: #{failed_expectations.join(', ')}"
+      str << "\nExpected: #{filter_attributes(presented_edition)}"
+      str << "\nActual: #{filter_attributes(content_item)}" if content_item
+      str << "\n\n"
+      str
+    end
+
+  private
+
+    def filter_attributes(attrs)
+      attrs.symbolize_keys.select { |a| meaningful_attributes.include?(a) }
+    end
+
+    def meaningful_attributes
+      %i(schema_name document_format title public_updated_at)
+    end
+
+    def presented_edition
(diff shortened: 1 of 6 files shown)


brenetic:
@sihugh Would be great if you could have a look through this, let us know if you need any context. The CI Dev tests are failing, but CI Integration is fine (need to check this with @mgrassotti because of https://github.com/alphagov/publisher/commit/7e6e8da1d389fb59d589a3d4377803483f38e497)

sihugh:
Having a look now, thanks :-)

sihugh:
Do we know (I presume from Core) what sorts of things they found were inconsistent?  Was it the existence of the item, or some more specific things.  I was wondering whether it was worth checking a field that is likely to change across versions.  The `schema_name`, `document_type` and `title` might all be quite static so if we did a bulk republish, we might not be able to verify that it had all worked.

lib/tasks/sync_checks.rake
```diff
@@ -0,0 +1,44 @@
...
+    puts "Checking live content"
+    check_content(['published'], 'content-store')
+  end
+
+  def check_draft
+    puts "Checking draft content"
+    # This can be changed to use the PRELUDIAL_STATES constant from the Edition
+    # class shortly
+    check_content(
+      %w(amends_needed draft fact_check fact_check_received in_review ready),
```
davidbasalla: We're missing `scheduled_for_publishing`?

```
 Edition.pluck(:state).uniq
=> ["archived", "published", "fact_check", "ready", "amends_needed", "draft", "fact_check_received", "in_review", "scheduled_for_publishing"] 
```

davidbasalla:
good point @sihugh, we'll investigate whether we check the 'public_updated_at' attribute or similar... 
We thought about testing `details['body']` but that's a bit trickier because of GovSpeak I think.

davidbasalla:
I'm assuming you mean for the other formats to be migrated? Currently we're testing for schema_name which we've added as an expectation in the rake task. I imagine once we move onto the next format, we'll add more rake tasks per format (with the appropriate schema_name expectations). We'll have a look at tests today hopefully!

sihugh:
Lovely jubbly.

lib/sync_checker.rb
```diff
@@ -0,0 +1,112 @@
...
+private
+
+  def check(edition)
+    base_path = "/#{edition.slug}"
+    response = content_store.content_item(base_path)
+
+    content_item = response.to_h
+    compare_content(content_item, edition, base_path)
+  rescue
+    Failure.new(
```
whoojemaflip: Would a `NotFoundFailure` class help here?

davidbasalla:
@whoojemaflip see last commit? Will squash it if you're happy with it?

test/unit/tasks/republish_content_test.rb
```diff
@@ -13,9 +13,11 @@ class RepublishContentTest < ActiveSupport::TestCase
       PublishingAPIRepublisher.expects(:perform_async).with(@published_edition.id.to_s)
       PublishingAPIUpdater.expects(:perform_async).with(@draft_edition.id.to_s)
 
+      # This is to prevent the rake task ouputting to the console
       silence_stream(STDOUT) do
         silence_stream(STDERR) do
-          Rake::Task['publishing_api:republish_content'].invoke
+          Rake::Task['publishing_api:republish_content'].reenable
+          Rake.application.invoke_task('publishing_api:republish_content')
```
sihugh: Why does this need a different invocation to the others?

test/unit/lib/sync_checker_test.rb
```diff
@@ -0,0 +1,41 @@
...
+      SyncChecker::Failure.stubs(new: stub(to_s: "boo"))
+      SyncChecker::Failure.expects(:new).once
+
+      @sync_checker.add_expectation("schema_name") do |content_item, _|
+        content_item["schema_name"] == "other_format"
+      end
+
+      @sync_checker.call
+    end
+  end
```
sihugh: We should probably have a test for the `NotFoundFailure` too

PR #544, opened 2017-01-31
Add republishing for draft content by format

brenetic:
Previously our republishing by format rake task was only acting on content in the `published` state. We also need all draft content (and other pre published states) to be `put` onto the draft content
store.


lib/tasks/publishing_api.rake
```diff
@@ -8,15 +8,24 @@ namespace :publishing_api do
...
-    editions = Artefact.published_edition_ids_for_format(args[:format])
+    format = args[:format]
+    edition_ids = Edition.published_by_format(format).map(&:id).map(&:to_s)
 
-    puts "Scheduling republishing of #{editions.count} #{args[:format]}s."
+    puts "\nScheduling republishing of #{edition_ids.count} #{format}s."
 
-    editions.each do |edition_id|
+    edition_ids.each do |edition_id|
       PublishingAPIRepublisher.perform_async(edition_id)
```
h-lame: As we're doing a republish, should we also be sending `update_type: "republish"` to these params?  Like it does [here](https://github.com/alphagov/publisher/blob/17be550910d65901deee0610db4cf4dc704121e4/lib/republish_content.rb#L4) - which of course, leads me to wonder, what that lib file is for vs. this rake task - because they're both about republishing, but now they do different things.

h-lame:
This looks good to me - it seems to bring it more in line with how whitehall republishes content, and I'm doing something similar to manuals-publisher.  Consistency FTW.

lib/tasks/publishing_api.rake
whoojemaflip: Totes agree.

PR #549, opened 2017-02-03
Fix update type when publishing content

brenetic:
Previously in Publisher we were sending the update_type as "minor" for
every publish. This resulted in the publishing-api never knowing about
"major" updates.

We now send nil as the default update_type as we can allow
publishing-api to make the decision to use the editions update type as
seen here:

https://docs.publishing.service.gov.uk/apis/publishing-api/api.html#post-v2-content-content-id-publish


sihugh:
👍

PR #550, opened 2017-02-06
Send locale when unpublishing

davidbasalla:
Previously we were not sending the locale when unpublishing, which meant that Welsh content would not be unpublished properly against the publishing api.

For https://trello.com/c/WfsT2KM6/634-fix-unpublishing-welsh-content-bug-in-publisher


brenetic:
👍

whoojemaflip:
:yes:

sihugh:
edrych yn dda i mi - or something

PR #551, opened 2017-02-06
Fix ISO 8601 timezone format in downstream documents

whoojemaflip:
Documents returned from the content-store have inconsistant date formats. Some use the trailing modifier `Z` to signify UTC, whereas others are represented using the `+00:00` time zone offset.

For consistency, and because the publishing-platform has standardised on the latter, this changes datetime strings in rendered JSON to use the `rfc3339` standard.

screen shot 2017-02-06 at 18 34 39

### DIFF
--- a/app/presenters/formats/edition_format_presenter.rb
+++ b/app/presenters/formats/edition_format_presenter.rb
@@ -31,7 +31,7 @@ module Formats
         schema_name: schema_name,
         document_type: artefact.kind,
         need_ids: [],
-        public_updated_at: public_updated_at,
+        public_updated_at: public_updated_at.to_datetime.rfc3339(3),
         publishing_app: "publisher",
         rendering_app: "frontend",
         routes: routes,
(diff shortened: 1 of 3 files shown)


fofr:
@gpeng I understood the +00:00 format was the preferred one?

whoojemaflip:
@fofr @gpeng Literally the [first format example I looked at](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/case_study/frontend/examples/archived.json) to check uses `UTC`, `GMT + 0` and `BST` in its first three datetime objects. Personally, I think `UTC` (via the `Z` suffix is the least ambiguous spec to follow.

gpeng:
I think firstly started checking with the `+00:00 format as that matches Publishing API was setting for `first_published_at`.  On having a look through the JSON schema spec the 'correct' date format is `rfc3339` like we check [here](https://github.com/alphagov/whitehall/blob/master/lib/sync_checker/formats/edition_base.rb#L170) but our schemas appear to be valid with any format. [JSON Schema Spec](http://json-schema.org/latest/json-schema-validation.html#rfc.section.7.3.1)

So... I think we should standardise to that but there will be work to do through the pipeline to get that to happen.

whoojemaflip:
@gpeng @fofr Thanks for the feedback. PR updated accordingly to standardise on `rfc3339`

sihugh:
👍

PR #552, opened 2017-02-07
Remove segments from mulitpart redirects

brenetic:
Publisher currently allows segment redirects within its' GUI. For our temporary task to export multipart redirects to router-data we will be removing all segment redirects and only redirect to the top of the page.

https://trello.com/c/4TxuwEgo/636-fix-publisher-multipart-csv-export-task

### DIFF
--- a/lib/tasks/router_data.rake
+++ b/lib/tasks/router_data.rake
@@ -13,7 +13,7 @@ namespace :router_data do
 
     csv = "Source,Destination,Type\n"
     csv << artefacts
-      .map { |arr| "/#{arr[0]},#{arr[1]},prefix" }
+      .map { |arr| "/#{arr[0]},#{arr[1].gsub(/#.*/, '')},prefix" }
       .join("\n")
 
     File.write(filename, csv)


davidbasalla:
Looks good. If this work around gets any more complex, we should think about adding some tests.

fofr:
What's a segment redirect?

brenetic:
e.g `/guidance/animal-welfare#transporter-authorisation`

In [this PR](https://github.gds/gds/router-data/commit/e1a0f5c3affdb261f87104547c03ed4b23c4b591) we allowed the segments to redirect to a page fragment identifier. However going forward we don't feel that is robust as content changes regularly and redirects like these may break in the future.

fofr:
Thanks for the clarification @brenetic.

> However going forward we don't feel that is robust as content changes regularly and redirects like these may break in the future.

If a segment is renamed or removed, then the page would still load, but it wouldn't scroll to the right position (essentially what will happen after this PR, I believe). For this reason I've seen these type of redirects as an enhancement – if the content is there and valid, the user is taken to the right place, to the thing they need.

Taking this out feels like a regression for all the working segment redirects, negatively impacting users.

PR #553, opened 2017-02-07
[DISCUSS] - Add segments column to multipart redirects export task

brenetic:
[This PR](https://github.com/alphagov/publisher/pull/552) changed the export of the multipart redirects to remove linking to a page fragment identifier. This was felt to not be robust as content changes regularly and these identifiers may not exist in the future.

An alternative to this is to add the `Segments` column to the exported csv. Router-API would then [take care](https://github.com/alphagov/router-api/blob/8828caf69a69533023a1f6fe68bc7276ccc42b6f/app/models/route.rb#L77) of whether or not to preserve or ignore the segment redirect. Broken page fragment identifiers would simply send the user to the top of the page.

### DIFF
--- a/lib/tasks/router_data.rake
+++ b/lib/tasks/router_data.rake
@@ -11,14 +11,18 @@ namespace :router_data do
 
     puts "Writing file with #{artefacts.count} redirects"
 
-    csv = "Source,Destination,Type\n"
+    csv = "Source,Destination,Type,Segments\n"
     csv << artefacts
-      .map { |arr| "/#{arr[0]},#{arr[1].gsub(/#.*/, '')},prefix" }
+      .map { |arr| "/#{arr[0]},#{arr[1]},prefix#{segment(arr[1])}" }
       .join("\n")
 
     File.write(filename, csv)
 
     puts "#{filename}"
     puts "Complete."
+  end
+
+  def segment(redirect_url)
+    redirect_url.include?('#') ? ',ignore' : ''
   end
 end


lib/tasks/router_data.rake
```diff
@@ -11,9 +11,9 @@ namespace :router_data do
 
     puts "Writing file with #{artefacts.count} redirects"
 
-    csv = "Source,Destination,Type\n"
+    csv = "Source,Destination,Type,Segments\n"
     csv << artefacts
-      .map { |arr| "/#{arr[0]},#{arr[1].gsub(/#.*/, '')},prefix" }
+      .map { |arr| "/#{arr[0]},#{arr[1]},prefix" }
```
whoojemaflip: I think you need to swap the positions of `arr[1]` and `prefix` for this to work

brenetic: Should be ok as is. Keeping the same structure as [here](https://github.gds/gds/router-data/blob/5c446e3e2a31499cc0afdb19a8dcc91576c78dab/data/worldwide_priority_redirects.csv)

whoojemaflip: I was reading this wrong, apols.

brenetic: No, you're quite right. Something is required there or router-api will complain. Behold! The beauty.

davidbasalla:
As discussed at standup, I'm going to close this PR. https://github.com/alphagov/publisher/pull/552 was the solution to dealing with fragment redirects. In general this rake task only exists to support [a temporary solution in router-data](https://github.gds/gds/router-data/pull/619) until PublishingAPI supports unpublishing of 'prefix' routes ([trello card](https://trello.com/c/fv1MfDsg/877-support-route-redirects-in-unpublishing-to-same-degree-as-a-redirect-document-type-2-blocking-custom)) and we've agreed with our PM that this solution is fine until then (as it only affects 2 redirects). Please come talk to us if you have any objections.

PR #555, opened 2017-02-08
Remove publish to publishing api from edition callback

davidbasalla:
When an edition's state changed to 'published' as part of the state machine transitions, Publisher would make calls to PublishingAPI and Rummager as a callback after save. We had previously separated any `put_content` calls to PublishingAPI from the edition and put them into the controllers instead because it's easier to follow when these calls happen. This does the same for the publish-related calls. We also create some service objects to group related calls together.

We also remove some legacy code that is not being used anymore.


brenetic:
👍  LGTM MWA

whoojemaflip:
💯

PR #556, opened 2017-02-08
Draft stack auth bypass for fact-checkers

whoojemaflip:
Using [JWT](https://trello.com/c/rROSYwYF/95-allow-token-based-access-to-preview-for-fact-checking-3) we can allow unauthenticated users to view specific pages in the draft
stack.

This change adds the JWT token to the generated URL in the fact-check
invitation emails. It will allow anyone with that URL to access the
specific draft edition it refers to. The URL will cease to work once
the edition has been published, and won't work for future editions.

https://trello.com/c/510Jc9Cl/628-migrate-help-page-fact-check-without-signon-account-3-3-8

### DIFF
--- a/app/helpers/paths_helper.rb
+++ b/app/helpers/paths_helper.rb
@@ -49,7 +49,7 @@ protected
 
   def preview_url(edition)
     if edition.migrated?
-      Plek.current.find("draft-frontend")
+      Plek.current.find("draft-origin")
     else
       Plek.current.find("private-frontend")
     end
--- a/app/helpers/paths_helper.rb
+++ b/app/helpers/paths_helper.rb
@@ -3,6 +3,17 @@ module PathsHelper
     "#{preview_url(edition)}/#{edition.slug}"
   end
 
+  def fact_check_edition_path(edition)
+    if edition.fact_check_id
+      token = jwt_token(sub: edition.fact_check_id)
+      url = preview_edition_path(edition)
+      url << "&token=#{token}"
+      url
+    else
+      preview_edition_path(edition)
+    end
+  end
+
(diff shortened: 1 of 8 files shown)


davidbasalla:
@danielroseman would you mind having a look at this one?

app/helpers/paths_helper.rb
```diff
@@ -3,6 +3,17 @@ def edition_front_end_path(edition)
     "#{preview_url(edition)}/#{edition.slug}"
   end
 
+  def fact_check_edition_path(edition)
```
whoojemaflip: is this missing a unit test?

lib/enhancements/edition.rb
```diff
@@ -140,4 +140,13 @@ def artefact
   def self.convertible_formats
     Artefact::FORMATS_BY_DEFAULT_OWNING_APP["publisher"] - ["local_transaction"]
   end
+
+  def fact_check_id
```
davidbasalla: @brenetic mentioned this was inspired by SecureRandom.uuid's implementation? Would be nice to leave a comment about that in the commit message, as this isn't the easiest code to understand at a glance

whoojemaflip: what would you like us to write about this?

danielroseman: I don't understand why this needs to be deterministic, rather than just using SecureRandom.

whoojemaflip: @danielroseman Rather than introduce the complexity of managing a (randomly generated) UUID over the object lifecycle, we opted for a more functional approach. The `#fact_check_id` method delivers an encoding of the `Edition.id` that validates as UUID only when the content workflow should allow fact-check access.

davidbasalla:
Merging this 🎉

PR #558, opened 2017-02-09
Fix schema type clash in upstream project CI

whoojemaflip:
Content-Schemas is failing CI because of a default Artefact `kind` being set in the content-models factory. This override should ensure we don't run into this problem again during migration.


brenetic:
👍

PR #559, opened 2017-02-09
Reduce test run time by 50s

whoojemaflip:
Current test runs take 263s on my dev VM. This change reduces that by 50s.


davidbasalla:
Noice! 👍

PR #562, opened 2017-02-13
Update sync checker to support other formats

davidbasalla:
Update sync checker to support other formats (for future migrated formats)


sihugh:
Looks good.  I guess we don't really have the need for an `ALL_FORMATS` option at the moment. MWG :+1:

PR #563, opened 2017-02-14
Improve CI when running as a downstream schema tests

whoojemaflip:
There were CI failures against upstream govuk-content-schema branches
because of the configuration of the linter in this Jenkinsfile which was
running against the diff of deployed-to-production and origin/master.

There's really no need for the linter to run when we're testing schema
compatibility, so we skip it in this case.


brenetic:
👍  LGTM. MWAWG

PR #564, opened 2017-02-14
Add answer to list of migrated formats

whoojemaflip:
For preview purposes, the `Artefact.migrated_format?` method must include `answer`

### DIFF
--- a/test/unit/edition_test.rb
+++ b/test/unit/edition_test.rb
@@ -72,7 +72,8 @@ class EditionTest < ActiveSupport::TestCase
 
     context "for a format that has not yet been migrated" do
       should "return nil" do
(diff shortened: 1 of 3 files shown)


test/unit/edition_test.rb
```diff
@@ -72,7 +72,8 @@ class EditionTest < ActiveSupport::TestCase
 
     context "for a format that has not yet been migrated" do
       should "return nil" do
-        edition = FactoryGirl.create(:edition, state: 'fact_check_received', id: 123)
+        artefact = FactoryGirl.create(:artefact, kind: 'licence')
```
davidbasalla: You could use the licence edition factory from content models?

PR #565, opened 2017-02-15
Update meaning of migrated

whoojemaflip:
We discovered an error when republishing Answer content. Namely,
our assumption that Artefact#kind will always equal the Edition type
was wrong. When an Edition changes format, the Artefact#kind doesn't
change with the edition.

### DIFF
--- a/lib/tasks/sync_checks.rake
+++ b/lib/tasks/sync_checks.rake
@@ -18,7 +18,7 @@ namespace :sync_checks do
   end
 
   def check_content(format, states, store)
-    scope = "#{format.classify}Edition".constantize
+    scope = Edition.by_format(format)
     editions = scope.where(state: { '$in' => states })
     checker = SyncChecker.new(editions, store)
     puts "#{editions.count} #{format.titleize} from #{store}"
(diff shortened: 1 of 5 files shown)


davidbasalla:
Ah good catch! We did [this](https://github.com/alphagov/publisher/blob/master/lib/tasks/sync_checks.rake#L21-L22) in the sync checker to determine the Edition class. If we did that we could remove the `by_format` method since it seems to only be called in the republishing rake task?

davidbasalla:
Cool! Can we squash the last commit?

lib/tasks/sync_checks.rake
```diff
@@ -27,8 +27,8 @@ namespace :sync_checks do
       content_item["schema_name"] == format
     end
 
-    checker.add_expectation("document_type") do |content_item, _|
-      content_item["document_type"] == format
+    checker.add_expectation("document_type") do |content_item, edition|
+      content_item["document_type"] == edition.artefact.kind
```
davidbasalla: So in light of artefacts not updating 'kind' when the format changes, won't this fail for some Editions?

davidbasalla:
Thinking about this, it does make me wonder whether we are setting 'document_type' correctly [here](https://github.com/alphagov/publisher/blob/master/app/presenters/formats/edition_format_presenter.rb#L32) 🤔 since conceivably we could end up content items in PublishingApi with differing schema_names and document_types when an artefact has changed format (but not updated 'kind')...

I know it's got nothing to do with this PR but probably worth a chat!

PR #566, opened 2017-02-16
Send document_type based on edition type

davidbasalla:
Previously this was determined by the 'kind' attribute on the edition's artefact. However when an edition was converted into another type (eg from Answer to Guide), the artefact's 'kind' would not updated. When sending this information to the publishing api, it could end up being misleading to have differening schema_name and document_type (to my knowledge, the two different attributes can be used define specialisations), so we prefer to send the document type as directly determined by the edition.

### DIFF
--- a/app/presenters/formats/edition_format_presenter.rb
+++ b/app/presenters/formats/edition_format_presenter.rb
@@ -29,7 +29,7 @@ module Formats
         base_path: base_path,
         description: edition.overview || "",
         schema_name: schema_name,
-        document_type: artefact.kind,
+        document_type: schema_name,
         need_ids: [],
         public_updated_at: public_updated_at.to_datetime.rfc3339(3),
         publishing_app: "publisher",
--- a/lib/tasks/sync_checks.rake
+++ b/lib/tasks/sync_checks.rake
@@ -28,7 +28,7 @@ namespace :sync_checks do
     end
 
     checker.add_expectation("document_type") do |content_item, edition|
-      content_item["document_type"] == edition.artefact.kind
+      content_item["document_type"] == format
     end
 
     checker.add_expectation("title") do |content_item, edition|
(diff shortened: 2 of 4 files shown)


app/presenters/formats/edition_format_presenter.rb
```diff
@@ -49,6 +49,13 @@ def schema_name
       "override me"
     end
 
+    def document_type
+      # alternatively this could just be schema_name since schema and
```
sihugh: Given that we expect `document_type` and `schema` to be the same, shouldn't we define them in the same way?  It feels like a bit of mental baggage to have to work out whether `edition.class.to_s.underscore.gsub('_edition', '')` is the same as `"answer"`.  Am happy to go either way on how to do it.

davidbasalla: Agreed, have changed this to `schema_name`

sihugh: Works for me :-)

lib/tasks/sync_checks.rake
```diff
@@ -28,7 +28,7 @@ namespace :sync_checks do
     end
 
     checker.add_expectation("document_type") do |content_item, edition|
-      content_item["document_type"] == edition.artefact.kind
+      content_item["document_type"] == format
```
brenetic: Will this not fail when we run a sync check for a document that has changed format type at some point in its' lifetime? I think that's what @whoojemaflip and I came across and why it was changed.

davidbasalla:
🎉 @brenetic we should theoretically republish any content which previously had different artefact.kinds (due to a format change). In practice we've only migrated help page so far so I don't think we need to do anything extra. And from now whenever we republish a migrated format, the document_type will always match the schema_name...

PR #560, opened 2017-02-09
Local transaction schema

davidbasalla:
Add presenter and sync checks for LocalTransaction format

CI won't pass until https://github.com/alphagov/govuk-content-schemas/pull/516 gets merged (and deployed)

PR #567, opened 2017-02-16
Add LocalTransaction to migrated formats for Preview

davidbasalla:
Without this preview won't link to draft stack but private frontend instead.


davidbasalla:
@whoojemaflip @brenetic I think this was missing from https://github.com/alphagov/publisher/pull/560

brenetic:
👍

PR #570, opened 2017-02-20
Force text fields to be strings

davidbasalla:
In the case of any of these fields being nil, they will fail against the published schemas, so we force them to be empty strings.


fofr:
Why go this way rather than making the schema more flexible?
Could the fields be omitted if they are empty and not required?

davidbasalla:
good point @fofr, what do you think about https://github.com/alphagov/publisher/pull/573 instead?

davidbasalla:
Closing in favour of https://github.com/alphagov/publisher/pull/573

PR #571, opened 2017-02-20
Add publishing-api presenter for Guide editions

whoojemaflip:
Trello: https://trello.com/c/r2gA1lxN/640-migrate-guide-format-to-rendered

Paired with @emmabeynon

### DIFF
--- a/app/presenters/formats/guide_presenter.rb
+++ b/app/presenters/formats/guide_presenter.rb
@@ -1,0 +1,31 @@
+module Formats
+  class GuidePresenter < EditionFormatPresenter
+  private
+
+    def schema_name
+      'guide'
+    end
+
+    def details
+      {
+        parts: parts,
+        external_related_links: external_related_links,
+      }
+    end
+
+    def parts
+      edition.parts.in_order.map do |part|
+        {
+          title: part.title.to_s,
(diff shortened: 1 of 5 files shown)


davidbasalla:
Are you planning to have another PR for adding the format to `MIGRATED_EDITION_CLASSES` for updating the preview link?

whoojemaflip:
@davidbasalla I think I must have pushed that commit at exactly the same time as your comment.

davidbasalla:
You could add a test for the migrated format to https://github.com/alphagov/publisher/blob/master/test/unit/edition_test.rb#L81-L101

Looks good otherwise! There might be a small edge case where a user could add a new part but not fill in the title. Publisher does not seem to validate the presence of `title` and lets you save the edition, but presumably pushing to the publishing api will fail because it expects a string (actually that depends if it gets `nil` or an empty string I suppose). Anyway, I had a look in Publisher in integration and couldn't find any published Guides where the title was `nil`, so it's probably nothing to worry about.

whoojemaflip:
Yep, always getting caught out by the nil values. @davidbasalla fixed

davidbasalla:
🎉 👍

PR #573, opened 2017-02-20
Do not present nil values for local transactions

davidbasalla:
The fields `introduction`, `more_information`, `need_to_know` and `lgil_override` are not required in the schema for local transactions. Therefore when these are nil, we don't send them at all (rather than sending nil which will fail schema validation or sending an empty string which is less explicit)

### DIFF
--- a/app/presenters/formats/local_transaction_presenter.rb
+++ b/app/presenters/formats/local_transaction_presenter.rb
@@ -7,34 +7,74 @@ module Formats
     end
 
     def details
-      {
-        lgsl_code: edition.lgsl_code,
-        lgil_override: edition.lgil_override,
-        service_tiers: service_tiers,
-        introduction: [
-          {
-            content_type: "text/govspeak",
-            content: edition.introduction,
-          },
-        ],
-        more_information: [
-          {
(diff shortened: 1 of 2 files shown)


brenetic:
👍 MWA

app/presenters/formats/local_transaction_presenter.rb
```diff
@@ -7,36 +7,76 @@ def schema_name
...
+    def required_details
       {
         lgsl_code: edition.lgsl_code,
-        lgil_override: edition.lgil_override,
         service_tiers: service_tiers,
+      }
+    end
+
+    def optional_details
+      {}.merge(lgil_override)
```
whoojemaflip: `{}.merge(lgil_override)` is identical to `lgil_override` - so you could save a few characters. Might not look so stylish though

fofr:
This new approach is good 👍

app/presenters/formats/local_transaction_presenter.rb
davidbasalla: I'm all style over substance 😎

PR #575, opened 2017-02-23
Fix unpublishing formats with parts redirects

emmabeynon:
https://trello.com/c/jWVgYcxQ/618-investigate-unpublishing-parted-formats-with-redirects

- [x] [API Adapters](https://github.com/alphagov/gds-api-adapters/pull/669)

Mobbed with @brenetic & @whoojemaflip

### DIFF
--- a/test/unit/presenters/search_index_presenter_test.rb
+++ b/test/unit/presenters/search_index_presenter_test.rb
@@ -1,7 +1,6 @@
 require "test_helper"
 
-class RegisterableEditionTest < ActiveSupport::TestCase
-
+class SearchIndexPresenterTest < ActiveSupport::TestCase
   def template_answer(version_number = 1)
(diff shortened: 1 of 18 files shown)


test/unit/presenters/search_index_presenter_test.rb
```diff
@@ -33,27 +32,27 @@ def template_unpublished_answer(version_number = 1)
 
   context "state" do
     should "return live if the edition is published" do
-      registerable = RegisterableEdition.new(@edition)
+      registerable = SearchIndexPresenter.new(@edition)
```
sihugh: I wonder if it's worth considering a better name than `registerable` now?

emmabeynon: Now changed to `presenter` ⚡️

PR #576, opened 2017-02-24
Migrate Completed Transaction to rendered

brenetic:
Migrate the Completed Transaction format

https://trello.com/c/unTNbCHi/658-migrate-completed-transaction-format-to-rendered

Note: tests will not pass until https://github.com/alphagov/govuk-content-schemas/pull/538 is merged.

Mobbed with @emmabeynon @whoojemaflip @binaryberry

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ gem 'gds-sso', '~> 11.2'
 gem 'gds-api-adapters', '~> 40.1.0'
 gem 'govspeak', '~> 3.4.0'
 gem 'govuk_admin_template', '4.2.0'
-gem "govuk_content_models", "~> 42.0.1"
+gem "govuk_content_models", "~> 43.0.0"
 gem 'govuk_sidekiq', '0.0.4'
 gem 'has_scope'
 gem 'inherited_resources'
--- a/app/presenters/formats/local_transaction_presenter.rb
+++ b/app/presenters/formats/local_transaction_presenter.rb
@@ -16,8 +16,6 @@ module Formats
       .merge(external_related_links: external_related_links)
     end
 
-  private
-
     def required_details
       {
         lgsl_code: edition.lgsl_code,
--- a/test/unit/presenters/formats/completed_transaction_presenter_test.rb
+++ b/test/unit/presenters/formats/completed_transaction_presenter_test.rb
@@ -1,0 +1,82 @@
+require 'test_helper'
+
+class CompletedTransactionPresenterTest < ActiveSupport::TestCase
+  include GovukContentSchemaTestHelpers::TestUnit
+
+  def subject
+    Formats::CompletedTransactionPresenter.new(edition)
+  end
+
+  def edition
+    @_edition ||= FactoryGirl.create(
+      :completed_transaction_edition,
+      :published,
+      title: "Whacked all moles",
+      slug: "done/good",
+      panopticon_id: artefact.id,
+    )
+  end
+
+  def artefact
+    @_artefact ||= FactoryGirl.create(:artefact, kind: "completed_transaction")
+  end
+
+  def result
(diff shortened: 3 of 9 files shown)


app/presenters/formats/local_transaction_presenter.rb
```diff
@@ -12,10 +12,6 @@ def details
       .merge(external_related_links: external_related_links)
     end
 
-    def registers_exact_route?
```
davidbasalla: Looks like this is being added in the previous comment and removed in this commit, so let's just remove completely from the diff

test/unit/presenters/formats/completed_transaction_presenter_test.rb
```diff
@@ -0,0 +1,72 @@
...
+  should "be valid against schema" do
+    assert_valid_against_schema(result, 'completed_transaction')
+  end
+
+  should "[:schema_name]" do
+    assert_equal 'completed_transaction', result[:schema_name]
+  end
+
+  context "[:details]" do
+    should "[:promotion_choice]" do
```
davidbasalla: This looks like a test for the nil case, could we add another for testing an actual promotion choice?

davidbasalla:
The last three commits could probably all be squashed into https://github.com/alphagov/publisher/pull/576/commits/d46109e375aaed77a7767648f61318341e10c4d8. Arguably the same first the first commit as it only changes one line

davidbasalla:
For previous formats, we added tests to test/unit/services/edition_presenter_factory_test.rb and test/unit/edition_test.rb to ensure that the format is migrated, we could do the same here

test/unit/presenters/formats/completed_transaction_presenter_test.rb
emmabeynon: We noticed this, so changed it to test for a promotion choice, and added a new test for when there is no promotion.

app/presenters/formats/local_transaction_presenter.rb
emmabeynon: I think is dealt with after we squashed a few commits 👍

emmabeynon:
@davidbasalla We've added in those tests - thanks!

emmabeynon:
We will need https://github.com/alphagov/govuk_content_models/pull/408 to pass and be merged for the tests here to pass.

test/unit/presenters/formats/completed_transaction_presenter_test.rb
```diff
@@ -0,0 +1,82 @@
...
+      assert_equal expected, result[:details][:promotion]
+    end
+
+    should "no [:promotion]" do
+      edition.presentation_toggles["promotion_choice"] = {
+        'choice' => 'none',
+        'url' => ''
+      }
+
+      assert_equal nil, result[:details][:promotion]
```
davidbasalla: I could have sworn I've seen the linter complain about this, prefering `assert_nil`. Maybe I'm imagining things... 😄

davidbasalla:
This should now add a `document_type` method to the Presenter, see https://github.com/alphagov/publisher/pull/581

emmabeynon:
@davidbasalla We've added in the `document_type` method to the presenter now, thanks for the heads up.

Gemfile
```diff
@@ -10,7 +10,13 @@ gem 'gds-sso', '~> 11.2'
 gem 'gds-api-adapters', '~> 40.1.0'
 gem 'govspeak', '~> 3.4.0'
 gem 'govuk_admin_template', '4.2.0'
-gem "govuk_content_models", "~> 42.0.1"
+
```
davidbasalla: I think we've kind of ping-pong-ed on this before with adding/removing the dev env stuff. I think we agreed to not include it in the Gemfile and just do it locally if we have to for local development purposes...

PR #577, opened 2017-02-28
Add fact check token to preview link for preludial states

cbaines:
Users of Publisher want to provide access to the draft content, even
when its not in the fact check state. Therefore, add the fact check
token to the Preview button link.

As the preview link now includes the token, remove the
fact_check_edition_path function and changes uses of it to use
preview_edition_path.

Alter the model to allow for the generation of the fact check id
regardless of the state.

Note that the Publishing API currently strips the access limited
information when sending published content to the draft content
store. This means that the token won't have any effect, therefore to
avoid confusion, don't include the token if this is the case.

https://trello.com/c/LTxUDSbU/661-enable-preview-links-for-external-users-for-draft-content

### DIFF
--- a/app/helpers/paths_helper.rb
+++ b/app/helpers/paths_helper.rb
@@ -3,21 +3,16 @@ module PathsHelper
     "#{preview_url(edition)}/#{edition.slug}"
   end
 
-  def fact_check_edition_path(edition)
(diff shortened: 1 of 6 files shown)


davidbasalla:
Looks good, would be good to poke at in Integration!

PR #579, opened 2017-03-01
Migrate Place

cbaines:
Migrate the Place format, such that it is rendered through the Publishing API.

Do not merge yet, as the merging/release of this needs to be coordinated with Frontend.

### DIFF
--- a/app/presenters/formats/place_presenter.rb
+++ b/app/presenters/formats/place_presenter.rb
@@ -1,0 +1,32 @@
+module Formats
+  class PlacePresenter < EditionFormatPresenter
+  private
+
+    def schema_name
+      'place'
+    end
+
+    def document_type
+      'place'
+    end
+
+    def details
+      details = { external_related_links: external_related_links }
+
+      details[:place_type] = edition.place_type if edition.place_type
+
+      %i(introduction more_information need_to_know).each do |field|
+        next if edition[field].blank?
+
+        details[field] = [
+          {
+            content_type: "text/govspeak",
+            content: edition[field]
+          }
+        ]
+      end
+
+      details
+    end
+  end
+end
--- a/test/unit/presenters/formats/place_presenter_test.rb
+++ b/test/unit/presenters/formats/place_presenter_test.rb
@@ -1,0 +1,70 @@
+require 'test_helper'
+
+class PlacePresenterTest < ActiveSupport::TestCase
+  include GovukContentSchemaTestHelpers::TestUnit
+
+  def subject
+    Formats::PlacePresenter.new(edition)
+  end
+
+  def edition
+    @_edition ||= FactoryGirl.create(:place_edition, panopticon_id: artefact.id)
+  end
+
+  def artefact
+    @_artefact ||= FactoryGirl.create(:artefact, kind: "place", slug: "find-food")
+  end
+
+  def result
+    subject.render_for_publishing_api
+  end
+
+  should "be valid against schema" do
+    assert_valid_against_schema(result, 'place')
+  end
+
+  should "[:schema_name]" do
+    assert_equal 'place', result[:schema_name]
+  end
+
+  context "[:details]" do
+    should "[:introduction]" do
+      edition.update_attribute(:introduction, 'foo')
+      expected = [
+        {
+          content_type: 'text/govspeak',
+          content: 'foo'
+        }
+      ]
+      assert_equal expected, result[:details][:introduction]
+    end
+
+    should "[:external_related_links]" do
+      link = { 'url' => 'www.foo.com', 'title' => 'foo' }
+      artefact.update_attribute(:external_links, [link])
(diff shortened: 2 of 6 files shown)


davidbasalla:
This should now add a `document_type` method to the Presenter, see https://github.com/alphagov/publisher/pull/581

app/presenters/formats/place_presenter.rb
```diff
@@ -0,0 +1,42 @@
...
+      'place'
+    end
+
+    def details
+      details = { external_related_links: external_related_links }
+
+      details = details.merge(govspeak_fields)
+
+      if @edition.place_type
+        details = details.merge(place_type: @edition.place_type)
```
whoojemaflip: there's an attribute reader for the `@edition` variable, so we can lose the snail.

test/unit/presenters/formats/place_presenter_test.rb
```diff
@@ -0,0 +1,65 @@
+require 'test_helper'
```
whoojemaflip: could this be in the same commit as the class under test?

whoojemaflip:
This could be squashed

app/presenters/formats/place_presenter.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def document_type
+      'place'
+    end
+
+    def details
+      details = { external_related_links: external_related_links }
+
+      details = details.merge(govspeak_fields)
```
whoojemaflip: It's a little hard to follow what keys this adds to the details hash, since the implementation is scattered through the whole class.  Could it be structured so that the keys are inline in this method to make it more obvious what it returns?

app/presenters/formats/place_presenter.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      if edition.place_type
+        details = details.merge(place_type: edition.place_type)
```
whoojemaflip: The `merge` method creates a new hash, copies the combined contents into it, then we're assigning that new hash to the same local variable. 

Seems it might make more sense to just modify the existing hash in-place.

`details[:place_type] = edition.place_type`

app/presenters/formats/place_presenter.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    def govspeak_fields
+      %i(introduction more_information need_to_know).inject({}) do |details, field|
+        next details if edition[field].blank?
+
+        details[field] = [
+          {
+            content_type: "text/govspeak",
+            content: edition[field]
```
whoojemaflip: I'm a bit confused by the `[]` operator here - is it common to access properties of the `Edition` model using the hash accessor?

whoojemaflip: If so, it's a neat trick for working with the dynamic property thing.

app/presenters/formats/place_presenter.rb
cbaines: I wrote this, but I think @brenetic improved it as you describe :)

test/unit/presenters/formats/place_presenter_test.rb
cbaines: Done.

test/unit/presenters/formats/place_presenter_test.rb
```diff
@@ -0,0 +1,65 @@
...
+  def artefact
+    @_artefact ||= FactoryGirl.create(:artefact, kind: "place", slug: "find-food")
+  end
+
+  def result
+    subject.render_for_publishing_api
+  end
+
+  should "be valid against schema" do
+    puts result
```
cbaines: Yep, removed :)

app/presenters/formats/place_presenter.rb
cbaines: I've changed it to hopefully make it easier to read, let me know what you think.

app/presenters/formats/place_presenter.rb
cbaines: I believe this is a Mongoid thing.

whoojemaflip:
WFM :sheep: :it:

davidbasalla:
We've previously added a test to check that `#migrated?` returns true for each migrated format here `https://github.com/alphagov/publisher/blob/master/test/unit/edition_test.rb#L75-L110)` so might be good to add for consistency?

davidbasalla:
Same here https://github.com/alphagov/publisher/blob/cefe2eecde3f6b788638323825cd9d2821804904/test/unit/services/edition_presenter_factory_test.rb

cbaines:
> We've previously added a test to check that #migrated? returns true for each migrated format here so might be good to add for consistency?

Yep, I've done this now :)

davidbasalla:
Looks good, merge when ready! 👍

PR #581, opened 2017-03-01
Fix document types for content store & rummager

tijmenb:
- Use the artefact "kind" as document type. For non-migrated editions we're currently sending the schema name (`generic_with_external_related_links`) to the content store as document type. This is affecting the analytics gathering, which uses document type.

  This separates the two concepts by having a separate method for the type. This will make sure we won't confuse the two anymore.

  As noted in #566, `@artefact.kind` isn't always the correct type (the kind doesn't change
if the format changes). This will be solved by migrating all the documents. In the meantime, the `kind` is preferable to `generic_with_external_related_links`.

- Send the same data to rummager as content-store. This makes sure that the rendering_app, publishing_app and content_store_document_type are the same in rummager and content store. Currently we're sending "publisher" as rendering app wrongly. We're also sending the "kind" as document type, which is not always true.

cc @davidbasalla

### DIFF
--- a/app/presenters/formats/answer_presenter.rb
+++ b/app/presenters/formats/answer_presenter.rb
@@ -6,6 +6,10 @@ module Formats
       'answer'
     end
 
+    def document_type
+      'answer'
+    end
+
     def details
       {
         body: [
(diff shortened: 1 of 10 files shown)


davidbasalla:
Looks good, this fixes the described issues!

PR #582, opened 2017-03-06
Migrate Transaction

whoojemaflip:
Adds the TransactionPresenter for migrating Transaction content to the new publishing-platform.

The second commit contains a solution to the Downtime problem, specifically working around the previous implementation's reliance on content-api logic and the shared mongo database.  We had to carefully engineer around the publishing-api's workflow - see the commit message for more info.

[Trello](https://trello.com/c/B0WRTEfM/674-migrate-transaction-format-to-rendered)


app/presenters/formats/transaction_presenter.rb
```diff
@@ -0,0 +1,37 @@
...
+      {
+        introductory_paragraph: govspeak(:introduction),
+        will_continue_on: edition.will_continue_on,
+        transaction_start_link: edition.link,
+        more_information: govspeak(:more_information),
+        other_ways_to_apply: govspeak(:alternate_methods),
+        what_you_need_to_know: govspeak(:need_to_know),
+        external_related_links: external_related_links,
+        department_analytics_profile: edition.department_analytics_profile
+      }.delete_if { |_, value| value.nil? }
```
cbaines: :+1: for conciseness

test/integration/downtime_test.rb
```diff
@@ -1,62 +1,113 @@
...
-  teardown do
-    Sidekiq::ScheduledSet.new.clear
+  setup do
+    setup_users
+
+    @edition = FactoryGirl.create(
+      :transaction_edition,
+      :published,
+      title: 'Apply to become a driving instructor',
+      slug: 'foo'
```
issyl0: Could we make this the slugified version of the title?

binaryberry: Thanks, addressed now!

app/presenters/formats/edition_format_presenter.rb
```diff
@@ -62,7 +60,7 @@ def details
     end
 
     def external_related_links
-      edition.artefact.external_links.map do |link|
```
binaryberry: It doesn't really - extracted to a separate commit: 2c7ed93f.

app/services/publishing_api_workflow_bypass_publisher.rb
```diff
@@ -0,0 +1,56 @@
...
+    live_edition.present?
+  end
+
+  def discard_draft
+    Services.publishing_api.discard_draft(artefact.content_id)
+  end
+
+  def republish_currently_live_edition
+    PublishingAPIUpdater.new.perform(live_edition.id)
+    PublishingAPIPublisher.new.perform(live_edition.id)
```
davidbasalla: Should this be `PublishingAPIPublisher.new.perform(edition_id, 'republish')` instead? Really not sure on this but wondering if a 'normal' publish would trigger some sort of commpunication/emails/fanfares somewhere?

whoojemaflip: it defaults to 'minor' which shouldn't trigger notifications, but you are right I think this should be 'republish'.

PR #583, opened 2017-03-09
Set the phase appropriately if the content is in beta

cbaines:


davidbasalla:
Great! Looks like we'll have to republish our migrated content...

PR #585, opened 2017-03-14
Untangle some service/worker responsibility confusion

whoojemaflip:
Services enqueueing worker threads! Workers containing API call logic! It's Madness!

The new paradigm is much simpler: 
 - Service classes are called synchronously, and wrap potentially complex business logic into a single operation.
 - Worker classes contain only scheduling logic and delegate any actual work to Service objects.

Now that there is some clarity of function, some class names have been updated to be clearer as to what task they perform.

All the synchronous calls to a Worker object's `perform` method have been replaced with calls to the relevant Service class instead. The PublishService was enqueuing API operations to run on a background thread, so calls to that class have been replaced with calls to a PublishWorker. 

Finally, there was an erroneous `put-content` call to the publishing-api in the publish service class. This has now been removed.

[Trello](https://trello.com/c/AeiDvHYz/691-code-tidy-resolve-service-worker-confusion-in-publisher)


cbaines:
What does "SRP" in the first commit message stand for?

whoojemaflip:
@cbaines It's the `S` in [SOLID](https://en.wikipedia.org/wiki/SOLID_(object-oriented_design))

davidbasalla:
CI is unhappy

whoojemaflip:
This is rebased on top of https://github.com/alphagov/publisher/pull/582 so it'll need that and https://github.com/alphagov/govuk-content-schemas/pull/552 to be merged before it can be merged into master.

whoojemaflip:
I've updated the base-branch to be master, so this is merge-able once CI blesses the two dependant PRs.

PR #586, opened 2017-03-15
Retire video format

emmabeynon:
- [x] Awaiting copy change for retired message from Ben Hazell

The video format is being retired.  This PR deals with removing users' ability to create video artefacts and editions, as well as converting other formats in to videos, in Publisher.  We also display a message to users if they are viewing a retired format:

image

And we indicate that a format is retired in the format selection dropdown menu:
screen shot 2017-03-16 at 10 31 41

This PR is dependent on https://github.com/alphagov/frontend/pull/1173

Trello card: https://trello.com/c/tTF5fAxc/684-retire-video-format-3
Paired with @cbaines


app/views/videos/_fields.html.erb
```diff
@@ -1,36 +0,0 @@
-<%= f.inputs do %>
```
davidbasalla: What impact does removing this file have? Can we still see the details for a video edition on the show page?

emmabeynon: We removed this and then reinstated it...not sure why it's still showing as removed.  Will try again!

davidbasalla:
Would it be possible/easy to add an integration test to check that the 'Create Edition' button does _not_ appear on the index page for Videos?

We could also update the Video related text on the Readme (maybe just say it's been retired), the present link is already questionable since it redirects to another format...

emmabeynon:
@davidbasalla 
> Would it be possible/easy to add an integration test to check that the 'Create Edition' button does not appear on the index page for Videos?
Now added.

> We could also update the Video related text on the Readme (maybe just say it's been retired), the present link is already questionable since it redirects to another format...
I've removed the link and added a retired formats section.

issyl0:
🎉

PR #588, opened 2017-03-15
Add a test:schemas rake task

tijmenb:
This rake task only runs tests that have the word "schema" in them. We can run this task instead of the full test (which takes 8 minutes) when testing as a downstream build from govuk-content-schemas.

Having "schema" in the test file is obviously going to cause false positives, but hopefully no false negatives - which would be worse.

This test run takes 10s to complete.


cbaines:
I'm going to have a quick look at how this affects the effectiveness of the tests, also, the name of the task in the commit message is wrong.

tijmenb:
@cbaines updated.

PR #590, opened 2017-03-15
Don't let users preview archived editions

issyl0:
- For archived content, Publisher shouldn't link to the draft content
  store because it might be inaccurate (and show a published edition).
  So, disable the preview button, remove the link on the slug on the
  index page, and remove the preview link on the history and notes page.
- This relies on a change in content_models to the `GuideEdition` and
  `Edition` `safe_to_preview?` method to make it false for archived
  content.

https://trello.com/c/5d86KxiY/685-make-ui-changes-in-publisher-for-previewing-published-and-archived-editions-2

Before:

screenshot 2017-03-15 17 10 35

screenshot 2017-03-15 17 10 58

screenshot 2017-03-15 17 11 25

After:

screenshot 2017-03-15 17 08 33

screenshot 2017-03-15 17 07 49

screenshot 2017-03-15 17 08 00

### DIFF
--- a/app/helpers/edition_activity_buttons_helper.rb
+++ b/app/helpers/edition_activity_buttons_helper.rb
@@ -54,6 +54,8 @@ module EditionActivityButtonsHelper
   def preview_button(edition)
     if edition.published?
       link_to('View this on the GOV.UK website', "#{Plek.new.website_root}/#{edition.slug}", class: 'btn btn-primary btn-large')
+    elsif edition.archived?
+      link_to('Preview', '#', class: 'btn btn-primary btn-large disabled')
     else
       link_to('Preview', preview_edition_path(edition), class: 'btn btn-primary btn-large')
     end
--- a/app/views/root/_publication.html.erb
+++ b/app/views/root/_publication.html.erb
@@ -14,8 +14,6 @@
       <%= link_to "/#{publication.slug}", "#{Plek.new.website_root}/#{publication.slug}", class: 'link-muted' %>
     <% elsif publication.safe_to_preview? %>
       <%= link_to "/#{publication.slug}", preview_edition_path(publication), class: 'link-muted' %>
-    <% else %>
-      <span class="text-muted"><%= publication.slug %></span>
     <% end %>
 
     <span class="text-muted"> &middot; <span title="Edition <%= publication.version_number %>">#<%= publication.version_number %></span></span>
--- a/app/views/shared/_history.html.erb
+++ b/app/views/shared/_history.html.erb
@@ -67,7 +67,7 @@
     <% elsif resource.safe_to_preview? %>
       Preview edition at <%= link_to preview_edition_path(resource), preview_edition_path(resource) %>.<br />
     <% else %>
-      This edition can’t be previewed yet.<br />
+      This edition can’t be previewed.<br />
     <% end %>
     Send fact check responses to <%= mail_to resource.fact_check_email_address %>
   </p>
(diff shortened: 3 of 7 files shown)


app/views/root/_publication.html.erb
```diff
@@ -12,7 +12,7 @@
 
     <% if publication.published? %>
       <%= link_to "#{Plek.new.website_root}/#{publication.slug}", class: 'link-muted' %>
-    <% elsif publication.safe_to_preview? %>
+    <% elsif !publication.archived? %>
```
davidbasalla: There seems to be some subtlety for `safe_to_preview?` for GuideEditions (in content models) so you might have to keep it around. Possibly `publication.safe_to_preview? && !publication.archived?` _or_ update the `safe_to_preview?` method in content models to return false for archived content...?

app/views/shared/_history.html.erb
```diff
@@ -64,10 +64,10 @@
   <p class="add-bottom-margin">
     <% if resource.published? %>
       View this on the GOV.UK website <%= link_to "#{Plek.new.website_root}/#{resource.slug}" %>.<br />
-    <% elsif resource.safe_to_preview? %>
+    <% elsif !resource.archived? %>
```
davidbasalla: Same as the other comment regarding `safe_to_preview?`

app/views/root/_publication.html.erb
```diff
@@ -12,7 +12,7 @@
...
       <%= link_to "/#{publication.slug}", preview_edition_path(publication), class: 'link-muted' %>
     <% else %>
       <span class="text-muted"><%= publication.slug %></span>
```
issyl0: Now, if the publication is archived, this will just display (as per the screenshot) the plain slug (without linking to anything). Should the slug disappear entirely in this case?

app/views/root/_publication.html.erb
issyl0: Oh, thanks. I only looked at the Edition `safe_to_preview?`, I didn't realise GuideEditions were special. In this case, to avoid getting too deep into content-models, I'll just do the former: `&& !publication.archived?`.

issyl0: OK, so I looked at `content_models` and it wasn't that bad (and probably best in the long run), so here's https://github.com/alphagov/govuk_content_models/pull/417.

app/views/root/_publication.html.erb
issyl0: The answer from product review was "yes". I've removed this now.

app/views/root/_publication.html.erb
issyl0: I reverted the removal of `safe_to_preview?`, but it won't work until that content_models PR is merged, a new gem version released, and a gem version bump in here.

app/helpers/edition_activity_buttons_helper.rb
```diff
@@ -54,6 +54,8 @@ def publish_button(edition)
   def preview_button(edition)
     if edition.published?
       link_to('View this on the GOV.UK website', "#{Plek.new.website_root}/#{edition.slug}", class: 'btn btn-primary btn-large')
+    elsif edition.archived?
+      link_to('Preview', preview_edition_path(edition), class: 'btn btn-primary btn-large disabled')
```
cbaines: It might be neater if the path for this disabled button was just `#`?

issyl0: Yeah, that's definitely neater. Amended.

PR #591, opened 2017-03-16
Stop sending JSON as part of routes payload

emmabeynon:
Frontend's JSON API is being retired (see
https://github.com/alphagov/frontend/pull/1180) so we remove code that sends
JSON to Frontend as part of the routes payload.

Trello: https://trello.com/c/7IRxAVoJ/688-retire-frontend-s-json-api

### DIFF
--- a/test/unit/presenters/formats/completed_transaction_presenter_test.rb
+++ b/test/unit/presenters/formats/completed_transaction_presenter_test.rb
@@ -75,7 +75,6 @@ class CompletedTransactionPresenterTest < ActiveSupport::TestCase
     edition.update_attribute(:slug, 'foo')
     expected = [
       { path: '/foo', type: 'prefix' },
-      { path: '/foo.json', type: 'exact' }
     ]
     assert_equal expected, result[:routes]
   end
--- a/app/presenters/formats/edition_format_presenter.rb
+++ b/app/presenters/formats/edition_format_presenter.rb
@@ -71,16 +71,11 @@ module Formats
     def routes
       [
         { path: "#{base_path}", type: path_type },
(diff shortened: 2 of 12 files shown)


test/unit/presenters/formats/completed_transaction_presenter_test.rb
```diff
@@ -75,7 +75,6 @@ def result
     edition.update_attribute(:slug, 'foo')
     expected = [
       { path: '/foo', type: 'prefix' },
```
whoojemaflip: is this trailing comma signifiant?

cbaines: I don't believe so, as far as I understand ruby, trailing commas are optional. I think the GDS styleguide even encourages it https://github.com/alphagov/styleguides/blob/master/ruby.md

davidbasalla: It's about reducing github diff noise when editing arrays right?

whoojemaflip: I concur.

whoojemaflip: If it's worth doing, for the sake of neatness it might be worth keeping the trailing comma in the other examples then..

emmabeynon: Honestly I just missed deleting the comma, but I can them in to the other examples if required.

emmabeynon: Now all examples have the trailing comma ⚡️

davidbasalla:
Looks good 👍

PR #592, opened 2017-03-16
Rename fact_check_id to auth_bypass_id

binaryberry:
We want to be able to give people access to the draft stack for other states
than just fact check (i.e. draft or ready) so we rename the field to be less
specific.

Do not merge until https://github.com/alphagov/govuk-content-schemas/pull/541 and other PRs from https://trello.com/c/7oRBYwnn/690-rename-fact-check-id-to-auth-bypass-id are ready.


app/helpers/paths_helper.rb
```diff
@@ -52,9 +52,9 @@ def preview_url(edition)
 
 private
 
-  def should_have_fact_check_id?(edition)
+  def should_have_auth_bypass_id?(edition)
     edition.migrated? &&
     %w(published archived).exclude?(edition.state) &&
-    edition.fact_check_id
+    edition.auth_bypass_id
```
cbaines: I think the linter is unhappy here.

PR #587, opened 2017-03-15
View published editions on origin production, not draft

issyl0:
- It makes no sense for users to be directed to draft-origin to view
  already published editions, so link them to "View this on the GOV.UK
  website" corresponding to the environment they're in.

Trello: https://trello.com/c/5d86KxiY/685-make-ui-changes-in-publisher-for-previewing-published-and-archived-editions-2

Before:
screenshot 2017-03-15 13 18 50

screenshot 2017-03-15 12 46 49

After:

screenshot 2017-03-15 13 40 31

screenshot 2017-03-15 13 49 02

PR #593, opened 2017-03-17
Fix the `link_to` parameters for viewing published editions on GOV.UK

issyl0:
- I made a mistake in #587 - `link_to` didn't have enough arguments, so it was showing the entire URL as link text.

Before:

screenshot 2017-03-17 15 03 36

After:

screenshot 2017-03-17 15 02 53

### DIFF
--- a/app/views/root/_publication.html.erb
+++ b/app/views/root/_publication.html.erb
@@ -11,7 +11,7 @@
     </h4>
 
     <% if publication.published? %>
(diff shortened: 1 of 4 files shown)


davidbasalla:
Looks like this wasn't caught in the tests so can we add some that checks the actual link? Would also be nice if we could add a test for the link on the index page.

issyl0:
@davidbasalla Thanks. I added some tests for link text and URL.

PR #594, opened 2017-03-17
Add maxitest gem

binaryberry:
This allows to run a single test easily by specifying the line number, using
`bundle exec mtest -I test path_to_test_file.rb:56` to run the test that is on
line 56.

Full list of features: https://github.com/grosser/maxitest#features


davidbasalla:
That's pretty cool, I'm up for it!

PR #595, opened 2017-03-20
Move Services module to /app/lib

whoojemaflip:
Sidekiq workers were having trouble loading this class because class
autoloading is disabled for worker threads and /lib isn't configured
to be eager_loaded.

Rather than configure eager loading for /lib and have to load all the
rake tasks in production, I've moved the Services module to an
autoloadable path.


davidbasalla:
👍  merge when green

PR #598, opened 2017-03-22
Add ability to switch the rendering app depending on format

mcgoooo:
this is part of the work to consolidate applications and moves
the answers format to government frontend. the frontend work
has been done for this.

### DIFF
--- a/app/presenters/formats/answer_presenter.rb
+++ b/app/presenters/formats/answer_presenter.rb
@@ -10,6 +10,10 @@ module Formats
       'answer'
     end
 
+    def rendering_app
+      "government-frontend"
+    end
+
     def details
       {
         body: [
--- a/app/presenters/formats/edition_format_presenter.rb
+++ b/app/presenters/formats/edition_format_presenter.rb
@@ -37,7 +37,7 @@ module Formats
         need_ids: [],
         public_updated_at: public_updated_at.to_datetime.rfc3339(3),
         publishing_app: "publisher",
-        rendering_app: "frontend",
+        rendering_app: rendering_app,
         routes: routes,
         redirects: [],
         update_type: update_type(republish),
@@ -57,6 +57,10 @@ module Formats
 
     def details
       {}
+    end
+
+    def rendering_app
+      "frontend"
     end
 
     def external_related_links
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -1,9 +1,11 @@
 namespace :publishing_api do
+  desc 'republish content'
   task republish_content: [:environment] do
     republish Edition.published
     republish_draft Edition.draft_in_publishing_api
   end
 
+  desc 'republish by format'
   task :republish_by_format, [:format] => :environment do |_, args|
     format_editions = Edition.by_format(args[:format])
 
@@ -11,6 +13,7 @@ namespace :publishing_api do
     republish_draft format_editions.draft_in_publishing_api
   end
 
+  desc 'republish drafts by format'
   task :republish_drafts_by_format, [:format] => :environment do |_, args|
     format_editions = Edition.by_format(args[:format])
 
--- a/test/unit/presenters/formats/answer_presenter_test.rb
+++ b/test/unit/presenters/formats/answer_presenter_test.rb
@@ -60,4 +60,8 @@ class AnswerPresenterTest < ActiveSupport::TestCase
     ]
     assert_equal expected, result[:routes]
   end
+
+  should "[:rendering_app]" do
+    assert_equal "government-frontend", result[:rendering_app]
+  end
 end


lib/tasks/publishing_api.rake
```diff
@@ -1,16 +1,19 @@
...
 
+  desc 'republish by format'
   task :republish_by_format, [:format] => :environment do |_, args|
     format_editions = Edition.by_format(args[:format])
 
     republish format_editions.published
     republish_draft format_editions.draft_in_publishing_api
   end
 
+  desc 'republish drafts by reformat'
```
whoojemaflip: typo `reformat` => `format`

test/unit/presenters/formats/edition_format_presenter_test.rb
```diff
@@ -171,5 +172,15 @@ def result
         refute result.has_key?(:access_limited)
       end
     end
+
+    should "[:format]" do
```
whoojemaflip: this is poorly named, as it's testing the `rendering_app` attribute of the rendered hash.

whoojemaflip: also, there's a test for the `rendering_app` attribute above on line 103

test/unit/presenters/formats/edition_format_presenter_test.rb
```diff
@@ -171,5 +172,15 @@ def result
...
+      edition.expects(:format).returns('foo')
+      assert_equal 'frontend', result[:rendering_app]
+    end
+
+    should "[:format_gf]" do
```
whoojemaflip: the convention in this test class is to use `context` blocks to identify behaviours based on alternative data scenarios.  However, this case should probably be in the `answer_presenter_test` class under `rendering_app`

app/presenters/formats/edition_format_presenter.rb
```diff
@@ -37,7 +41,7 @@ def required_fields(republish)
         need_ids: [],
         public_updated_at: public_updated_at.to_datetime.rfc3339(3),
         publishing_app: "publisher",
-        rendering_app: "frontend",
+        rendering_app: RENDERING_APP.fetch(edition.format.try(:to_sym), "frontend"),
```
whoojemaflip: I'd like to see this functionality rewritten as a ['default' value in this base class](https://github.com/alphagov/publisher/blob/master/app/presenters/formats/edition_format_presenter.rb#L54), and overridden in the [child class](https://github.com/alphagov/publisher/blob/master/app/presenters/formats/answer_presenter.rb) as appropriate. 

I'm happy to pair on it with you if that will help.

app/presenters/formats/answer_presenter.rb
```diff
@@ -1,5 +1,10 @@
 module Formats
   class AnswerPresenter < EditionFormatPresenter
+
+  def render_for_publishing_api(republish: false, rendering_app: "government-frontend")
```
whoojemaflip: I was thinking more along the lines of there being a `rendering_app` method in this class, which would override the same method on the super class.
eg.
```
def rendering_app
  'government-frontend'
end
```

app/presenters/formats/edition_format_presenter.rb
```diff
@@ -1,12 +1,13 @@
...
   class EditionFormatPresenter
+
     def initialize(edition)
       @edition = edition
       @artefact = edition.artefact
     end
 
-    def render_for_publishing_api(republish: false)
-      required_fields(republish).merge optional_fields
+    def render_for_publishing_api(republish: false, rendering_app: "frontend")
```
whoojemaflip: there's no need to change this method

app/presenters/formats/edition_format_presenter.rb
```diff
@@ -27,7 +28,7 @@ def optional_fields
       fields
     end
 
-    def required_fields(republish)
+    def required_fields(republish, rendering_app)
```
whoojemaflip: nor to pass in a new parameter here

app/presenters/formats/edition_format_presenter.rb
whoojemaflip: instead, there could easily be a new method in this class with the default rendering_app. eg.:
```
def rendering_app
  'frontend'
end
```

test/unit/presenters/formats/answer_presenter_test.rb
```diff
@@ -60,4 +60,8 @@ def result
     ]
     assert_equal expected, result[:routes]
   end
+
+  should "[:rendering_app]" do
```
whoojemaflip: 👍

test/unit/presenters/formats/edition_format_presenter_test.rb
```diff
@@ -145,6 +145,17 @@ def result
       end
     end
 
+    context "[:rendering_app]" do
```
whoojemaflip: there's already a test for this in this class.

test/unit/presenters/formats/edition_format_presenter_test.rb
```diff
@@ -145,6 +145,17 @@ def result
...
+      should "return 'frontend' by default " do
+        assert_equal "frontend", result[:rendering_app]
+      end
+
+      should "return requested renderer when specified" do
```
whoojemaflip: once we've removed the complexity above, this won't be needed.

app/presenters/formats/edition_format_presenter.rb
```diff
@@ -37,7 +38,7 @@ def required_fields(republish)
         need_ids: [],
         public_updated_at: public_updated_at.to_datetime.rfc3339(3),
         publishing_app: "publisher",
-        rendering_app: "frontend",
+        rendering_app: rendering_app,
```
whoojemaflip: 👍

app/presenters/formats/edition_format_presenter.rb
```diff
@@ -1,5 +1,6 @@
 module Formats
   class EditionFormatPresenter
+
```
whoojemaflip: remove the extra space 👾

whoojemaflip:
`edition_format_presenter_test` needs a bit of a tidy up, but 👍

test/unit/presenters/formats/edition_format_presenter_test.rb
```diff
@@ -58,6 +58,10 @@ def result
       assert_equal "override me", result[:schema_name]
     end
 
+    should "[:rendering_app]" do
```
whoojemaflip: This is still [duplicated](https://github.com/alphagov/publisher/blob/e053089c496a6074c54cd30163561f4be84717c6/test/unit/presenters/formats/edition_format_presenter_test.rb#L106) in this class

test/unit/presenters/formats/edition_format_presenter_test.rb
```diff
@@ -54,8 +54,8 @@ def result
       end
     end
 
-    should "[:schema_name]" do
-      assert_equal "override me", result[:schema_name]
+    should "[:rendering_app]" do
```
whoojemaflip: one more time, for luck... 😆

test/unit/presenters/formats/answer_presenter_test.rb
```diff
@@ -60,4 +60,10 @@ def result
     ]
     assert_equal expected, result[:routes]
   end
+
+  context "[:rendering_app]" do
```
whoojemaflip: no need for this `context` block if there's only a single assertion.

PR #605, opened 2017-03-27
Add rake task for discarding drafts with schema_name "placeholder"

cbaines:
These are leftover data from the migration process to the Publishing API.

### DIFF
--- a/lib/tasks/bulk_discard_placeholder_draft_content.rake
+++ b/lib/tasks/bulk_discard_placeholder_draft_content.rake
@@ -1,0 +1,20 @@
+task bulk_discard_placeholder_draft_content: :environment do
+  # Result of running ContentItem.where(schema_name: "placeholder", publishing_app: "publisher").pluck(:base_path, :content_id)
+  # in the draft content store, paths included for some readability.
+  [
+    ["/change-vehicle-tax-rate", "3fd872e6-0b51-4a13-be5f-f67da80ecc5d"],
+    ["/done/deposit-foreign-marriage", "2f0812ec-9671-4851-9d46-6fa6fb3985e1"],
+    ["/diffuse-mesothelioma-payments", "ff352e2b-1269-4ab6-8b12-ea9040b25951"],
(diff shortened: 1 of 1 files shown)


lib/tasks/bulk_discard_placeholder_draft_content.rake
```diff
@@ -0,0 +1,20 @@
...
+    ["/erdf-low-carbon-grant", "36ef73e9-b2ec-4631-b837-3643bbef348f"],
+    ["/find-vehicle-scrap-yard", "63d42a67-83d7-4f1b-a56c-e91a7af3b02a"],
+    ["/graduate-recruitment-grap-programme", "4355fdc9-833e-44c9-949c-66cf4a704178"],
+    ["/low-carbon-building-project", "8b883728-1ec8-4383-b8fb-981cbba71c82"],
+    ["/ready-for-business", "4a159e85-6848-492c-8f2e-28d24e9989eb"],
+    ["/rural-payments", "82cb2e94-769f-4545-96f8-84c8010c7e04"],
+    ["/sustainabuild", "2aefe932-8abf-4a4a-9ae4-468fd09d3d83"],
+    ["/work-after-state-pension-age", "effa2e51-352e-4a4d-b974-c881283633c6"]
+  ].each do |(_, content_id)|
+    Services.publishing_api.discard_draft(content_id)
```
binaryberry: Am I right in thinking these drafts are near duplicates as an identical version of them exists in the content store, with a schema name of a migrated format? If so, could the task include another Publishing API call to check that the draft exists with a schema name of a migrated format in the content-store, before deleting it?

cbaines: `/rural-payments` is the only base_path in the live content store for which something exists, and in that case its a redirect (from a unpublished edition in the Publishing API).

binaryberry: 👍

PR #609, opened 2017-03-29
Remove references to migrated formats

binaryberry:
All publisher formats will have been migrated, so this code is no longer
necessary.


issyl0:
I don't think you meant to commit `rubocop-e6d96e5.html` in 6ebd770.

binaryberry:
Argh, indeed! Fixed now, thanks.

And I thought I was being all smart running Rubocop locally before pushing to github. I will know about this side effect in the future.

PR #612, opened 2017-03-31
Stop sending paths/prefixes to Rummager

sihugh:
Rummager doesn't care about paths and prefixes so there's little point in
continuing to send them.  We still send the `link`.

Any fields that are not part of [the schema for the type][1] we're sending [will be silently ignored][2]

Also adding the linter to the default rake task so that we pick up silly things that stop us merging PRs more quickly.

[1]: https://github.com/alphagov/rummager/blob/master/config/schema/base_elasticsearch_type.json
[2]: https://github.com/alphagov/rummager/blob/master/docs/documents.md#post-indexdocuments

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,4 +10,4 @@ Rake.application.options.trace = true
 
 Rails.application.load_tasks
 
-task :default => [:test, 'jasmine:ci']
+task :default => [:lint, :test, 'jasmine:ci']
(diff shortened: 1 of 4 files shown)


sihugh:
Merging for the moment - we'll look into this further after the blitz.

Tracking it initially in this card https://trello.com/c/oXaNtveI/49-publisher-clean-up

PR #615, opened 2017-04-07
Fix bug with downtime messages not being published when no draft present

fofr:
`%w(published, archived)` produces: 
`["published, ", "archived"]` rather than the expected:
`["published", "archived"]`

* Use `%w(published archived)`
  We should have some rubocop protection against using `,` within `%w`

This fix prevents unnecessary calls to `discard_draft` which were throwing an error because no draft existed. This error stopped the downtime message from being published:
https://github.com/alphagov/publisher/blob/fix-downtime/app/services/publishing_api_workflow_bypass_publisher.rb#L14

Error that was being thrown:
```
Response body: {"error":{"code":422,"message":"There is not a draft
edition of this document to discard"}} Request body: {}
```
This error was also masked by the way `discard_draft` was being stubbed on an instance of publishing_api. Removing the stub revealed the failing test.

I'm still not entirely sure why the stubbed method in the test was masking that the method was being called: https://github.com/alphagov/publisher/blob/fix-downtime/test/unit/services/publishing_api_workflow_bypass_publisher_test.rb#L33

Zen: https://govuk.zendesk.com/agent/tickets/1963886

cc @whoojemaflip @cbaines

### DIFF
--- a/app/services/publishing_api_workflow_bypass_publisher.rb
+++ b/app/services/publishing_api_workflow_bypass_publisher.rb
@@ -49,7 +49,7 @@ private
   def draft_edition
     Edition
       .where(panopticon_id: artefact.id)
-      .where(state: { "$nin" => %w(published, archived) })
+      .where(state: { "$nin" => %w(published archived) })
       .first
   end
 end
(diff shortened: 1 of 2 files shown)


fofr:
Corresponding linting change to avoid this happening again: https://github.com/alphagov/govuk-lint/pull/71

PR #621, opened 2017-05-04
Add explicit date to fact check deadline message

h-lame:
The default text for the email that is sent when requesting fact check had a "Deadline: 5 working days" added in #620.  This PR extends that to include actual dates.  We need to get a list of public holidays, so we're using the new `GdsApi::Calendars` adapter to read them from https://www.gov.uk/bank-holidays.json.

Relies on https://github.com/alphagov/gds-api-adapters/pull/702

For: https://trello.com/c/VSg9iWPO/156-factcheck-email-include-the-date-in-email

Text before:

```
**Deadline: 5 working days**
```

Text now:

```
**Deadline: 8 May 2017 (5 working days from today - 27 April 2017)**
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ gem 'erubis'
 gem 'formtastic', '2.3.0'
 gem 'formtastic-bootstrap', '3.0.0'
 gem 'gds-sso', '~> 11.2'
-gem 'gds-api-adapters', '~> 41.0.0'
+gem 'gds-api-adapters', '45.0.0'
 gem 'govspeak', '~> 3.4.0'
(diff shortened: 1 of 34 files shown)


h-lame:
Flagged as DNM because we need to release a new version of the gds-api-adapters gem and make this PR use that version instead of a git reference.

keithiopia:
34 files changes?! Good gravy!
Anyway, LGTM. Thanks a lot!

h-lame:
Removed DNM flag as the version of gds-api-adapters (v 45.0.0) with the calendars adapter has been released.

PR #623, opened 2017-05-08
Add migration to remove Whitehall data

davidbasalla:
Whitehall artefacts were sent to the shared database for tagging in Panopticon.
Panopticon has been retired, so this data is not needed anymore.

This reduces the count in the 'artefacts' collection in the shared database from 208586 to 4922.

I've checked that that this data can be removed with a few people, but please shout if you see any problems with this. The original data is still in Whitehall, and tagging has migrated from Panopticon to Content Tagger.

Also wasn't sure whether Publisher was the best place to add this migration, but seemed like the best access to the shared database.

Related to https://github.com/alphagov/govuk_content_models/pull/435


rubenarakelyan:
Looks good to me, but I don't have the background to say whether this will affect anything else

PR #625, opened 2017-05-17
Fix linter issues in the entire project

carvil:
There are inconsistencies between running the linter locally or on
Jenkins. Sometimes on Jenkins it reports problems on files not edited,
which is not great.

In order to get around that problem, I've fixed all linter errors on
this project in 2 ways:

- first I can the linter with `-a` option and let it fix most of the
  issues automatically;
- finally, I manually fixed the remaining 50 issues.

This will hopefully help reducing the noise going forward.


app/workers/scheduled_publisher.rb
```diff
@@ -2,10 +2,10 @@ class ScheduledPublisher
   include Sidekiq::Worker
 
   # 5 retries over 10 mins
-  sidekiq_options :retry => 5
+  sidekiq_options retry: 5
   sidekiq_retry_in do |count|
     # 16s, 31s, 96s, 271s, 640s
-    count ** 4 + 15
+    count**4 + 15
```
Davidslv: is that the preference for an exponential?

Davidslv:
@carvil I've read all the changes and it seems good to me. Happy to merge this for you

app/workers/scheduled_publisher.rb
carvil: This was one of the automatic fixes. That is the preferred way according to https://github.com/bbatsov/ruby-style-guide#spaces-operators

PR #629, opened 2017-06-08
Add rake tasks for claiming & drafting calculator start page (a transaction)

fofr:
Move the start page for the child tax benefit calculator from calculators to Publisher/Frontend.
Part of https://trello.com/c/rz7jFckZ/

This will allow content designers to make changes to the content and will make the page consistent with other start pages.

* The path needs to be claimed so that Publisher can publish to it
  (this will stop Calculators from publishing to that path too)
* Draft a start page with content that's currently live
* Delete any old artefact owned by calculators so new artefact can be created
  * Maintain Need IDs
  * Use existing content_id so that content doesn't need to be unpublished
    (which would risk users not seeing the page)

Steps to switch the start page:
* Merge and deploy: https://github.com/alphagov/calculators/pull/185
* Run `claim` rake task
* Run `draft` rake task
* Preview draft, send to review, skip review then publish

cc @kevindew

### DIFF
--- a/lib/tasks/calculators.rake
+++ b/lib/tasks/calculators.rake
@@ -51,5 +51,8 @@ namespace :calculators do
 
     edition.save!
     puts "Draft edition created: https://publisher.integration.publishing.service.gov.uk/editions/#{edition.id}"
+
+    UpdateWorker.perform_async(edition.id.to_s)
+    puts "Pushing draft to publishing API: https://draft-origin.integration.publishing.service.gov.uk/child-benefit-tax-calculator"
   end
 end
--- a/lib/tasks/calculators.rake
+++ b/lib/tasks/calculators.rake
@@ -1,0 +1,55 @@
+require 'gds_api/publishing_api_v2'
+
+namespace :calculators do
+  desc "Claim /child-benefit-tax-calculator path and allow new transaction to be created"
+  task claim_calculator: :environment do
+    # Claim child benefit tax calculator path from calculators
+    # https://github.com/alphagov/publishing-api/blob/master/doc/api.md#put-pathsbase_path
+    puts "Claiming path /child-benefit-tax-calculator"
+    endpoint = Services.publishing_api.options[:endpoint_url]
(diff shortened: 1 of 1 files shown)


fofr:
Closing until https://github.com/alphagov/calculators/pull/185 is re-opened with fix.

fofr:
Problem with alphagov/calculators#185 fixed, re-opening.

lib/tasks/calculators.rake
```diff
@@ -0,0 +1,57 @@
...
+
+      edition[:link] = link
+      edition[:introduction] = introduction
+      edition[:more_information] = more_information
+      edition[:overview] = overview
+
+      if edition.save
+        puts "Draft edition created: https://publisher.integration.publishing.service.gov.uk/editions/#{edition.id}"
+      end
+    end
```
kevindew: Do you need to catch if either of these fail? you might want to just run `artefact.save!` so it throws an exception and then you can dodge the conditionals.

fofr: Commit updated.

lib/tasks/calculators.rake
```diff
@@ -0,0 +1,55 @@
...
+    link = "/child-benefit-tax-calculator/main"
+    introduction = "Use this tool to work out:\r\n\r\n* how much Child Benefit you receive in a tax year\r\n* the High Income Child Benefit tax charge you or your partner may have to pay"
+    more_information = "You may be affected by the tax charge if your income is over £50,000.\r\n\r\nYour partner is responsible for paying the tax charge if their income is more than £50,000 and higher than yours.\n\nYou’ll need the dates Child Benefit started and, if applicable, [Child Benefit stopped](/child-benefit/eligibility)."
+
+    edition[:link] = link
+    edition[:introduction] = introduction
+    edition[:more_information] = more_information
+    edition[:overview] = overview
+
+    edition.save!
```
kevindew: You should just need to add this in to which hopefully will update draft stack.

UpdateWorker.perform_async(edition.id.to_s)

PR #630, opened 2017-06-19
Rake task to create draft licence finder transaction

fofr:
__Rake task depends on changes to https://github.com/alphagov/licence-finder__

* Replace already used calculators rake task with licence finder one
* Copies content from live [/licence-finder](https://www.gov.uk/licence-finder) page
* The path needs to be claimed so that Publisher can publish to it (this will stop Licence Finder from publishing to that path too)
* Draft a start page with content that's currently live
* Delete any old artefact owned by licencefinder so new artefact can be created
  * Use existing content_id so that content doesn't need to be unpublished (which would risk users not seeing the page)

## Example draft created
screen shot 2017-06-19 at 08 24 14

@binaryberry


lib/tasks/licence_finder.rake
```diff
@@ -0,0 +1,51 @@
+require 'gds_api/publishing_api_v2'
+
+namespace :licence_finder do
+  desc "Create draft start page for licence finder"
+  task licence_finder_draft: :environment do
+    # Claim child benefit tax calculator path from calculators
```
binaryberry: Both comments need to be updated to licence-finder

lib/tasks/licence_finder.rake
```diff
@@ -0,0 +1,51 @@
...
+
+    edition[:link] = link
+    edition[:introduction] = introduction
+    edition[:overview] = overview
+
+    edition.save!
+    puts "Draft edition created: https://publisher.integration.publishing.service.gov.uk/editions/#{edition.id}"
+
+    UpdateWorker.perform_async(edition.id.to_s)
+    puts "Pushing draft to publishing API: https://draft-origin.integration.publishing.service.gov.uk/child-benefit-tax-calculator"
```
binaryberry: url needs to be updated

lib/tasks/licence_finder.rake
```diff
@@ -0,0 +1,51 @@
...
+    link = "/licence-finder/sectors"
+    introduction = "You may need a licence for:\r\n\r\n* some business activities\r\n* other activities, eg street parties\r\n\r\nUse this tool to find out which licences you may need."
+
+    edition[:link] = link
+    edition[:introduction] = introduction
+    edition[:overview] = overview
+
+    edition.save!
+    puts "Draft edition created: https://publisher.integration.publishing.service.gov.uk/editions/#{edition.id}"
+
```
binaryberry: Should we also add tagging data to this, as it determines the related links and breadcrumbs? Or did you want to add it in the UI later?

fofr: Tagging data will automatically be included as tags are associated with paths.

lib/tasks/licence_finder.rake
fofr: Fixed and updated.

lib/tasks/licence_finder.rake
fofr: Fixed and updated. I've also removed the environment part of URL as this will vary.

PR #631, opened 2017-06-20
Incorporate govuk content models

alecgibson:
We recently [retired the `content_api`](https://trello.com/c/QnDthUyD/45-epic-retiring-the-content-api). This was one of two applications still using the [`govuk_content_models` gem](https://github.com/alphagov/govuk_content_models). The other application using this gem was `publisher`.

Since `govuk_content_models` existed purely to share common models between applications, there's no need for it to exist anymore since `publisher` is the only application to use it.

This change incorporates all the models into `publisher`, and removes the dependency on `govuk_content_models`, so that we can retire the gem.

The change is almost entirely a copy-and-paste operation, with the exception of having to fix linter issues (which were done in separate commits).

### Trello

https://trello.com/c/LK5YLDmO/218-move-govukcontentmodels-into-publisher


lib/enhancements/base_processor.rb
```diff
@@ -1,25 +0,0 @@
...
-
-    private
-
-      def make_record_action_noises(new_action, action_name)
-        NoisyWorkflow.make_noise(new_action).deliver_now
-        NoisyWorkflow.request_fact_check(new_action).deliver_now if action_name.to_s == "send_fact_check"
-      end
-    end
-  end
-end
```
h-lame: There are other files in lib/enhancements that could be flattened down into the appropriate models like you've done here with `GovukContentModels::ActionProcessors::BaseProcessor`.

alecgibson: Right yeah I'll have a look at those - I'd only done `BaseProcessor` because tests were actively failing.

alecgibson: @h-lame done.

h-lame:
yup yup. 🚢 🇮🇹

PR #635, opened 2017-07-04
[DO NOT MERGE] Allow choosing mot reminders as a promo for completed transactions

h-lame:
For: https://trello.com/c/gyrP0MaI/174-alternative-message-after-giving-feedback

DVSA want to promote MOT reminders and this adds it as an allowed value for promotion choice.  

We also add tests that make sure we're keeping the list of allowed promotion choices in publisher in sync with the list of allowed promotions in the schemas (see commit notes for some gnarly details).

Relies on https://github.com/alphagov/govuk-content-schemas/pull/632 and a TBD frontend PR to display the new promo.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -59,6 +59,7 @@ end
 
 group :development, :test do
   gem 'govuk-lint', '~> 2.1.0'
+  gem 'govuk_schemas'
   gem 'jasmine', '2.5.2'
   gem 'jasmine-core', '2.5.2'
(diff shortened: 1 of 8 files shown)


h-lame:
Added DNM as we want to confirm with DVSA first that this is what they actually want and the PR on schemas should probably go first either way.

h-lame:
Closing because after discussions with DVSA they decided they didn't want this.

PR #638, opened 2017-08-23
Enable editing of Transaction Start button text

alecgibson:
This change allows publishers to choose the text that appears in the
green “start” button on a Transaction. This behaviour is basically a
direct read-across from the similar functionality in Simple Smart
Answers.

Publishers will be able to select from a drop-down, which only contains
two choices:

- “Start now”
- “Sign in”

“Start now” is the text that is currently displayed, and remains the
default.

“Sign in” is the new copy requested specifically for the Universal
Credit Sign In page.

We use a drop-down to restrict the choices. Copy will need to go through
approval to be included in the drop-down, and having a list of these
items will help to keep consistency across GOV.UK (for example, we will
not have variations like “Sign in”, “Signin”, “Log in”, etc.).

## Before

universal-credit-start-now

## After

universal-credit-sign-in

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -260,6 +260,7 @@ protected
     when :transaction_edition
       [
         :introduction,
+        :start_button_text,
         :will_continue_on,
         :link,
         :more_information,
--- a/app/models/transaction_edition.rb
+++ b/app/models/transaction_edition.rb
@@ -8,10 +8,12 @@ class TransactionEdition < Edition
   field :need_to_know, type: String
   field :department_analytics_profile, type: String
   field :alternate_methods, type: String
+  field :start_button_text, type: String, default: "Start now"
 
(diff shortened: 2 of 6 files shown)


alecgibson:
[(2) Change green button text on transaction page](https://trello.com/c/zo4casKM/722-2-change-green-button-text-on-transaction-page)

fofr:
The UI should be radio buttons when there are only two options. Dropdowns are not easy to use.

alecgibson:
@fofr I used a `select` because that's what we're already using on Simple Smart Answers. If we want to use radio buttons, shouldn't we use them everywhere?

fofr:
Let's start with this.

More guidance on why selects are bad here: https://paper.dropbox.com/doc/Select-boxes-v7PWDvPDV2tmc9pzDbm3s

alecgibson:
I've raised this Trello card to switch both formats over to radio buttons: https://trello.com/c/2JyPdctt/727-change-publisher-start-button-text-select-to-radio-buttons

PR #639, opened 2017-08-25
Substitute select tags with radio buttons

mgrassotti:
Trello: https://trello.com/c/2JyPdctt/727-2-change-publisher-start-button-text-select-to-radio-buttons

According to [our guidance](https://paper.dropbox.com/doc/Select-boxes-v7PWDvPDV2tmc9pzDbm3s), we should avoid `select` elements where possible.

We should then use radio buttons instead of dropdowns on the "Start Button Text" fields for both Transactions and Simple Smart Answers.

### DIFF
--- a/app/views/simple_smart_answers/_fields.html.erb
+++ b/app/views/simple_smart_answers/_fields.html.erb
@@ -10,8 +10,12 @@
 <div class="builder-container">
   <div class="row">
     <div class="col-md-8">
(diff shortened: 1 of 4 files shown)


fofr:
Are there some screenshots of how this looks?

ikennaokpala:
@fofr I have updated the PR with screenshots.

PR #643, opened 2017-09-13
Add organisation content report

DennisTsiang:
For: https://trello.com/c/5oeDn5Xy

This commit brings back the organisation content report that was
removed before. A few of the columns are no longer stored in
the publisher app, but have been moved to the publishing_api.
As such, calls are made to the publishing_api to retrieve the
missing information.

### DIFF
--- a/app/presenters/organisation_content_presenter.rb
+++ b/app/presenters/organisation_content_presenter.rb
@@ -1,0 +1,60 @@
+class OrganisationContentPresenter < CSVPresenter
+  def initialize(scope = Artefact.where(owning_app: "publisher"))
+    super(scope)
+    self.column_headings = [
+        :id,
+        :name,
+        :format,
+        :slug,
+        :state,
+        :browse_pages,
+        :topics,
+        :organisations,
+        :need_ids
+    ]
+  end
+
+private
+
+  def artefact_contains_no_editions(artefact)
+    artefact.latest_edition.nil?
+  end
+
+  def get_value(header, artefact)
+    return super if artefact_contains_no_editions(artefact)
+
+    content_id = artefact.content_id
+
+    case header
+    when :browse_pages
+      expanded_links(content_id, %w(mainstream_browse_pages base_path), /\/browse\//)
+    when :topics
+      expanded_links(content_id, %w(topics base_path), /\/topic\//)
+    when :organisations
+      expanded_links(content_id, %w(organisations title))
+    when :need_ids
+      expanded_links(content_id, %w(meets_user_needs details need_id))
+    when :format
+      artefact.kind
+    else
+      super
+    end
+  end
+
+
+  def expanded_links(content_id, keys, pattern_to_remove = nil)
+    begin
+      response = Services.publishing_api.get_expanded_links(content_id)
+      join_strings(response.to_hash, keys, pattern_to_remove)
+    rescue GdsApi::HTTPNotFound, GdsApi::HTTPClientError
+      return ''
+    end
+  end
+
+  def join_strings(response, keys, pattern_to_remove)
+    results = response.dig('expanded_links', keys[0])
+    values = results.map { |result| result.dig(*keys.drop(1)) }
+    values = values.map { |value| value.gsub(pattern_to_remove, '') } unless pattern_to_remove.nil?
+    values.join(',')
+  end
+end
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -15,6 +15,8 @@ Rails.application.configure do
   # Show full error reports.
   config.consider_all_requests_local = true
 
+  config.jwt_auth_secret = '123'
+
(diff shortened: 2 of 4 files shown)


test/unit/organisation_content_presenter_test.rb
```diff
@@ -0,0 +1,53 @@
...
+                       panopticon_id: document.id)
+
+    OrganisationContentPresenter.new(
+      Artefact.where(owning_app: "publisher").not_in(state: ["archived"])
+    ).to_csv
+  end
+
+  should "retrieve the information needed from the JSON response" do
+    JSON = { 'expanded_links' => { 'organisations' => ['title' => 'HMRC'] } }.freeze
+    str = OrganisationContentPresenter.new.send(:join_strings, JSON, 'organisations', 'title', /.^/)
```
h-lame: It's not usually a good idea to test private methods directly, but to test the outcome of them via the public methods (not least because public methods tend to be more stable than private ones, and it gives you more flexibility with your implementation - a private method is often a candidate for extraction somewhere else during a refactor for example).  

I'd prefer to see a test similar to the first one that pushes an artefact through the presenter and checks the values in the outputted CSV data.  This time though also stubs out the publishing_api calls to return some JSON that has browse pages, orgs, etc... 

There are test helpers in the gds-api-adapters gem for stubbing the API calls and providing canned responses: https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/publishing_api_v2.rb and probably there are example tests in this project that use them which you can use for inspiration.

config/environments/development.rb
```diff
@@ -15,6 +15,8 @@
   # Show full error reports.
   config.consider_all_requests_local = true
 
+  config.jwt_auth_secret = '123'
```
h-lame: It might be worth separating this change out into it's own commit as it's not entirely related to the rest of the changes, and feels like something that could do with it's own commit message to explain why we had to do it.

app/presenters/organisation_content_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def expanded_links(content_id, outer_key, inner_key, regex)
+    response = Services.publishing_api.get_expanded_links(content_id)
+    if response.nil? || response.raw_response_body.empty?
+      return ''
+    end
+    join_strings(response.to_hash, outer_key, inner_key, regex)
+  end
+
+  def join_strings(response, outer_key, inner_key, regex)
```
h-lame: Can we make the regex an optional argument?  It'd be nice if we didn't need to pass in the "everything matches" regex when we are fetching organisations.  It took me a moment to work out what the `/.^/` regex was for when looking at that method call, and we could avoid that effort on the part of the reader, and avoid the computational effort of doing the match if we want the whole string.

Something like the following might work:

```
def join_string(response, outer_key, inner_key, regex = nil)
  results = response.fetch('expanded_links', {}).fetch(outer_key, {})
  values = results.map { |result| results.fetch('inner_key', '') }
  values = values.map { |value| value.gsub(regex, '') } unless regex.nil?
  values.join(',') 
end
```

Might be worth renaming `regex` to something like `pattern_to_remove` or something like that so it's obvious what the regex is used for.  `regex` is really a type name that tells me what the variable is, but not what it's going to be used for and what it's going to be used for is often more helpful when reading the code.

DennisTsiang:
Updated commit with the changes @h-lame requested

app/presenters/organisation_content_presenter.rb
```diff
@@ -0,0 +1,60 @@
...
+      response = Services.publishing_api.get_expanded_links(content_id)
+      join_strings(response.to_hash, keys, pattern_to_remove)
+    rescue GdsApi::HTTPNotFound, GdsApi::HTTPClientError
+      return ''
+    end
+  end
+
+  def join_strings(response, keys, pattern_to_remove)
+    results = response.dig('expanded_links', keys[0])
+    values = results.map { |result| result.dig(*keys.drop(1)) }
```
h-lame: Nice use of dig!

test/unit/organisation_content_presenter_test.rb
```diff
@@ -0,0 +1,86 @@
...
+    ).to_csv
+    data = CSV.parse(csv, headers: true)
+
+    assert_equal "Important document", data[0]["Name"]
+    assert_equal "answer", data[0]["Format"]
+    assert_equal "123456,123321,654321", data[0]["Need ids"]
+    assert_equal "HMRC", data[0]["Organisations"]
+    assert_equal "business-tax/vat", data[0]["Topics"]
+    assert_equal "business/support,tax/vat", data[0]["Browse pages"]
+  end
```
h-lame: This test is much clearer now, and gives us plenty of wiggle room to refactor the implementation while still keeping this test as a harness to make sure we don't break things!

PR #645, opened 2017-09-14
Change dig to fetch when searching through hash

DennisTsiang:
Changes dig to fetch when searching through the hash that is returned from publishing_api in organisation_content_presenter.rb. This allows us to give the value a default empty hash if it can't find it, whereas dig returns nil if the key is not present, and map cannot be called on a nil object.

### DIFF
--- a/app/presenters/organisation_content_presenter.rb
+++ b/app/presenters/organisation_content_presenter.rb
@@ -52,8 +52,8 @@ private
   end
 
   def join_strings(response, keys, pattern_to_remove)
-    results = response.dig('expanded_links', keys[0])
-    values = results.map { |result| result.dig(*keys.drop(1)) }
+    results = response.fetch('expanded_links', {}).fetch(keys[0], {})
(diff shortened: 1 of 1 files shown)


app/presenters/organisation_content_presenter.rb
```diff
@@ -52,7 +52,7 @@ def expanded_links(content_id, keys, pattern_to_remove = nil)
   end
 
   def join_strings(response, keys, pattern_to_remove)
-    results = response.dig('expanded_links', keys[0])
+    results = response.fetch('expanded_links', {}).fetch(keys[0], {})
     values = results.map { |result| result.dig(*keys.drop(1)) }
```
h-lame: I guess we might have the same problem with the `dig` on the next line.  We could use `result.dig(*keys.drop(1)) || ''` to make sure we get a string, or `values = results.map { ... }.compact` to compress the results by throwing away any `nil` entries.

DennisTsiang: Good idea

PR #647, opened 2017-10-09
Change artefact kind when we change edition format

h-lame:
Tangentially for: https://trello.com/c/RRJOTFQo/252-remove-business-support-code-that-still-exist-in-publisher

While exploring how best to remove and properly redirect business finance support editions from publisher I noticed that artefacts did not always have the a `kind` that matched up with the `format` of their `latest_edition`.  Since #428 we've allowed users to create a new edition of a published document with a new type, but we never made sure the artefact kind and edition format synced up.

It's unlikely to cause any real problems, but it felt strange to me so I thought I'd attempt to fix it.  There are notes on each commit, so worth reading them one-by-one.

### DIFF
--- a/db/migrate/20171009143335_synchronise_artefact_kinds_with_latest_edition_formats.rb
+++ b/db/migrate/20171009143335_synchronise_artefact_kinds_with_latest_edition_formats.rb
@@ -1,0 +1,21 @@
+class SynchroniseArtefactKindsWithLatestEditionFormats < Mongoid::Migration
+  def self.up
+    # We don't want to run any update callbacks, just change the column
+    Artefact.skip_callback(:update, :after, :update_editions)
+
+    Artefact.each do |artefact|
+      latest_edition = artefact.latest_edition
+      next if latest_edition.nil?
+      next if latest_edition.kind_for_artefact == artefact.kind
+
+      puts "Changing #{artefact.slug} (#{artefact.content_id}) from #{artefact.kind} to #{latest_edition.kind_for_artefact}"
+      artefact.update_attribute(:kind, latest_edition.kind_for_artefact)
+    end
+
+    # restore the update callback for future migrations
+    Artefact.set_callback(:update, :after, :update_editions)
+  end
+
+  def self.down
+  end
+end
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -328,6 +328,10 @@ class Edition
     format
   end
 
+  def kind_for_artefact
+    format.underscore
+  end
+
   def has_video?
     false
   end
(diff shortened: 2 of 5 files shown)


davidbasalla:
Nice! 👌  I remember coming across this back on Custom team, and I can't remember why we didn't just fix it back then... 😐 

Because you mentioned `CompletedTransaction`s in your commits, I'm now wondering what happens with the slug of an artefact that gets changed from eg an `Answer` to a `CompletedTransaction` and if the '/done' gets appended to it... but that's probably not related to this particular PR.

db/migrate/20171009143335_synchronise_artefact_kinds_with_latest_edition_formats.rb
```diff
@@ -0,0 +1,22 @@
+class SynchroniseArtefactKindsWithLatestEditionFormats < Mongoid::Migration
+  def self.up
+    # We don't want to run any update callbacks, just change the column
+    Artefact.skip_callback(:update, :after, :update_editions)
+
+
```
davidbasalla: extra empty line here

h-lame:
@davidbasalla good point about converting things into `CompletedTransaction`s. I don't really know how important the `done/` prefix is on the slug, so I did some tests (on a local db replicated a couple of days ago on this branch):

```ruby
Artefact.where(kind: 'completed_transaction').count
#=> 150
werent_always_done_pages = Artefact.where(kind: 'completed_transaction').select { |x| Edition.where(panopticon_id: x.id).map(&:kind_for_artefact).uniq != ['completed_transaction'] }
# => [#<Artefact _id: 5072de35a4254a58fa00003c, slug: "driving-transaction-finished",  ... >, #<Artefact _id: 5076dd61e5274a66de000002, slug: "transaction-finished", ... >]
werent_always_done_pages.map &:state
# => ["archived", "archived"]
werent_always_done_pages.map &:redirect_url
# => ["/done/driving-transaction-finished", "/done/transaction-finished"]
werent_always_done_pages.map &:valid?
# => [false, false]
```

Looks like we think the `done/` prefix is important as the two examples we have of artefacts that were turned into done pages were ultimately archived and redirected to new `done/` prefixed versions.  You can't change the slug of a document in publisher (can you?) so this makes sense.  I think in a different PR we probably want to remove done pages from the list of things you can convert things into.

db/migrate/20171009143335_synchronise_artefact_kinds_with_latest_edition_formats.rb
h-lame: oops! rebased. it never happened.

davidbasalla:
@h-lame that sounds sensible to disable the conversion of formats into a CompletedTransaction (in a potential future PR)! 👍 Looks good to merge

PR #648, opened 2017-10-13
Republish business support editions as publishing api redirects or gones

h-lame:
For: https://trello.com/c/RRJOTFQo/252-remove-business-support-code-that-still-exist-in-publisher

This is part 1 of the work for this card to remove business support code and data from publisher and make sure they're all redirected properly in publishing-api instead of router-data.  Part 2 actually deletes the code and edition data (see: #649), part 3 removes the entries from router-data (see: alphagov/router-data#5), and part 4 removes the schema from content-schemas (see: alphagov/govuk-content-schemas#653).

This is a migration that works through all the business_support artefacts in the publisher database and decides what to do with them.  Ultimately we unpublish with a redirect, create a new redirect content item, create a new gone content item, or destroy the artefact.  The choices are somewhat involved so it's best to review this PR commit by commit as we build up the migration in stages.

Open to any thoughts on an alternate approach to this.  For example should we always create new redirect/gone content items with new content_ids and claim they came from short-url-manager?

### DIFF
--- a/app/models/artefact.rb
+++ b/app/models/artefact.rb
@@ -294,7 +294,18 @@ class Artefact
     Downtime.for(self)
   end
 
+  def exact_route?
+    le = latest_edition
+    return le.exact_route? if le.present?
+    return edition_class_name.in? Edition::EXACT_ROUTE_EDITION_CLASSES if owning_app == 'publisher'
+    prefixes.empty?
+  end
+
 private
+
+  def edition_class_name
+    "#{kind.camelcase}Edition"
+  end
 
   def validate_prefixes_and_paths
     if ! self.prefixes.nil? && self.prefixes_changed?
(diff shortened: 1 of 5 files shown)


h-lame:
Oh, this sits on top of #647 so when we merge that I'll rebase this.

h-lame:
Rebased to remove the commits from #647 now it's merged.

h-lame:
FWIW: this has been run successfully on a recent (17th Oct) backup of router, publishing-api, and publisher databases.  In fact it was this successful run that was used to generate the automatic removals in alphagov/router-data#5.

cbaines:
> FWIW: this has been run successfully on a recent (17th Oct) backup of router, publishing-api, and publisher databases. In fact it was this successful run that was used to generate the automatic removals in alphagov/router-data#5.

I've had a quick look through this, and if it works for you, I'm happy with going ahead with this approach.

PR #650, opened 2017-10-18
Set up application for removal of alphagov-deployment

kevindew:
This sets up environment variable conditionals for the various configurations set up in [alphagov-deployment](https://github.com/alphagov/alphagov-deployment/tree/master/publisher) to make way for it's removal.

More details on individual commits.

### DIFF
--- a/config/fact_check.yml
+++ b/config/fact_check.yml
@@ -1,18 +1,13 @@
 # Configuration for the fact check system
-# This file is overwritten on deployment
 
 # The main address pattern sent out as the From address on fact check emails
 # If we ever want to change this format, we'll need to support some notion of
 # legacy formats, so we still pick up emails to old addresses
-address_format: "factcheck+dev-{id}@alphagov.co.uk"
+address_format: <%=ENV.fetch("FACT_CHECK_ADDRESS_FORMAT", "factcheck+dev-{id}@alphagov.co.uk") %>
 
-# In deployed versions, this contains connection details for the fact check
-# account, eg:
-#
-#     address: imap.example.com
-#     port: 993
-#     user_name: factcheck@example.com
-#     password: correct horse battery staple
-#     enable_ssl: true
-#
-fetcher: {}
+fetcher:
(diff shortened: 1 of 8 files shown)


config/fact_check.yml
```diff
@@ -1,18 +1,13 @@
...
-#
-#     address: imap.example.com
-#     port: 993
-#     user_name: factcheck@example.com
-#     password: correct horse battery staple
-#     enable_ssl: true
-#
-fetcher: {}
+fetcher:
+  address: <%=ENV.fetch("FACT_CHECK_ADDRESS", "imap.gmail.com") %>
```
sihugh: I don't think `FACT_CHECK_ADDRESS` or `FACT_CHECK_PORT` are set anywhere.  Is it worth just setting them to "imap.gmail.com" and "993" so that I don't get confused?

config/schedule.rb
```diff
@@ -1 +1,13 @@
-# This file is overwritten on deploy
+set :output, error: 'log/cron.error.log', standard: 'log/cron.log'
+
+schedule_task_prefix = ENV.fetch('SCHEDULE_TASK_PREFIX', '/usr/local/bin/govuk_setenv publisher')
```
sihugh: Is there a reason for looking for this in an env var?  It's not likely to change particularly, is it?

sihugh:
Feel free to deploy to integration when ready 🐙

config/fact_check.yml
kevindew: No they're not set, it just relies on the fallback for them as a default. We can indeed set them up in hieradata to be more concrete about them. We could even just hard code them it just makes it difficult to experiment with different imap servers. 

I'm frequently unsure whether it's better to set these sort of things just as fallbacks to allow easy overriding or to set them up in hieradata, as there are quite a few things you want to normally use default but allow overriding.

config/schedule.rb
kevindew: That's to make it so that it's possible to set this up to run outside of our infrastructure. As it stands your only way to run this is to have `/usr/local/bin/govuk_setenv` available. Which can be quite a pain point, but doesn't actually solve any current problems.

PR #651, opened 2017-10-19
Fix equality typo in republishing business_support migration

h-lame:
We had `=` instead of `==` when detecting the
`jobcentre-plus-vacancy-filling-system-uk` artefact which means we picked
the wrong artefact to check the redirect against.  Unluckily for us this
artefact's routes had the same conditions as the
`jobcentre-plus-vacancy-filling-system-uk` one so picking the wrong
artefact didn't cause an exception to be raised.  Ultimately what this
does is mean we filtered this artefact out of the list of live ones to
process and don't redirect it properly.  Replacing the `=` with `==`
means we pick the correct artefact for
`jobcentre-plus-vacancy-filling-system-uk` and only filter it out.


h-lame:
The upshot of this is that when we ran this on integration it left 2 business_support artefacts as "live".  One we expect, the one for `jobcentre-plus-vacancy-filling-system-uk`, but the other we didn't; it's the one for `urban-development-grant-udg-belfast` which turns out to be the first business_support artefact that is live.  Fixing this typo means we only leave the `jobcentre-plus-vacancy-filling-system-uk` Artefact as "live".

PR #655, opened 2017-11-14
Add a Dockerfile for this app

adrianclay:
Is required for running publisher under publishing-e2e-tests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM ruby:2.3.1
+RUN apt-get update -qq && apt-get upgrade -y
+
+RUN apt-get install -y build-essential nodejs && apt-get clean
+
+ENV GOVUK_APP_NAME publisher
+ENV MONGODB_URI mongodb://mongo/govuk_content_development
+ENV TEST_MONGODB_URI mongodb://mongo/govuk_content_publisher_test
+ENV PORT 3000
+ENV RAILS_ENV development
+ENV REDIS_HOST redis
+
+ENV APP_HOME /app
+RUN mkdir $APP_HOME
+
+WORKDIR $APP_HOME
+ADD Gemfile* $APP_HOME/
+RUN bundle install
+ADD . $APP_HOME
+
+CMD bash -c "rm -f tmp/pids/server.pid && bundle exec rails s -p $PORT -b '0.0.0.0'"
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,14 @@ RUN apt-get update -qq && apt-get upgrade -y
 
 RUN apt-get install -y build-essential nodejs && apt-get clean
 
+ENV PHANTOM_JS phantomjs-2.1.1-linux-x86_64
+
+RUN wget https://bitbucket.org/ariya/phantomjs/downloads/$PHANTOM_JS.tar.bz2 && \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ruby:2.3.1
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential nodejs && apt-get clean
```
thomasleese: I think it's better if the `upgrade` and the `install` steps are separate so the upgraded image can be cached.

Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ruby:2.3.1
+RUN apt-get update -qq && apt-get upgrade -y
+
+RUN apt-get install -y build-essential nodejs && apt-get clean
+
+ENV GOVUK_APP_NAME publisher
+ENV MONGODB_URI mongodb://mongo/govuk_content_development
+ENV TEST_MONGODB_URI mongodb://localhost/govuk_content_publisher_test
+ENV PORT 3006
```
kevindew: Publisher normally runs on port 3000: https://github.com/alphagov/govuk-puppet/blob/31cd78613e34669305f781eed082d2f75e00c0cc/development-vm/Procfile#L1

Hopefully that doesn't cause problems given its common usage

adrianclay: Thanks for the link 👍 .  I'll use 3000 to reduce potential confusion.

adrianclay:
I'll spend an hour getting rspec to run and see how we go.

adrianclay:
I added PhantomJS to the Dockerfile which then allows the integration tests to run.

kevindew:
@adrianclay Thanks! - what command are you using to run the tests I'm getting errors for `docker-compose run publisher bundle exec rake`

adrianclay:
I lost the command I ran, was doing it the bash of a container.

This seems to work though:

```
docker-compose run -e RAILS_ENV=test -e SENTRY_DSN= -e TEST_MONGODB_URI=mongodb://mongo/govuk_content_publisher_test -v $(pwd)/apps/govuk-content-schemas/:/govuk-content-schemas/ -e GOVUK_CONTENT_SCHEMAS_PATH=/govuk-content-schemas/ publisher bundle exec rake test
```

kevindew:
Aces - I was caught on RAILS_ENV being set. Ok looks like we should probably update the `TEST_MONGODB_URI` in this dockerfile and include the `GOVUK_CONTENT_SCHEMA` stuff in docker-compose.yml

Looks like there's a problem running jasmine, seems to be using same port as specified for the app

kevindew:
Argh after a bit of digging I think it's because jasmine is tied to localhost, doesn't seem worth sorting out. Bit annoying but nevermind.

adrianclay:
Strange you say that as I did have a Jasmine issue where it was giving a socket error which suggested a port problem, it has since stopped and is running all the tests fine.

I've updated the docker-compose and Dockerfile and Runs smoothly with ` docker-compose run -e RAILS_ENV=test -e SENTRY_DSN=  publisher bundle exec rake test ` now.

kevindew:
that won't be including the jasmine tests as they are ran through [jasmine:ci](https://github.com/alphagov/publisher/blob/master/Rakefile#L13)

adrianclay:
Doh, of course.

PR #656, opened 2017-11-21
[Do not merge] Update the publisher interface to point to the new guide location

nickcolley:
Depends on: https://github.com/alphagov/govspeak-preview/pull/18

https://trello.com/c/ppqXaGPk/122-create-a-govspeak-component-for-buttons


nickcolley:
Will hold off merging until after the freeze, so if someone has an emergency deployment they don't have to figure out if this change is going to block them 😄

PR #657, opened 2017-11-23
Unpublish pages added for Self Assessment A/B test

leenagupte:
Trello: https://trello.com/c/xdmgIGIO

Add a rake task to unpublish the three pages we added as part of the
Self Assessment A/B test. That test finished on 17/11/2017 and these
particular pages are no longer needed.

- [x] Tested on integration


leenagupte:
Thanks @emmabeynon. Will change it.

PR #660, opened 2017-11-24
Explicitly set the uid of the test user.

adrianclay:
Was defaulting to saving a null value previously.  Also log message
in the event that the user already exists.

### DIFF
--- a/db/seeds/user.rb
+++ b/db/seeds/user.rb
@@ -1,7 +1,13 @@
-gds_organisation_id = "af07d5a5-df63-4ddc-9383-6a666845ebe9"
+if User.where(name: "Test user").present?
+  puts "Skipping because user already exists"
+else
+  gds_organisation_id = "af07d5a5-df63-4ddc-9383-6a666845ebe9"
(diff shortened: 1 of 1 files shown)


db/seeds/user.rb
```diff
@@ -1,7 +1,13 @@
-gds_organisation_id = "af07d5a5-df63-4ddc-9383-6a666845ebe9"
+if User.where(name: "Test user").present?
+  puts "Skipping because user already exists"
+else
+  gds_organisation_id = "af07d5a5-df63-4ddc-9383-6a666845ebe9"
+  user_id = "0d825034-6ea5-49e9-98a0-167ea2a6c487"
```
kevindew: Does it matter which uid this is? i.e. the gds one is the same one that matches the [actual GDS one](https://www.gov.uk/api/organisations/government-digital-service) - if not might be more revealing of intent to use `SecureRandom.uuid`

adrianclay: Sounds good 👍

PR #662, opened 2017-11-27
Upgrade to Ruby 2.4.2 & Bundle update

danielburnley:
Ahead of enabling Dependabot on the repository, we want to be using the latest
dependancies, to minimise the onslaught of incremental PRs.

https://trello.com/c/vX8qwh2X/46-ruby-upgrade-of-all-apps


Gemfile.lock
```diff
@@ -480,4 +491,4 @@ DEPENDENCIES
   whenever
 
 BUNDLED WITH
-   1.14.6
+   1.15.1
```
RoryMacDonald: Would drop this

RoryMacDonald: Actually, scrap that - 1.15.1 is used elsewhere - so lets bump this

PR #664, opened 2017-11-29
Add example service_signin YAML file

leenagupte:
Trello card: https://trello.com/c/bvVMGlvn

## Motivation

We are creating a new "service_sign_in" document type (see PR https://github.com/alphagov/govuk-content-schemas/pull/667) that displays a set of pages to the user to allow them to choose how to sign on or to create an account.

These pages will be rendered in government-frontend. For the MVP, we are not building the Publisher workflow. Instead, we are going to read the content from a YAML file and publish the content item to publishing api via a rake task and a new presenter that we will create in a later PR.

In order to create a service_sign_in content item, we will need content designers to fill out a YAML file with the required content data.  We create an example YAML file to help guide content designers with this task.

### DIFF
--- a/lib/service_sign_in/example.yaml
+++ b/lib/service_sign_in/example.yaml
@@ -1,0 +1,46 @@
+start_page_slug: log-in-file-self-assessment-tax-return
+locale: en
+choose_sign_in:
+  title: Prove your identity to continue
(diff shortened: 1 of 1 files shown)


leenagupte:
@nickcolley We will be 😄  imminently!

PR #666, opened 2017-12-05
Add service_sign in_presenter

emmabeynon:
For [Trello](https://trello.com/c/obBGmd3o/244-create-a-presenter-for-servicesignin-pages-in-publisher-3)
Paired with @leenagupte 

Due to time restrictions in Q3, we are not creating a UI for content designers to create "Service Sign In" pages themselves so we decided to import content item data from a YAML file (see https://github.com/alphagov/publisher/pull/664).  For these reasons we also decided not to create a "ServiceSignIn" model to persist the YAML data to a database and instead decided to follow the Specialist Publisher model of sending the data directly to the Publishing API.

We now create a presenter that reads the content of a given YAML file and constructs a payload that can be sent to the Publishing API and is valid against the `service_sign_in` content schema.

### DIFF
--- a/app/presenters/formats/service_sign_in_presenter.rb
+++ b/app/presenters/formats/service_sign_in_presenter.rb
@@ -1,0 +1,12 @@
+module Formats
+  class ServiceSignInPresenter
+    def initialize
+    end
+
+    def render_for_publishing_api
+      {
+        schema_name: "service_sign_in",
+      }
+    end
+  end
+end
--- a/app/presenters/formats/service_sign_in_presenter.rb
+++ b/app/presenters/formats/service_sign_in_presenter.rb
@@ -87,7 +87,18 @@ module Formats
             content: content[:choose_sign_in][:description]
           }
         ],
+        options: options,
       }
+    end
+
+    def options
+      options = content[:choose_sign_in][:options]
+      options.each do |option|
+        if option.key?(:slug)
+          option[:slug] = "#{base_path}/#{option[:slug]}"
+          option[:url] = option.delete :slug
+        end
+      end
     end
 
     def parent
--- a/app/presenters/formats/service_sign_in_presenter.rb
+++ b/app/presenters/formats/service_sign_in_presenter.rb
@@ -105,6 +105,7 @@ module Formats
     def create_new_account
       {
         title: content[:create_new_account][:title],
+        slug: content[:create_new_account][:slug],
       }
     end
 
--- a/app/presenters/formats/service_sign_in_presenter.rb
+++ b/app/presenters/formats/service_sign_in_presenter.rb
@@ -15,6 +15,7 @@ module Formats
         locale: locale,
         update_type: update_type,
         change_note: change_note,
+        base_path: base_path,
       }
     end
 
@@ -30,6 +31,10 @@ module Formats
 
     def change_note
       content[:change_note]
+    end
+
+    def base_path
+      "/#{content[:start_page_slug]}/sign-in"
     end
   end
 end
--- a/app/presenters/formats/service_sign_in_presenter.rb
+++ b/app/presenters/formats/service_sign_in_presenter.rb
@@ -16,6 +16,7 @@ module Formats
         update_type: update_type,
         change_note: change_note,
         base_path: base_path,
+        routes: routes,
       }
     end
 
@@ -35,6 +36,12 @@ module Formats
 
     def base_path
       "/#{content[:start_page_slug]}/sign-in"
+    end
+
+    def routes
+      [
+        { path: base_path.to_s, type: "prefix" },
+      ]
     end
   end
 end
--- a/app/presenters/formats/service_sign_in_presenter.rb
+++ b/app/presenters/formats/service_sign_in_presenter.rb
@@ -18,6 +18,7 @@ module Formats
         base_path: base_path,
         routes: routes,
         title: title,
+        description: description,
       }
     end
 
@@ -46,7 +47,15 @@ module Formats
     end
 
     def title
-      Edition.where(slug: parent_slug).last.title
+      parent.title
+    end
+
+    def description
+      parent.overview
+    end
+
+    def parent
+      @parent ||= Edition.where(slug: parent_slug).last
     end
 
     def parent_slug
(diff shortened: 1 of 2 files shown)


app/lib/services.rb
```diff
@@ -12,4 +13,8 @@ def self.publishing_api
   def self.calendars
     @calendars ||= GdsApi::Calendars.new(Plek.new.find('calendars'))
   end
+
+  def self.content_store
```
tijmenb: The content store should not be used in publishing apps because it isn't synchronous: there are a number of message queues and caches between the publishing-api and content-store. You can never be sure that you'll get back directly what you just wrote.

If you really need the info in the content-item*, you should use the publishing-api.

app/presenters/formats/service_sign_in_presenter.rb
```diff
@@ -0,0 +1,133 @@
...
+        title: title,
+        description: description,
+        public_updated_at: public_updated_at,
+        details: details,
+      }
+    end
+
+    def content_id
+      begin
+        content_item["content_id"]
```
tijmenb: I would hard code the `content_id` to avoid having to talk to content-store/publishing-api.

app/presenters/formats/service_sign_in_presenter.rb
```diff
@@ -0,0 +1,133 @@
...
+      parent.title
+    end
+
+    def description
+      parent.overview
+    end
+
+    def public_updated_at
+      return DateTime.now.rfc3339 if update_type == "major"
+      content_item["public_updated_at"]
```
tijmenb: Since this will only be run when there's an actual change, you could choose to always use the current time. That would probably prevent you from having to talk to publishing-api/content-store at all.

app/lib/services.rb
emmabeynon: I agree that normally publishing apps shouldn't talk to the content store, but in this case we need to because it's the only way to get the content_id - we're using YAML files to construct and update content items, rather than a database, and it seems the only way to get information about an existing content item from the Publishing API is to look it up using its content_id...which we don't have.

app/presenters/formats/service_sign_in_presenter.rb
leenagupte: These pages could be republished to fix a typo or change the grammar in the text. In that case, you want to publish it as a "minor" update and not want to change the "public_updated_at" date which is for major updates.

app/presenters/formats/service_sign_in_presenter.rb
leenagupte: To be fair, if we were to hard-code the content-id, someone would still have to do a manual lookup of the content-id. We're trying to save them the trouble. Also, we're trying to avoid copy and paste errors as much as possible.

Happy to through our reasoning and approach with you in person.

tijmenb: > someone would still have to do a manual lookup of the content-id

When would you need to do that?

leenagupte: When you republish the content-item. 
Are you suggesting that a random uuid is generated before the first publish, and then added to the YAML config?

app/lib/services.rb
leenagupte: Thanks. We missed `lookup_content_id` when we were looking through the publishing-api docs.

app/presenters/formats/service_sign_in_presenter.rb
tijmenb: I've checked with @kevindew, and he says that when you don't send `public_updated_at`, and the update type is `minor`, the `public_updated_at` in the publishing-api will not changed.

emmabeynon: Ok we should probably include this information in the Publishing API docs, because to me this reads as `public_updated_at` is required unless the document_type is "contact" or "government": https://github.com/alphagov/publishing-api/blob/master/doc/api.md#put-v2contentcontent_id

kevindew: Oh dear! yes we should fix that

app/presenters/formats/service_sign_in_presenter.rb
```diff
@@ -0,0 +1,129 @@
+module Formats
+  class ServiceSignInPresenter
```
sihugh: How are you planning to invoke this when you have a new yaml file?

emmabeynon: We're planning to create a rake task to handle publishing this format, and we'll probably pass the file path as an argument to the rake task, load the file and then send it as an argument to ServiceSignInPresenter.

test/unit/presenters/formats/service_sign_in_presenter_test.rb
```diff
@@ -0,0 +1,209 @@
...
+    @artefact ||= FactoryGirl.create(:artefact, kind: "transaction")
+    @parent ||= FactoryGirl.create(
+      :transaction_edition,
+      panopticon_id: @artefact.id,
+      slug: parent_slug
+    )
+  end
+
+  def parent_slug
+    "log-in-file-self-assessment-tax-return"
```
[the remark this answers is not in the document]
emmabeynon: Probably ⚡️

test/unit/presenters/formats/service_sign_in_presenter_test.rb
```diff
@@ -0,0 +1,209 @@
...
+        publishing_api_has_lookups(base_path => nil)
+        assert_equal "random-uuid-string", subject.content_id
+      end
+    end
+
+    should "use existing content_id if content_item already exists in content-store" do
+      content_id = "random-content-id"
+      publishing_api_has_lookups(base_path => content_id)
+
+      assert_equal content_id, subject.content_id
```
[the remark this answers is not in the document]
emmabeynon: Here we're trying to test that the method returns the content_id for an existing content item instead of generating a new one for us, so it's  correct that we're checking `subject.content_id`.
Also `existing_content_id` is a private method so we won't be able to call that :)

app/presenters/formats/service_sign_in_presenter.rb
sihugh: 👍

PR #669, opened 2017-12-12
Add conditionals to allow running publisher in production on E2E tests

adrianclay:
Hey :wave:,

The publishing-e2e-tests are being built in production mode now, and as part of that initial work we need to disable secure cookie due to not using SSL and sending emails due to no mail server.

The Dockerfile ARG allows us to only compile the assets when we require allowing for easy switching back to development mode and not slowing down any existing usage of this Docker image.

### DIFF
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -2,5 +2,5 @@
 
 Rails.application.config.session_store :cookie_store,
   key: '_publisher_session',
-  secure: Rails.env.production?,
+  secure: Rails.env.production? && !ENV['DISABLE_SECURE_COOKIES'],
   http_only: true
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,4 +20,7 @@ ADD Gemfile* $APP_HOME/
 RUN bundle install
 ADD . $APP_HOME
 
+ARG COMPILE_ASSETS=false
+RUN if [ "$COMPILE_ASSETS" = "true" ] ; then bundle exec rails assets:precompile ; fi
+
 CMD bash -c "rm -f tmp/pids/server.pid && bundle exec rails s -p $PORT -b '0.0.0.0'"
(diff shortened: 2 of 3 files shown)


config/initializers/session_store.rb
```diff
@@ -2,5 +2,5 @@
 
 Rails.application.config.session_store :cookie_store,
   key: '_publisher_session',
-  secure: Rails.env.production?,
+  secure: Rails.env.production? && !ENV['DISABLE_PROD_SSL'],
```
kevindew: Would it be better to have this as `ENV.fetch("USE_SECURE_COOKIES", Rails.env.production?)` ?

adrianclay: We want to be able to specify an ENV var to disable the SSL behaviour.  Given Ruby will cast any string provided in the ENV to a string, we went down this route.

Could do something similar to this:

```ruby
default_secure_cookie = if Rails.env.production? then "yes" else "no" end
use_secure_cookie = ENV.fetch("USE_SECURE_COOKIES", default_secure_cookie) == "yes"
```

kevindew: Ah yes - sorry, my idea was rubbish! 

Your approach seems good - though I guess we might have to add this to loads of repos so I wonder if we can make it more concise. Could be that we went down route of `ENV.fetch("DISABLE_SECURE_COOKIES", !Rails.env.production?)`

adrianclay: Think it would need to be ` secure: !ENV.fetch("DISABLE_SECURE_COOKIES", !Rails.env.production?), ` in full.  Is the double negative too jarring?

kevindew: Yeah I'd agree it's a bit jarring. Probably best to just switch back to what it was originally but name it about cookies.

PR #673, opened 2017-12-14
Generate state machine for docs

elenatanasoiu:
This is an attempt at increasing documentation for the state machine in publisher.

Although you can try to deduce how the state machine works by reading the transitions
in the [workflow](https://github.com/alphagov/publisher/blob/master/app/models/workflow.rb) module, it requires more mental effort than reading a diagram.

I've used the [`state_machines-graphviz` gem](https://github.com/state-machines/state_machines-graphviz) to generate a diagram of the state
machine defined in the [workflow](https://github.com/alphagov/publisher/blob/master/app/models/workflow.rb) module.

The diagram can be (re)generated by doing:

`bundle exec rake state_machines:draw CLASS=Edition`


h-lame:
I think this is a good idea, but I also suggested to @elenatanasoiu how it might be done, so I figure that someone else should be the "2i" on this.

Gemfile
```diff
@@ -37,6 +37,7 @@ gem 'mongoid', '6.1.0'
 gem 'state_machines', '~> 0.4'
 gem 'state_machines-mongoid', '~> 0.1'
 gem 'statsd-ruby', '~> 1.4.0', require: false
+gem 'state_machines-graphviz'
```
h-lame: maybe this should be in a development group only, we don't really want/need it on production.

elenatanasoiu: moved

elenatanasoiu:
Thanks :+1: I've added a reference in the README

PR #674, opened 2017-12-15
Add Service Sign In YAML validator

emmabeynon:
For [Trello](https://trello.com/c/UHPwDcmd/273-validate-the-contents-of-the-yaml-file)

In https://github.com/alphagov/publisher/pull/668 we created a rake task that takes the contents of a YAML file and prepares a payload in order to publish "Service Sign In" content to the Publishing API.  In this PR we create a validator to ensure that the required content is present in the YAML file in order to perform this task successfully.  After merging this PR we will update the rake task to use this validator.

### DIFF
--- a/lib/service_sign_in_yaml_validator.rb
+++ b/lib/service_sign_in_yaml_validator.rb
@@ -1,15 +1,27 @@
 class ServiceSignInYamlValidator
+  attr_reader :yaml_file
+
   def initialize(file_name)
     @file_name = file_name
+    @errors = []
   end
 
   def validate
-    load_yaml_file
+    valid? ? @yaml_file : @errors
   end
 
 private
 
+  def valid?
+    load_yaml_file
+  end
+
   def load_yaml_file
-    YAML.load_file(@file_name)
+    begin
+      @yaml_file = YAML.load_file(@file_name)
+    rescue SystemCallError
+      @errors << "Invalid file path: #{@file_name}"
+      return false
+    end
   end
 end
--- a/lib/service_sign_in_yaml_validator.rb
+++ b/lib/service_sign_in_yaml_validator.rb
@@ -1,6 +1,9 @@
 class ServiceSignInYamlValidator
   attr_reader :yaml_file
 
+  REQUIRED_TOP_LEVEL_FIELDS =
+    %w(change_note choose_sign_in locale start_page_slug update_type).freeze
+
   def initialize(file_name)
     @file_name = file_name
     @errors = []
@@ -14,6 +17,8 @@ private
 
   def valid?
     load_yaml_file
+    return unless @yaml_file.present?
+    check_for_top_level_required_fields
     @errors.empty?
   end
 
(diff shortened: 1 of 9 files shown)


test/unit/lib/service_sign_in_yaml_validator_test.rb
```diff
@@ -0,0 +1,24 @@
...
+class ServiceSignInYamlValidatorTest < ActiveSupport::TestCase
+  def service_sign_in_yaml_validator(file)
+    ServiceSignInYamlValidator.new(file)
+  end
+
+  def valid_yaml_file
+    "lib/service_sign_in/example.yaml"
+  end
+
+  def validated_hash
```
leenagupte: I think I'd rename this. It is returning a valid hash, but not one validated by our validator.

emmabeynon: Any suggestions?

leenagupte: `content`? Naming things is hard!

emmabeynon: Now updated 🌮

lib/service_sign_in_yaml_validator.rb
```diff
@@ -19,6 +20,7 @@ def valid?
     load_yaml_file
     return unless @yaml_file.present?
     check_for_top_level_required_fields
+    check_for_choose_sign_in_required_fields if choose_sign_in.present?
```
leenagupte: Shouldn't this error if choose_sign_in isn't present first as choose_in_sign itself is required?

emmabeynon: If `choose_sign_in` isn't present it will be picked up in `check_for_top_level_required_fields` and already logged as an error.

lib/service_sign_in_yaml_validator.rb
```diff
@@ -48,13 +48,28 @@ def check_for_choose_sign_in_required_fields
       return
     end
 
+    check_choose_sign_in_top_level_fields(error_message)
+    return unless choose_sign_in["options"].present?
```
leenagupte: Shouldn't this line return an error as there must be at least one (two?) options?

emmabeynon: `check_choose_sign_in_top_level_fields` will check if `options` is present and log an error if it isn't.  However, we should probably check if `options` has an array of options, similar to what we do in line 46.

lib/service_sign_in_yaml_validator.rb
leenagupte: Sorry. I missed that!

lib/service_sign_in_yaml_validator.rb
leenagupte: We don't need the guard then, as "options" must be present at this point. But you're right, we should check that options has an array.

emmabeynon: If `options` isn't present that we don't want to call `check_choose_sign_in_options_fields` as it'll blow up, so we return in this scenario.
Going to add in a check that if `options` is present then it's an array 👍

lib/service_sign_in_yaml_validator.rb
emmabeynon: If `choose_sign_in` isn't present that we don't want to call `check_for_choose_sign_in_required_fields` as it'll blow up, so we return in this scenario.

PR #675, opened 2017-12-15
Pass "locale" to service_sign_in publish call

leenagupte:
Related to PR #668 

Publishing-api assumes that the locale is "en" if you do not
explicitly pass the locale in the publish call. As we are also
publishing Welsh content, we should always pass the locale of the
content item.

We have to pass the update_type too now because of the order that
publishing-api expects to see the params.

### DIFF
--- a/app/services/service_sign_in_publish_service.rb
+++ b/app/services/service_sign_in_publish_service.rb
@@ -1,8 +1,9 @@
 class ServiceSignInPublishService
   def self.call(content)
     content_id = content.content_id
+    update_type = nil
     Services.publishing_api.put_content(content_id, content.render_for_publishing_api)
     Services.publishing_api.patch_links(content_id, links: content.links)
-    Services.publishing_api.publish(content_id)
+    Services.publishing_api.publish(content_id, update_type, locale: content.locale)
   end
 end
--- a/test/unit/services/service_sign_in_publish_service_test.rb
+++ b/test/unit/services/service_sign_in_publish_service_test.rb
@@ -8,9 +8,10 @@ class ServiceSignInPublishServiceTest < ActiveSupport::TestCase
   end
 
   should "publish edition to PublishingAPI" do
+    update_type = nil
     Services.publishing_api.expects(:put_content).with(content_id, payload)
     Services.publishing_api.expects(:patch_links).with(content_id, links: links)
-    Services.publishing_api.expects(:publish).with(content_id)
+    Services.publishing_api.expects(:publish).with(content_id, update_type, locale: locale)
 
     ServiceSignInPublishService.call(presenter)
   end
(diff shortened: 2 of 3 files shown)


app/services/service_sign_in_publish_service.rb
```diff
@@ -3,6 +3,6 @@ def self.call(content)
     content_id = content.content_id
     Services.publishing_api.put_content(content_id, content.render_for_publishing_api)
     Services.publishing_api.patch_links(content_id, links: content.links)
-    Services.publishing_api.publish(content_id)
+    Services.publishing_api.publish(content_id, content.update_type, locale: content.locale)
```
emmabeynon: We don't need to send `update_type` in publish - it has been deprecated in favour of sending it in the put_content payload (see https://github.com/alphagov/publishing-api/blob/master/doc/api.md#post-v2contentcontent_idpublish)

test/unit/services/service_sign_in_publish_service_test.rb
```diff
@@ -10,7 +10,7 @@ class ServiceSignInPublishServiceTest < ActiveSupport::TestCase
   should "publish edition to PublishingAPI" do
     Services.publishing_api.expects(:put_content).with(content_id, payload)
     Services.publishing_api.expects(:patch_links).with(content_id, links: links)
-    Services.publishing_api.expects(:publish).with(content_id)
+    Services.publishing_api.expects(:publish).with(content_id, update_type, locale: locale)
```
emmabeynon: As above, sending `update_type` with publish has been deprecated.

test/unit/services/service_sign_in_publish_service_test.rb
```diff
@@ -20,6 +20,8 @@ def presenter
       render_for_publishing_api: payload,
       content_id: content_id,
       links: links,
+      update_type: update_type,
```
emmabeynon: As above, sending update_type with publish has been deprecated.

test/unit/services/service_sign_in_publish_service_test.rb
```diff
@@ -31,6 +33,14 @@ def links
     { parent: ["6a2bf66e-2313-4204-afd5-9940de5e1d66"] }
   end
 
+  def update_type
```
emmabeynon: As above, sending update_type with publish has been deprecated.

app/services/service_sign_in_publish_service.rb
emmabeynon: Sure, but it's an optional argument which is only there because lots of apps are still sending it with publish when they don't need to (see https://github.com/alphagov/gds-api-adapters/pull/719).  If we included it now we'd only have to remove it in future, so I think we shouldn't include this in our code now.

PR #689, opened 2017-12-18
Update example workflow service sign in

nickcolley:
Update the example YAML file so that is easier to duplicate and reuse.


sihugh:
@leenagupte is this still relevant?

leenagupte:
@sihugh Yeah. It makes the example more correct.

PR #694, opened 2017-12-18
Fix Service Sign In Presenter bug

emmabeynon:
The `service_sign_in` format considers `choose_sign_in` `description` to be an
optional field.  We found a bug in the Service Sign In Presenter where
`choose_sign_in` `description` is always included in the payload regardless of
whether it is provided in a service_sign_in YAML file, causing the
`service_sign_in` schema validation to fail when `description` is not provided.
This makes the inclusion of `choose_sign_in` `description` in the payload
dependent on its presence in the YAML file.

### DIFF
--- a/test/unit/presenters/formats/service_sign_in_presenter_test.rb
+++ b/test/unit/presenters/formats/service_sign_in_presenter_test.rb
@@ -139,14 +139,21 @@ class ServiceSignInTest < ActiveSupport::TestCase
           result[:details][:choose_sign_in][:slug]
       end
 
-      should "[:description]" do
-        expected = [
-          {
-            content_type: "text/govspeak",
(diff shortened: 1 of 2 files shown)


test/unit/presenters/formats/service_sign_in_presenter_test.rb
```diff
@@ -139,14 +139,21 @@ def result
...
 
-      should "[:description]" do
-        expected = [
-          {
-            content_type: "text/govspeak",
-            content: @content[:choose_sign_in][:description]
-          }
-        ]
-        assert_equal expected, result[:details][:choose_sign_in][:description]
+      context "[:description]" do
```
nickcolley: nice use of context 👍

PR #713, opened 2017-12-21
Add service sign in unpublish tasks

emmabeynon:
Trello: https://trello.com/c/HewVN7nP/237-create-a-rake-task-to-unpublish-service-sign-in-pages

As `service_sign_in` content cannot be published through the UI, we create rake tasks to unpublish it with the `gone` and `redirect` unpublishing types.  We've decided not to unpublish `service_sign_in` as a withdrawal as having the withdrawal notice above the radio buttons could be confusing to users.  We expect that `redirect` and `gone` will be most commonly used.

We've also found that when redirecting `service_sign_in` content, the redirect does work but the last slug is retained in the path e.g. `/check-income-tax-current-year/sign-in/prove-identity` => `/national-minimum-wage-rates/prove-identity`.  This is consistent with what occurs when redirecting guides so it seems reasonable to maintain this behaviour for the purposes of this particular piece of work.

### DIFF
--- a/lib/tasks/service_sign_in.rake
+++ b/lib/tasks/service_sign_in.rake
@@ -44,5 +44,18 @@ namespace :service_sign_in do
     end
   end
 
+  desc "Unpublish service_sign_in content with type 'gone'"
+  task :unpublish_without_redirect, %i(content_id locale) => :environment do |_, args|
+    USAGE_MESSAGE =
+      "> usage: rake service_sign_in:unpublish_without_redirect[content-id-example,cy]\n".freeze
+
+    content_id = args[:content_id]
+    locale = args[:locale]
+    abort USAGE_MESSAGE unless content_id && locale
+
+    ServiceSignInUnpublishService.call(content_id, locale)
+    puts "> #{content_id} has been unpublished"
+  end
+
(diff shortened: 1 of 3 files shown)


lib/tasks/service_sign_in.rake
```diff
@@ -46,4 +43,39 @@ namespace :service_sign_in do
...
+    abort USAGE_MESSAGE unless content_id && locale
+
+    ServiceSignInUnpublishService.call(content_id, locale)
+    puts "> #{content_id} has been unpublished"
+  end
+
+  desc "Unpublish service_sign_in content with type 'redirect'"
+  task :unpublish_with_redirect, %i(content_id locale redirect_path) => :environment do |_, args|
+    USAGE_MESSAGE =
+      "> usage: rake service_sign_in:unpublish_with_redirect[content-id-example,cy,/redirect/path]\n".freeze
```
nickcolley: Is it convention to have no spaces between arguments in a rake task?

emmabeynon: Yes the task tends to break if there are spaces between arguments.

nickcolley: Good to know thanks 👍

PR #726, opened 2018-01-04
Updated option and hint text

thomasjhughes:
I've changed the text on the options from "Use XXXX" to "Sign in with XXXX". I've also removed the introductory sentence ("If you’re filing online for the first time...") and added a similar sentence as hint text under the 3rd option. 

These changes were made in response to some user feedback we'd received over the last couple of weeks.

### DIFF
--- a/lib/service_sign_in/file_self_assessment.en.yaml
+++ b/lib/service_sign_in/file_self_assessment.en.yaml
@@ -3,7 +3,6 @@ locale: en
 choose_sign_in:
   title: Prove your identity to continue
   slug: prove-identity
-  description:
   options:
     - text: Sign in with Government Gateway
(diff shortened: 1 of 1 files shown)


lib/service_sign_in/file_self_assessment.en.yaml
```diff
@@ -3,16 +3,17 @@ locale: en
 choose_sign_in:
   title: Prove your identity to continue
   slug: prove-identity
-  description: If you’re filing online for the first time, you need to register for Self Assessment first.
+  description:
```
emmabeynon: `description` isn't a required field so we can remove this line from the file (otherwise it breaks if we try to import a blank value 😊 )

PR #727, opened 2018-01-04
Updated option and hint text

thomasjhughes:
I've updated the Welsh version of this service sign-in page to reflect the changes I've made to the English version (see my other pull request - #726)


lib/service_sign_in/file_self_assessment.cy.yaml
```diff
@@ -3,16 +3,17 @@ locale: cy
 choose_sign_in:
   title: Profwch pwy ydych chi i fwrw ymlaen
   slug: profi-pwy-ydw-i
-  description: Os ydych chi’n ffeilio ar-lein am y tro cyntaf, bydd angen i chi gofrestru ar gyfer Hunanasesiad yn gyntaf.
+  description:
```
emmabeynon: Same here, we can just remove this line if it's blank.

PR #737, opened 2018-01-05
Add Link Report Creator Service

SebAshton:
This PR adds two service classes, one [`LinkCheckReportCreator`] to call the `link-checker-api` to create reports, and another [`EditionLinkExtractor`] to extract links from various subclasses of `Edition`

### DIFF
--- a/app/services/edition_link_extractor.rb
+++ b/app/services/edition_link_extractor.rb
@@ -1,0 +1,43 @@
+require "govspeak/link_extractor"
+
+class EditionLinkExtractor
+  def initialize(edition:)
+    @edition = edition
+  end
+
+  def call
+    if has_parts?
+      links_in_govspeak_fields + links_in_parts
+    else
+      links_in_govspeak_fields
+    end
+  end
+
+private
+
+  attr_reader :edition
+
+  def has_parts?
+    edition.parts.any?
+  rescue NoMethodError
+    false
+  end
+
+  def links_in_govspeak_fields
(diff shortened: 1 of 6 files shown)


app/services/edition_link_extractor.rb
```diff
@@ -0,0 +1,43 @@
...
+      links_in_govspeak_fields
+    end
+  end
+
+private
+
+  attr_reader :edition
+
+  def has_parts?
+    edition.parts.any?
```
SebAshton: Not too sure/sold on this but I couldn't get `edition.has_attribute?()` to return correctly

SebAshton: Which I assume is because they are embedded?

app/services/edition_link_extractor.rb
```diff
@@ -0,0 +1,43 @@
...
+  rescue NoMethodError
```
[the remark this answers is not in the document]
SebAshton: So editions, e.g. [`PlaceEdition`](https://github.com/alphagov/publisher/blob/master/app/models/place_edition.rb) don't have `parts` embedded in them, so throws a NoMethodError

app/services/edition_link_extractor.rb
thomasleese: Is it because an edition doesn't have a `parts` or that `parts` is `nil` so `any?` doesn't exist?

app/services/edition_link_extractor.rb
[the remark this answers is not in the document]
thomasleese: Ah yes, that makes sense 👍

PR #740, opened 2018-01-08
Add LinkCheckReport#create

SebAshton:
This PR adds the ability to trigger link checks on all GovSpeak fields within an edition.

screen shot 2018-01-08 at 14 38 41

### DIFF
--- a/app/controllers/link_check_reports_controller.rb
+++ b/app/controllers/link_check_reports_controller.rb
@@ -1,0 +1,13 @@
+class LinkCheckReportsController < ApplicationController
+  before_filter :find_reportable
+
+  def create
+
+  end
+
+private
+
+  def find_reportable
+    @reportable = Edition.find(params[:edition_id])
+  end
+end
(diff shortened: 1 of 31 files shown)


app/models/edition.rb
```diff
@@ -441,6 +441,12 @@ def content_id
     artefact.content_id
   end
 
+  def latest_link_check_report
+    return nil if link_check_reports.empty?
```
thomasleese: I don't think we need this as `.last` returns `nil` if the array is empty.

app/controllers/link_check_reports_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    respond_to do |format|
+      format.js { render 'link_check_reports/create' }
+      format.html { redirect_to edition_url(@reportable.id) }
+    end
+  end
+
+private
+
+  def find_reportable
+    @reportable = Edition.find(params[:edition_id])
```
thomasleese: I would be tempted to call this `@edition` too.

PR #744, opened 2018-01-10
Add controller actions and UI for  triggering and visualising LinkCheckReports

SebAshton:
This PR adds `LinkCheckReportsController#create` and `LinkCheckReportsController#show` endpoints to publisher as well as the front end UI that calls these new endpoints.

screen shot 2018-01-08 at 14 38 41

Close https://github.com/alphagov/publisher/pull/740

### DIFF
--- a/app/controllers/link_check_reports_controller.rb
+++ b/app/controllers/link_check_reports_controller.rb
@@ -1,0 +1,13 @@
+class LinkCheckReportsController < ApplicationController
+  before_filter :find_reportable
+
+  def create
+
+  end
+
+private
+
+  def find_reportable
+    @reportable = Edition.find(params[:edition_id])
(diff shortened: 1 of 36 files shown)


app/controllers/link_check_reports_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  def convert_to_bson_object_id(id)
+    BSON::ObjectId.from_string(id)
+  end
+
+  def permitted_params
+    params.permit(:edition_id, :id)
+  end
+
+  def find_reportable
+    @reportable = Edition.find(permitted_params[:edition_id])
```
thomasleese: Since we're only working in editions here, I would call this `@edition`.

PR #755, opened 2018-01-18
Make the link check box work in smaller browser windows

boffbowsh:
For browser windows below 992px, the link check box would cause the
form fields to get squashed into the left hand side of the well and
become unusable. For breakpoints below 992px (`md` in Bootstrap terms),
disable the float and force the box to appear above the form.

This is possibly not the best design decision, but it’s causing issues
at present for one of our partially sighted users.

Before:
screen shot 2018-01-18 at 14 39 35

After:
screen shot 2018-01-18 at 14 39 43

And above 992px:
screen shot 2018-01-18 at 14 40 53

### DIFF
--- a/app/assets/stylesheets/broken_links_report.scss
+++ b/app/assets/stylesheets/broken_links_report.scss
@@ -1,6 +1,11 @@
 .link-check-report {
-  float: right;
   margin-top: 25px;
+}
+
+@media (min-width: $screen-md-min) {
+  .link-check-report {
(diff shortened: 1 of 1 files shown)


boffbowsh:
Probably more one for @fofr's team to review. Does this make sense to your team, Paul? Is this something you'd track and look at, or someone else? I'm going to merge it when it passes anyway, as it's definitely better than the current situation.

PR #861, opened 2018-05-21
HUB-143: Add description to service sign in pages

rachelthecodesmith:
Add a description paragraph that will be displayed under the title on each of the 4 services using the service sign in pattern, each description includes the service specific text. This "description" is already present in the example yaml, the schema and on the page template on government-frontend but was hitherto unused. Should this prove to be a good pattern according to the base line test we will add a description to the welsh pages as well.

https://govukverify.atlassian.net/browse/HUB-143

hub_143

solo: @rachelthecodesmith


Davidslv:
Can you add a link to the card please?

PR #862, opened 2018-05-24
Process mail based on env var

steventux:
https://trello.com/c/MjhoqAug/168-investigate-how-mainstream-publisher-should-be-checking-for-emails-on-integration-2

The environment var `RUN_FACT_CHECK_FETCHER` is set per environment to control processing of fact check emails.

The [previous implementation used this var to determine whether or not to set the
imap details](https://github.com/alphagov/publisher/blob/master/lib/mail_fetcher_config.rb#L9) but not to prevent attempts to process the emails on the server.

So `Mail#all` would attempt to retrieve mail based on defaults and this would result in a [connection refused error on localhost:110](https://sentry.io/govuk/app-publisher/issues/523285917/?query=environment:integration-blue-aws).

The PR alters the behaviour to a no-op if the env var is not set.

### DIFF
--- a/lib/fact_check_email_handler.rb
+++ b/lib/fact_check_email_handler.rb
@@ -34,12 +34,14 @@ class FactCheckEmailHandler
 
   # &after_each_message: an optional block to call after processing each message
   def process(&after_each_message)
-    Mail.all(read_only: false, delete_after_find: true) do |message|
-      message.skip_deletion unless process_message(message)
-      begin
-        after_each_message.call(message) if after_each_message
-      rescue => e
-        GovukError.notify(e)
+    if ENV.include?("RUN_FACT_CHECK_FETCHER")
+      Mail.all(read_only: false, delete_after_find: true) do |message|
+        message.skip_deletion unless process_message(message)
+        begin
+          yield(message) if block_given?
+        rescue StandardError => e
+          GovukError.notify(e)
+        end
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/fact_check_email_handler.rb
```diff
@@ -34,12 +34,14 @@ def process_message(message)
 
   # &after_each_message: an optional block to call after processing each message
   def process(&after_each_message)
-    Mail.all(read_only: false, delete_after_find: true) do |message|
-      message.skip_deletion unless process_message(message)
-      begin
-        after_each_message.call(message) if after_each_message
-      rescue => e
-        GovukError.notify(e)
+    if ENV["RUN_FACT_CHECK_FETCHER"]
```
thomasleese: I wonder if should we do `ENV.include?` here just in case someone thinks they can set the variable to `false` instead of just not setting it? I realise currently it's the same behaviour as before so not a big deal.

test/integration/fact_check_email_test.rb
```diff
@@ -1,6 +1,15 @@
 require 'integration_test_helper'
 
 class FactCheckEmailTest < ActionDispatch::IntegrationTest
+  setup do
```
thomasleese: Some other apps use [Climate Control](https://github.com/thoughtbot/climate_control) to change environment variable in tests which is quite nice, might be worth looking into here, but again realise it's the same behaviour as before, so I don't mind.

lib/fact_check_email_handler.rb
steventux: I think it's worth using `ENV.include?`, I agree it's a bit of a duff test otherwise.

PR #866, opened 2018-05-29
Bump to ruby-2.5.1

barrucadu:
[Trello card](https://trello.com/c/eE1scfjM/190-update-ruby-apps-to-use-ruby-25)


cbaines:
@barrucadu I think we may as well just temporarily disable `simplecov` until the bug is fixed, and I've put this change in https://github.com/alphagov/publisher/pull/885

PR #873, opened 2018-06-04
Update Government Gateway links for check-income-tax

thomasleese:
[Zendesk Ticket](https://govuk.zendesk.com/agent/tickets/2803860)


thomasleese:
I'm unsure whether the `_ga` parameters were put in by us to track these or HMRC and whether we need to keep them.

PR #878, opened 2018-06-13
Create experimental "knowledge API"

tijmenb:
Running this rake task creates a new content item at https://www.gov.uk/api/content/knowledge-alpha. It is currently populated with the content of a number of high-profile guides in a structured format.

- Minimum wage rates from https://www.gov.uk/national-minimum-wage-rates
- Income Tax rates and bands from https://www.gov.uk/income-tax-rates
- Self Assessment tax return deadlines from https://www.gov.uk/self-assessment-tax-returns/deadlines

This is part of an experiment. We're currently working with one of the voice providers (starts with an A) to see if we can get GOV.UK facts directly into the knowledge engine that underpins the service. We've checked this data format with them, and they said that they would be able to consume this.

Obviously this isn't a very scalable solution - the information from the guides is currently duplicated. Once the data changes we'll have to update these numbers as well (a process for this will be set up).

A better idea might be to either 1) extract the data from the content, 2) manage the data elsewhere and inject it into the content. This would require a lot of initial development however, so this initial "API" serves as a test case. If we find that the voice provider won't be able to use it, we can remove this content item easily.

This commit requires https://github.com/alphagov/govuk-content-schemas/pull/772 to be deployed.

https://trello.com/c/4mRr4Mzi

### DIFF
--- a/app/services/knowledge_api.rb
+++ b/app/services/knowledge_api.rb
@@ -1,0 +1,27 @@
+class KnowledgeApi
+  CONTENT_ID = "4dbbc2d1-a0db-441b-96c9-a743d067f724".freeze
+  KNOWLEDGE_ENTRIES = YAML.load_file(Rails.root.join("app", "services", "knowledge_api.yml")).freeze
+
+  def publish
+    Services.publishing_api.put_content(CONTENT_ID, payload)
+    Services.publishing_api.publish(CONTENT_ID, "major")
+  end
+
+  def payload
+    {
+      base_path: "/knowledge-alpha",
+      title: "Experimental knowledge API",
+      document_type: "knowledge_alpha",
+      phase: "alpha",
+      publishing_app: "publisher",
+      rendering_app: "frontend",
+      schema_name: "knowledge_alpha",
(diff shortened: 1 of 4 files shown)


app/services/knowledge_api.yml
```diff
@@ -0,0 +1,48 @@
+- id: minimum-wage
+  title: Minimum wage rates
+  source: https://www.gov.uk/national-minimum-wage-rates
+  data:
+    - age_range: 25 and over
+      minumum_wage: 7.83
```
rubenarakelyan: `minimum` :)

app/services/knowledge_api.rb
```diff
@@ -0,0 +1,27 @@
+class KnowledgeApi
+  CONTENT_ID = "4dbbc2d1-a0db-441b-96c9-a743d067f724"
+  ENTRIES = YAML.load_file("#{Rails.root}/app/services/knowledge_api.yml")
```
rubenarakelyan: Can do `.freeze` on these two

PR #885, opened 2018-06-26
Fix linting errors + change to Ruby 2.5.1

cbaines:
This took far too long, and I haven't even got started on what I was trying to do :(

For some reason, the Rubocop was tripping over around here [1], failing to actually parse the file. Updating to Ruby 2.5.1 (and/or running bundle update) fixes this.

1: https://github.com/alphagov/publisher/blob/master/lib/licence_content_importer.rb#L102-L105

Get to the point where all of the code passes, so that the Jenkinsfile can be altered to always check the entire codebase, hopefully ensuring we can't get in to such a state again.


cbaines:
> +1 I can't believe we don't use gravatars any more :-(

Haha, I think that may have been before my time. Thanks for taking a look.

PR #893, opened 2018-07-12
Use govuk_app_config health checks

thomasleese:
This makes the health check code simpler to understand and adds new health checks on the database and Sidekiq connection.

[Trello Card](https://trello.com/c/iuPQ6P88/318-add-common-parameterised-healthcheck-classes)


test/functional/healthcheck_controller_test.rb
```diff
@@ -1,123 +0,0 @@
...
-    end
-
-    should "report a critical error" do
-      get :check
-
-      assert_equal "critical", json["checks"]["schedule_queue"]["status"]
-    end
-  end
-
-  context "Mongo connection fails" do
```
steventux: Is there a reason why we're not continuing to cover the "No Mongo|Redis connection" scenarios?

thomasleese: I figured that these are better handled by separate health checks rather than part of this check. If an exception occurs during a health check, I think it's better for the `GovukHealthcheck` framework to handle it and show the check as critical.

PR #916, opened 2018-08-06
Fix location of assets

rubenarakelyan:
This commit changes the sortable table stylesheet to correctly use the Rails asset pipeline for images.


rubenarakelyan:
Smokey in integration is failing because it can't find these assets. Turns out that on AWS for an unknown reason the original assets are not deployed, but only the Rails asset pipeline versioned ones are.

PR #917, opened 2018-08-06
Add hide chapter navigation option to guides

andysellick:
Supercedes #914.

On some guides that occur within a step by step navigation, the guide navigation elements are conflicting with the step navigation, or needlessly duplicating them. This adds an option to publisher to allow content managers the ability to hide the guide navigation elements if that part is inside a step by step navigation.

Example of a guide part currently:

screen shot 2018-07-31 at 14 10 07

The same page but with this option turned on:

screen shot 2018-07-31 at 14 10 22

What this looks like in publisher:

screen shot 2018-08-07 at 14 17 58

**This PR is a work in progress**. Also I've not made a change to publisher before, so any suggestions welcome.

Related change to content schemas: https://github.com/alphagov/govuk-content-schemas/pull/802

Trello card: https://trello.com/c/RNsA7E4Q/763-hide-chapter-navigation-on-mainstream-guides

### DIFF
--- a/test/integration/guide_create_edit_test.rb
+++ b/test/integration/guide_create_edit_test.rb
@@ -27,6 +27,37 @@ class GuideCreateEditTest < JavascriptIntegrationTest
     assert_equal "Foo bar", g.title
   end
 
+  should "editing a GuideEdition, and hiding chapter navigation" do
+    guide = FactoryBot.create(:guide_edition,
+                               panopticon_id: @artefact.id,
+                               state: 'draft',
+                               title: "Foo bar #0")
+
+    visit_edition guide
+    refute find('#edition_hide_chapter_navigation').checked?
+
+    # using 'check' and similar capybara approaches does not work here for some reason
+    page.execute_script("$('#edition_hide_chapter_navigation').attr('checked', 'checked')")
+
+    assert find('#edition_hide_chapter_navigation').checked?
+    save_edition_and_assert_success
+
+    visit_edition guide
+    assert find('#edition_hide_chapter_navigation').checked?
+  end
+
+  should "show hide_chapter_navigation as selected" do
+    guide = FactoryBot.build(:guide_edition,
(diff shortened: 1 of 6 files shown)


andysellick:
@leenagupte I think this is all working apart from one test, which I can't get to work (may be a few extra failures until I've merged the content schemas change as well).

app/models/edition.rb
```diff
@@ -445,6 +445,10 @@ def latest_link_check_report
     link_check_reports.last
   end
 
+  def hide_chapter_navigation
```
leenagupte: I don't think you need this method here. I think adding the field to `GuideEdition` is enough.

test/integration/guide_create_edit_test.rb
```diff
@@ -27,6 +27,34 @@ class GuideCreateEditTest < JavascriptIntegrationTest
...
 
+  should "editing a GuideEdition, and hiding chapter navigation" do
+    guide = FactoryBot.create(:guide_edition,
+                               panopticon_id: @artefact.id,
+                               state: 'draft',
+                               title: "Foo bar #0")
+
+    visit_edition guide
+    refute find('#edition_hide_chapter_navigation').checked?
+    page.execute_script("$('#edition_hide_chapter_navigation').attr('checked', 'checked')")
```
andysellick: @leenagupte got it working 👍

leenagupte: It's a bit ugly (unreadable). Can you add a card to our backlog to revisit this test and get it working "properly". That could mean switching to using the headless Chrome browser.

andysellick: Added - https://trello.com/c/7gXatxse/618-improve-hidechapternavigation-test, also added a comment to the file to clarify.

leenagupte: Thanks @andysellick!

PR #937, opened 2018-09-11
EID-942 - Create GOV.UK sign-in interstitial page for Check your State Pension

JHjava:
- Create service sign in page for Check your State Pension, to make
the introduction to eIDAS option easier
- Include option on the sign in page so user can select to confirm
their identiy with another european country. For now until eIDAS is enabled
on verify, this will take them to the usual verify sign in page.
- Radio button grouping for the sign in page is situated within the choose_sign_in_presenter.rb
within the government-frontend repo. Where the last option in the order will be after the or.
- For now, this will not be linked to the current check-state-pension start page.

### DIFF
--- a/lib/service_sign_in/check-state-pension.en.yaml
+++ b/lib/service_sign_in/check-state-pension.en.yaml
@@ -1,0 +1,53 @@
+start_page_slug: check-state-pension
+locale: en
+choose_sign_in:
+  title: Prove your identity to continue
+  slug: prove-identity
+  options:
+    - text: Sign in with Government Gateway
+      url: https://www.tax.service.gov.uk/gg/sign-in?continue=%2Fcheck-your-state-pension%2Faccount&origin=nisp-frontend&accountType=individual
+      hint_text: You’ll have a user ID if you’ve signed up to do things like file your Self Assessment tax return online.
+    - text: Sign in with GOV.UK Verify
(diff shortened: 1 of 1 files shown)


barrucadu:
The current start button uses the [cross-domain tracking stuff](https://github.com/alphagov/static/blob/master/app/assets/javascripts/modules/cross-domain-tracking.js) powered by the `data-tracking-code` and `data-tracking-name` attributes:

```
<a class="gem-c-button govuk-button govuk-button--start"
   role="button"
   rel="external"
   data-module="cross-domain-tracking"
   data-tracking-code="UA-43414424-1"
   data-tracking-name="transactionTracker"
   href="https://www.tax.service.gov.uk/check-your-state-pension?_ga=2.186671915.222892899.1536675418-68097171.1526482935">Start now</a>
```

I imagine you want to keep this?

JHjava:
Yes, they want Google Analytics data on these pages. Would that require any additions here? We will need to carry out tests on these pages before it is connected to the service page.

barrucadu:
I think so, but I'm not sure exactly what the change needs to be: https://govuk-publishing-components.herokuapp.com/component-guide/button/with_data_attributes

JHjava:
Who would be best to speak to about this?

JHjava:
I assume the other services user service_sign_in pages use GA to monitor, but can't spot anything distinct in there.

barrucadu:
Maybe ask in `#frontend`

PR #944, opened 2018-09-20
Allow govspeak buttons for signin pages to be trackable

steventux:
https://trello.com/c/CbTYssq1/478-enable-cross-domain-tracking-for-cysp-govuk-sign-in-pages

Needed in addition to https://github.com/alphagov/publisher/pull/941 so that buttons in govspeak content are also trackable.


steventux:
After some discussion with the team this approach isn't going to achieve what we need.

PR #950, opened 2018-09-26
Update check-state-pension.cy.yaml

barrucadu:
gwirio-wladwriaeth-pensiwn is not a page


barrucadu:
Forgot to change the commit message.  I'll do that.

PR #985, opened 2018-12-13
HUB-378: Update Verify redirect links

jakubmiarka:
The Verify Hub supports journey hints to redirect users to the right journey flows
(registration, sign-in or eIDAS). However, the required hints are not being
implemented/passed by the services. We have developed a mechanism which
allows us to record the hint value (the option they selected on the GOV.UK
sign-in pages) by 'bouncing' the user via Verify first.

### DIFF
--- a/lib/service_sign_in/check-update-company-car-tax.cy.yaml
+++ b/lib/service_sign_in/check-update-company-car-tax.cy.yaml
@@ -8,10 +8,10 @@ choose_sign_in:
       url: https://www.tax.service.gov.uk/gg/sign-in?continue=/paye/company-car/details&accountType=individual&origin=PTA-companycar
       hint_text: Bydd gennych Ddynodydd Defnyddiwr (ID) os ydych wedi cofrestru i wneud pethau fel cyflwyno'ch Ffurflen Dreth Hunanasesiad ar-lein.
     - text: Defnyddio GOV.UK Verify
-      url: https://www.tax.service.gov.uk/paye/company-car/start-verify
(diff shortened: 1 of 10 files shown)


lib/service_sign_in/check-update-company-car-tax.cy.yaml
```diff
@@ -8,10 +8,10 @@ choose_sign_in:
       url: https://www.tax.service.gov.uk/gg/sign-in?continue=/paye/company-car/details&accountType=individual&origin=PTA-companycar
       hint_text: Bydd gennych Ddynodydd Defnyddiwr (ID) os ydych wedi cofrestru i wneud pethau fel cyflwyno'ch Ffurflen Dreth Hunanasesiad ar-lein.
     - text: Defnyddio GOV.UK Verify
-      url: https://www.tax.service.gov.uk/paye/company-car/start-verify
+      url: https://www.signin.service.gov.uk/initiate-journey/hmrc-company-car-tax?journey_hint=uk_idp_sign_in
```
chrisclayson: This URL doesn't work, is the simpleId correct?

chrisclayson: The simpleId is correct but we don't have the `headlessStartpage` defined for this service in config

jakubmiarka: this page is not live anyway, the service has been taken down with verify. The headless redirects to GG login anyway

PR #1010, opened 2019-01-22
Remove Royal Mail and Citizensafe

bjgill:
As of 1st March, they will no longer be acting as identity providers for Verify.

See https://trello.com/c/vi7RV11M


bjgill:
It looks as if this is a new product page or one we previously missed:

- [ ] https://www.gov.uk/claim-rural-payments/sign-in/prove-identity

sihugh:
Getting CI to run now

sihugh:
I won't publish these until after 12:00

PR #1023, opened 2019-02-13
Upgrade Ruby to 2.6.1

edwardkerry:
Upgrade Ruby to 2.6.1, see commits for more details.


elliotcm:
@edwardkerry This is I think ready to go now

PR #1039, opened 2019-03-27
Update knowledge api minimum wage rates

sihugh:
These values have been updated to match the content here: https://www.gov.uk/national-minimum-wage-rates#current-rates

The knowledge_api is used to tell Alexa about important values on GOV.UK so we should keep this up to date. (Check it out at https://www.gov.uk/api/content/knowledge-alpha)

We'll need to republish the content item on 1st April by running:
```
  rake publishing_api:publish_knowledge
```
against publisher on a backend box.


sihugh:
Thanks! 🎉

PR #1114, opened 2019-08-30
Update regex and add validation for redirect_url

AgaDufrat:
When a user enters url in an invalid format (eg ‘redirect-to’, ‘www.gov.uk/redirect-to', ‘gov.uk/redirect-to') “Something went wrong” error was displayed, the Edition status was set to “Archived” but the page was available to view and no redirect was created.

This makes the `make_govuk_url_relative` more resilient and makes it handle input such as ‘www.gov.uk/redirect-to', ‘gov.uk/redirect-to'.
A validations is also added to ensure that `make_govuk_url_relative` returns url in a correct format.
When a user enters a string that cannot be correctly formatted, a message will be displayed.
 
**Related Zendesk tickets**
- https://govuk.zendesk.com/agent/tickets/3782482
- https://govuk.zendesk.com/agent/tickets/3769569

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -213,7 +213,12 @@ class EditionsController < InheritedResources::Base
   def process_unpublish
     edition = Edition.find(params[:id])
     artefact = edition.artefact
-    success = UnpublishService.call(artefact, current_user, redirect_url)
+
+    if validate_redirect(redirect_url) || redirect_url.blank?
+      success = UnpublishService.call(artefact, current_user, redirect_url)
+    else
+      flash[:danger] = "Redirect path is invalid. #{description(resource)} has not been unpublished."
+    end
 
     if success
       notice = "Content unpublished"
@@ -343,7 +348,12 @@ private
   end
 
   def make_govuk_url_relative(url = "")
-    url.sub(%r{^https?://(www\.)?gov\.uk/}, "/")
+    url.sub(%r{^(https?://)?(www\.)?gov\.uk/}, "/")
+  end
+
+  def validate_redirect(redirect_url)
+    regex = /(\/([a-z0-9]+-)*[a-z0-9]+)+/
+    redirect_url =~ regex
   end
 
   def tagging_update_form
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -57,6 +57,22 @@
       "user_input": "params[:list]",
       "confidence": "High",
       "note": "The params[:list] argument is already checked before the template gets rendered by @presenter.acceptable_list?(@list)."
+    },
+    {
+      "warning_type": "Remote Code Execution",
+      "warning_code": 110,
+      "fingerprint": "9ae68e59cfee3e5256c0540dadfeb74e6b72c91997fdb60411063a6e8518144a",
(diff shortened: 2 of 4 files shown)


barrucadu:
Ah, it's this issue again:

> Confidence: Medium
> Category: Remote Code Execution
> Check: CookieSerialization
> Message: Use of unsafe cookie serialization strategy `:marshal` might lead to remote code execution
> Code: Rails.application.config.action_dispatch.cookies_serializer = :marshal
> File: config/initializers/cookies_serializer.rb
> Line: 5

That `:marshal` needs changing to `:hybrid` and the warning ignored - that could either be done in this PR or a separate one.  See also https://github.com/alphagov/whitehall/pull/4932

app/controllers/editions_controller.rb
```diff
@@ -343,7 +348,12 @@ def redirect_url
   end
 
   def make_govuk_url_relative(url = "")
-    url.sub(%r{^https?://(www\.)?gov\.uk/}, "/")
+    url.sub(%r{^(https?://)?(www\.)?gov\.uk/}, "/")
+  end
+
+  def validate_redirect(redirect_url)
+    regex = /\A\/([a-z0-9]+-)*[a-z0-9]+\z/
```
barrucadu: This only allows a `/` at the start, which would prevent redirecting to (eg) `/topic/keeping-farmed-animals`

AgaDufrat: Good point! Thanks. Should be ok now 🤞

app/controllers/editions_controller.rb
```diff
@@ -343,7 +348,12 @@ def redirect_url
   end
 
   def make_govuk_url_relative(url = "")
-    url.sub(%r{^https?://(www\.)?gov\.uk/}, "/")
+    url.sub(%r{^(https?://)?(www\.)?gov\.uk/}, "/")
+  end
+
+  def validate_redirect(redirect_url)
+    regex = /(\/([a-z0-9]+-)*[a-z0-9]+)+/
+    redirect_url =~ regex
```
barrucadu: I could be missing something but I'm not sure this is right:

```
irb(main):005:0> "/hello/world" =~ /(\/([a-z0-9]+-)*[a-z0-9]+)+/
=> 0
```

Should the final line be `redirect_url =~ regex == 0`?  The `redirect_url` is valid if the regex matches in position 0.

AgaDufrat: I tested the previous version in integration and `redirect_url =~ regex` worked (I got the 'Redirect path is invalid' message when I entered 'redirect-to'). 
I will check `redirect_url =~ regex == 0`

barrucadu: Actually it seems it was just me not understanding Ruby, 0 is truthy:

```
irb(main):015:0> if 0 || false
irb(main):016:1> "truthy"
irb(main):017:1> else
irb(main):018:1> "falsey"
irb(main):019:1> end
=> "truthy"
```

The regex is too lenient for what you're going for I think:

```
irb(main):021:0> "////hello/world-------" =~ /(\/([a-z0-9]+-)*[a-z0-9]+)+/
=> 3
```

But I'm not convinced that's a problem, it doesn't need to be terribly strict.

PR #1176, opened 2019-12-31
God's work

sihugh:
In this PR, I fix a couple of deprecation warning messages from gds-api-adapters and delete some unused models.

## Delete rendered manuals and prerendered entities

These were moved into specialist publisher (back in the day when it was the app that we now call manuals publisher).

This model code was moved into publisher along with all the others when we got rid of [govuk_content_models](https://github.com/alphagov/govuk_content_models/blob/master/app/models/rendered_manual.rb) but nothing actually references it anymore.

There are 0 rendered manuals in the publisher database:

```
irb(main):001:0> RenderedManual.all.count
=> 0
irb(main):002:0> Artefact.where(kind: "rendered_manual").count
=> 0
```

Prerendered entity is only referenced by rendered manual now (it was used by specialist documents at one point when those were also in publisher), so we should be OK to remove those too.


sihugh:
ps it's entirely possible that I've missed the point on rendered manuals and prerendered entities, so feel free to poke at stuff.

PR #1196, opened 2020-02-04
Remove deleted parts from the DOM after successful AJAX requests

edwardkerry:
When editing an edition, updates are made via an ajax request. Upon deleting a part, the part is hidden in the DOM. Although marked as "destroyed" in the DOM and successfully deleted from the database, subsequent update requests include the part details in the params. This caused a database error as the part no longer exists, resulting in a "Something gone's wrong" message. The user would have to refresh their page and lose any edits made since a deletion.

This PR ensures that, after a successful update, any deleted parts are not just hidden but removed from the DOM, to prevent their details being included in request params.

[trello](https://trello.com/c/ibsbiyZ4/1716-5-investigate-and-fix-the-bug-that-deletes-parts-in-publisher)

### DIFF
--- a/app/assets/javascripts/modules/ajax_save_with_parts.js
+++ b/app/assets/javascripts/modules/ajax_save_with_parts.js
@@ -15,6 +15,7 @@
             updatePart(parts[i]);
           }
         }
+        removeDeletedParts();
       }
 
       function error(evt, response) {
@@ -24,6 +25,11 @@
         if (partErrors) {
           $.each(partErrors[0], showPartErrors);
         }
+      }
+
+      function removeDeletedParts() {
+        var deletedParts = $("[id*='__destroy'][value='1']")
+        deletedParts.parents(".fields").remove();
       }
 
       function showPartErrors(partKey, partErrors) {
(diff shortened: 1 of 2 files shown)


cbaines:
> When editing an edition, updates are made via an ajax request. Upon deleting a part, the part is hidden in the DOM. Although marked as "destroyed" in the DOM and successfully deleted from the database, subsequent update requests include the part details in the params. This caused a database error as the part no longer exists, resulting in a "Something gone's wrong" message. The user would have to refresh their page and lose any edits made since a deletion.
> 
> This PR ensures that, after a successful update, any deleted parts are not just hidden but removed from the DOM, to prevent their details being included in request params.

If this is useful to say, it's probably useful to have at least some of it in the commit message.

app/assets/javascripts/modules/ajax_save_with_parts.js
```diff
@@ -26,6 +27,11 @@
         }
       }
 
+      function removeDeletedParts() {
+        var deletedParts = $("[id*='__destroy'][value='1']")
+        deletedParts.parents(".fields").remove();
+      }
```
cbaines: I haven't tried this out, but it looks like it'll work to me :+1:

edwardkerry:
> If this is useful to say, it's probably useful to have at least some of it in the commit message.

I've amended the commit

cbaines:
> > If this is useful to say, it's probably useful to have at least some of it in the commit message.
> 
> I've amended the commit

Awesome :+1:

PR #1212, opened 2020-02-19
Fix url in skip review emails

beccapearce:
The url in the skip review email was using the mailer default url (which is gov.uk) this corrects it by using plek to get the publisher url to use as the host.

https://trello.com/c/jUuJwRLW

### DIFF
--- a/app/mailers/noisy_workflow.rb
+++ b/app/mailers/noisy_workflow.rb
@@ -15,6 +15,7 @@ class NoisyWorkflow < ApplicationMailer
 
   def skip_review(action)
     @edition = action.edition
+    @edition_url = edition_url(@edition.id, host: Plek.find("publisher"), external: true)
     mail(
       to: EMAIL_GROUPS[:force_publish_alerts],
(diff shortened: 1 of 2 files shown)


app/mailers/noisy_workflow.rb
```diff
@@ -15,6 +15,7 @@ def make_noise(action)
 
   def skip_review(action)
     @edition = action.edition
+    @edition_url = edition_url(@edition.id, host: Plek.find("publisher"))
```
cbaines: The default host not being Publisher is weird...

Anyway, you'll probably want to pass `external: true` to Plek, as this'll get you a "external" URL. In Carrenza, the external URLs are the same as the internally (app to app) ones, but in AWS they're different:

```
irb(main):007:0> Plek.find('publisher')
=> "https://publisher.integration.govuk-internal.digital"
irb(main):008:0> Plek.find('publisher', external: true)
=> "https://publisher.integration.publishing.service.gov.uk"
```

PR #1229, opened 2020-03-10
Make entire repository pass Rubocop linting

ChrisBAshton:
This is depended on by https://github.com/alphagov/govuk-jenkinslib/pull/66.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,6 +7,10 @@ inherit_mode:
   merge:
     - Exclude
 
+AllCops:
+  Exclude:
+    - db/migrate/201*.rb
+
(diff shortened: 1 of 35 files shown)


.rubocop.yml
```diff
@@ -7,6 +7,10 @@ inherit_mode:
   merge:
     - Exclude
 
+AllCops:
+  Exclude:
+    - db/migrate/201*.rb
```
kevindew: You seem to have fixed issues in lots of these, is this still needed?

ChrisBAshton: Yes, unfortunately there are a number of issues that can't be fixed automatically.

PR #1232, opened 2020-03-13
[DO NOT MERGE] Disable publisher cron jobs for migration

sarahseewhy:
* Disables cron jobs which trigger the `mail_fetcher` script and `reports:generate` rake task.

* These changes will automatically update the `deploy` user crontab when the application is deployed using Jenkins.

* `mail_fetcher` runs every five minutes and updates editions in review with responses from fact checkers across government.

* `reports:generate` rake task runs hourly and generates a mainstream report (see here:  https://publisher.publishing.service.gov.uk/reports).

* Note: these cron jobs are configured using the `whenever` gem and **not** Puppet.


cbaines:
I guess this can work for you without ever being merged right? You can just rebase on master, then deploy the branch to Carrenza right before you dump the data for the move to AWS.

sarahseewhy:
@cbaines Indeed. It would be good to test this in Staging.

thomasleese:
@sarahseewhy Can this PR be closed now? I think it might not be needed anymore since the migration is done!

sarahseewhy:
Yes, thanks for following up!

PR #1243, opened 2020-04-03
Add MultiNoisyWorkflow to loop over addresses and instantiate NoisyWorkflows for each

benjamineskola:
Method based on https://github.com/alphagov/short-url-manager/pull/451

WIP: I still need to add new tests including some to replace those made obsolete by this change.

https://trello.com/c/QYkWTpbv/1889-publisher-send-one-email-per-recipient


issyl0:
`MultiNoisyWorkflow`? Whatever next...

PR #1256, opened 2020-04-28
Look for ID in subject field of fact check replies

benjamineskola:
This is in addition to checking the `to` address and in preference to it, in order to support Notify which doesn't allow setting unique reply-to addresses.

https://trello.com/c/MR3jrlX2/1878-5-publisher-replace-ses-in-email-receipt-workflow-%F0%9F%8D%90

### DIFF
--- a/config/fact_check.yml
+++ b/config/fact_check.yml
@@ -4,7 +4,7 @@
 # If we ever want to change this format, we'll need to support some notion of
 # legacy formats, so we still pick up emails to old addresses
 address_format: <%=ENV.fetch("FACT_CHECK_ADDRESS_FORMAT", "factcheck+dev-{id}@alphagov.co.uk") %>
-subject_format: <%=ENV.fetch("FACT_CHECK_SUBJECT_FORMAT", "Fact Checked [dev-{id}]") %>
+subject_format: <%=ENV.fetch("FACT_CHECK_SUBJECT_FORMAT", "\\[.+?\\] GOV.UK preview of new edition \\[dev-(?<id>.+?)\\]") %>
 
 fetcher:
   address: <%=ENV.fetch("FACT_CHECK_ADDRESS", "imap.gmail.com") %>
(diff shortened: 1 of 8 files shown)


config/fact_check.yml
```diff
@@ -4,6 +4,7 @@
 # If we ever want to change this format, we'll need to support some notion of
 # legacy formats, so we still pick up emails to old addresses
 address_format: <%=ENV.fetch("FACT_CHECK_ADDRESS_FORMAT", "factcheck+dev-{id}@alphagov.co.uk") %>
+subject_format: <%=ENV.fetch("FACT_CHECK_SUBJECT_FORMAT", "Fact Checked [dev-{id}]") %>
```
nicholsj, line 7: This needs to make sense to the recipient of the original fact check email. I suggest it's as close as possible to the current subject, but including the ID. 

Current format is `‘[content title]’ GOV.UK preview of new edition` 
eg ‘[Car and trailer driving test]’ GOV.UK preview of new edition

test/unit/fact_check_config_test.rb
```diff
@@ -4,67 +4,133 @@
...
   should "accept item IDs that aren't strings" do
     # For example, Mongo IDs, but let's not tie this test to Mongo
-    config = FactCheckConfig.new("factcheck+{id}@example.com")
+    config = FactCheckConfig.new("factcheck+{id}@example.com", "valid subject [{id}]")
     assert_equal "factcheck+1234@example.com", config.address(1234)
   end
+
+  should "fail on a nil subject format" do
+    assert_raises ArgumentError do
+      FactCheckConfig.new("factcheck+{id}@example.com", nil)
```
thomasleese, line 73: It might be worth pulling out the valid addresses and valid subjects into a variable, it might also make the tests clearer to read as instead of having to know it's a valid address, it could just said `valid_address`.

config/fact_check.yml
benjamineskola: That's a good point — currently, the [code](https://github.com/alphagov/publisher/pull/1256/files#diff-b5c2ae4ebc91d294985cc4742f1889abR20) assumes that the ID is the only changeable part of the input but this won't be the case. I'll have to think about how to implement that.

PR #1259, opened 2020-04-30
Make the ID prefix in the subject configurable

benjamineskola:
This is expected to be configured by Puppet.

The current code uses an email address pattern in which the environment    (production/staging/integration) is embedded, and uses this both to    parse incoming emails and generate outgoing emails (the environment is    part of an opaque string as far as the app is concerned).

However, here I've made the prefix configurable but not the subject as a    whole: partly because the subject was already defined in code and not    configurable through Puppet, and partly because trying to both parse and    generate from the same string is needlessly complex.

The prefix can be left blank (e.g., in production).

https://trello.com/c/MR3jrlX2/1878-5-publisher-replace-ses-in-email-receipt-workflow-%F0%9F%8D%90

### DIFF
--- a/app/mailers/noisy_workflow.rb
+++ b/app/mailers/noisy_workflow.rb
@@ -23,11 +23,12 @@ class NoisyWorkflow < ApplicationMailer
   def request_fact_check(action, recipient_email)
     @edition = action.edition
     fact_check_address = @edition.fact_check_email_address
+    fact_check_prefix = ENV.fetch("FACT_CHECK_SUBJECT_PREFIX", "")
     mail(
       to: recipient_email,
       reply_to: fact_check_address,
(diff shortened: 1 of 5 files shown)


app/mailers/noisy_workflow.rb
```diff
@@ -23,11 +23,12 @@ def skip_review(action, recipient_email)
   def request_fact_check(action, recipient_email)
     @edition = action.edition
     fact_check_address = @edition.fact_check_email_address
+    fact_check_prefix = ENV.fetch("FACT_CHECK_SUBJECT_PREFIX", "")
```
thomasleese, line 26: Is there some way we can get this from `Publisher::Application.fact_check_config` directly? It feels like something that might be forgotten if we decide to change this configuration somehow.

PR #1263, opened 2020-05-05
Use mail-notify gem to deliver via GOV.UK Notify

benjamineskola:
This re-applies the changes to send emails via Notify (cf. #1238, #1249) and takes into account what we have discovered since then regarding the fact-check workflow and reply-to.

In particular, this now *does not* attempt to set the from/reply-to address at all: that must be done in the Notify preferences.

https://trello.com/c/Qy5yaTHt/1951-publisher-use-notify-instead-of-amazon-ses

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,7 @@ gem "inherited_resources"
 gem "jquery-ui-rails", "~> 6.0"
 gem "kaminari", "~> 1.2"
 gem "kaminari-mongoid", "1.0.1"
+gem "mail-notify"
 gem "mlanett-redis-lock", "0.2.7" # Only used in some importers
 gem "momentjs-rails", "2.20.1"
 gem "mongo", "2.4.3"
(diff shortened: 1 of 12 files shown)


AlanGabbianelli:
It would be useful to have this GitHub description in the commit message itself.

benjamineskola:
I'm thinking that the test procedure for this will look like:

- Deploy the branch to integration;
- Request a fact check;
- Confirm that it's still sending the correct subject heading;
- Merge and deploy to staging;
- Repeat the test;
- Deploy to production.

We should probably also confirm with Content when they'd be okay with this going live, since it will affect the emails that external factcheckers see. (And consider rewording the text of the email they receive with a warning about the subject header.)

benjamineskola:
Looks like this works in integration; I've just remembered, though, that there needs to be a formatting change to work with Notify's HTML conversion, so I'll add that into this PR. (I tested the syntax change when testing that the emails from integration work as expected.)

PR #1269, opened 2020-05-12
Use mail-notify gem to deliver via GOV.UK Notify

benjamineskola:
This is just reapplying the changes that were made previously (cf. #1238, #1249, #1263), taking into account what we've subsequently learned about the fact-check workflow and Notify's reply-to behaviour.  In particular: this no longer sets the from address, which is defined by Notify and cannot be overridden, and the reply-to address must be specified as a UUID referring to an address that has already been configured in the Notify dashboard.

This also involved some alterations to the template evaluation and tests. We had to start explicitly making the helpers available to the template --- it's not clear why this had succeeded previously. It also means stubbing the bank holidays API endpoint, as this gets requested when evaluating the template.

https://trello.com/c/Qy5yaTHt/1951-publisher-use-notify-instead-of-amazon-ses

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,7 @@ gem "inherited_resources"
 gem "jquery-ui-rails", "~> 6.0"
 gem "kaminari", "~> 1.2"
 gem "kaminari-mongoid", "1.0.1"
+gem "mail-notify"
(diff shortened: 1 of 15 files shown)


benjamineskola:
This now works in integration, where we don't specify a reply-to ID; we'll also need to test in staging to ensure the correct address is being used.

PR #1271, opened 2020-05-14
Improve regex for parsing an edition ID from a subject header

benjamineskola:
Previously the regex would look for an exact match based on what was sent, pinned to the end of the string but not the start (to allow for 'Re:' and similar prefixes). However, it seems that in some cases there can also be text added to the end of the string (e.g., an ID added by the other party's own ticketing system) or even a complete modification of the subject, in which case all we can do is ask the user not to remove the ID.

I've made a few guesses as to what is an appropriate level of cautiousness: for example, this looks for any hexadecimal string surrounded by square brackes (and prefixed with an environment, if one is set); however, I've considered it to be an error if there's more than one match (because we can't be sure we'll pick the right one).

We could also try multiple ways of matching (e.g., look for a more exact match with the subject we sent, and if it fails then any hex string; or look for any hex string and if there are multiple matches then pick the one that seems to be immediately prefixed by our subject, if any). But since we don't know enough about the ways this could fail, it seemed like a premature optimisation; it would be better to revisit this if we start to see a common kind of failure in the future.

https://trello.com/c/KHVxF6ry/1969-3-update-publisher-interface-and-email-text-to-match-new-fact-check-workflow

### DIFF
--- a/lib/fact_check_config.rb
+++ b/lib/fact_check_config.rb
@@ -9,7 +9,7 @@ class FactCheckConfig
     @reply_to_id = reply_to_id
 
     @subject_prefix = subject_prefix.present? ? subject_prefix + "-" : ""
-    subject_format = "‘\\[.+?\\]’ GOV.UK preview of new edition \\[#{@subject_prefix}(?<id>.+?)\\]"
+    subject_format = "\\[#{@subject_prefix}(?<id>[0-9a-f]+)\\]"
 
     @address_prefix, @address_suffix = address_format.split("{id}")
     @address_pattern = Regexp.new(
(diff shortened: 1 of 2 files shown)


benjamineskola:
Do we need to be careful about deploying this? It _should_ only increase the likelihood of emails being processed properly, and the tests show that; but when it can only be tested in production, I'd prefer to err on the side of caution.

thomasleese:
Yeah, although it's a pain, I think we should go through the testing we've been doing for the previous PRs where we do a full end-to-end test of this in production very soon after deploying.

PR #1272, opened 2020-05-14
Warn users not to remove edition id from subject header

benjamineskola:
https://trello.com/c/EgwMYcXj/1987-3-update-publisher-interface-and-email-text-to-match-new-fact-check-workflow


1pretz1:
Text looks fine to me, my only Q would be is there any way to make it more visible, maybe making it bold (if allowed) or moving to the top of the list it's in?

benjamineskola:
> maybe making it bold (if allowed)

Notify's markup doesn't allow bold as they think it makes the text less readable.

> or moving to the top of the list it's in (don't know if that'd make much difference tbh)?

It's possible, but the position (third item in that list) was what JP in Content suggested when I asked.

PR #1274, opened 2020-05-15
Read reply-to address from a separate environment variable

benjamineskola:
This allows us to stop showing the ID in the reply-to address we're giving to users, without actually yet disabling it for incoming messages.

The caveat is, though, that it should be compatible with both the `address_format` (at least, compatible with the existing gmail filters that assume that address format) and matching the address that the `reply_to_id` refers to. Neither of these is guaranteed to break things if they're different, but inconsistency is likely to complicate things and it's difficult or impossible to reliably ensure that they're compatible.

https://trello.com/c/EgwMYcXj/1987-3-update-publisher-interface-and-email-text-to-match-new-fact-check-workflow

### DIFF
--- a/config/fact_check.yml
+++ b/config/fact_check.yml
@@ -5,6 +5,7 @@
 # legacy formats, so we still pick up emails to old addresses
 address_format: <%=ENV.fetch("FACT_CHECK_ADDRESS_FORMAT", "factcheck+dev-{id}@alphagov.co.uk") %>
 subject_prefix: <%=ENV.fetch("FACT_CHECK_SUBJECT_PREFIX", "") %>
+reply_to_address: <%=ENV.fetch("FACT_CHECK_REPLY_TO_ID", "factcheck+dev@alphagov.co.uk") %>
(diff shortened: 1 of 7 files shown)


config/fact_check.yml
```diff
@@ -5,6 +5,7 @@
 # legacy formats, so we still pick up emails to old addresses
 address_format: <%=ENV.fetch("FACT_CHECK_ADDRESS_FORMAT", "factcheck+dev-{id}@alphagov.co.uk") %>
 subject_prefix: <%=ENV.fetch("FACT_CHECK_SUBJECT_PREFIX", "") %>
+reply_to_address: <%=ENV.fetch("FACT_CHECK_REPLY_TO_ID", "factcheck+dev@alphagov.co.uk") %>
```
thomasleese, line 8: Should this line be:

```suggestion
reply_to_address: <%=ENV.fetch("FACT_CHECK_REPLY_TO_ADDRESS", "factcheck+dev@alphagov.co.uk") %>
```

PR #1285, opened 2020-06-03
Add the reply-to address to the body of the fact-check email and test it

benjamineskola:
This is so that HMRC can manually set the reply address, as their system ignores the reply-to header.

https://trello.com/c/jCpYumNF/2009-include-reply-to-address-in-body-of-publisher-fact-check-emails

### DIFF
--- a/test/integration/edition_workflow_test.rb
+++ b/test/integration/edition_workflow_test.rb
@@ -72,7 +72,25 @@ class EditionWorkflowTest < JavascriptIntegrationTest
 
     fact_check_email = ActionMailer::Base.deliveries.select { |mail| mail.to.include? "user@example.com" }.last
     assert fact_check_email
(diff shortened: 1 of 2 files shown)


test/integration/edition_workflow_test.rb
```diff
@@ -75,6 +75,25 @@ class EditionWorkflowTest < JavascriptIntegrationTest
...
+    ActionMailer::Base.deliveries.clear
+
+    within "#send_fact_check_form" do
+      fill_in "Email address", with: "user@example.com"
+      click_on "Send"
+    end
+
+    fact_check_email = ActionMailer::Base.deliveries.select { |mail| mail.to.include? "user@example.com" }.last
+    assert fact_check_email
+    pp fact_check_email.body.to_s
```
cbaines, line 93: Left over debugging?

PR #1300, opened 2020-06-25
Update create account page content

bethwalton:
- Updated eligibility criteria for Gateway and Verify 
- Added examples of the most popular other services you can use each account for 
- Edited headings and buttons to make it clearer that users need to choose between two options

This is part of work on GOV.UK Verify user journeys, being done by the hub/eIDAS team. The new page content has been approved by HMRC.

[Jira ticket is here. ](https://govukverify.atlassian.net/browse/HUB-660?atlOrigin=eyJpIjoiMTkxNzU0MzY5YTc1NDc0Yzg5ZGE4NDQ3ZjlkMjAxMjUiLCJwIjoiaiJ9)

### DIFF
--- a/lib/service_sign_in/check-income-tax.en.yaml
+++ b/lib/service_sign_in/check-income-tax.en.yaml
@@ -17,37 +17,35 @@ choose_sign_in:
       slug: create-account
       hint_text: If you do not already have one of these accounts, we'll help you choose whether to use Government Gateway or GOV.UK Verify.
 create_new_account:
-  title: Create an account
+  title: Choose a way to prove your identity
   slug: create-account
   body: |
-    To use this service, you need to create either a Government Gateway or GOV.UK Verify account. These are used to help fight identity theft.
+    You need to confirm who you are before you can check your income tax.
 
-    Once you have an account, you can use it to access other government services online.
+    To do this, you can create either a Government Gateway user ID or a GOV.UK Verify identity account.
 
-    ## Choose a way to prove your identity
+    ## Government Gateway
 
-    ### Government Gateway
+    You'll need:
 
-    Registering with Government Gateway usually takes about 10 minutes. It works best if you have:
+    - your National Insurance number or UK address
+    - a recent payslip or P60 or a valid UK passport
+    
+    You can use your user ID for things like your personal or business tax account, Self Assessment, Corporation Tax, PAYE for employers and VAT.
 
-    * your National Insurance number
-    * a recent payslip or P60 or a valid UK passport
+    {button}[Choose Government Gateway](https://www.tax.service.gov.uk/gg/sign-in?continue=/check-income-tax/what-do-you-want-to-do&accountType=individual&origin=PTA-CYIT-CY){/button}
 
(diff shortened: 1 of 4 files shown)


lib/service_sign_in/check-income-tax.en.yaml
```diff
@@ -17,37 +17,35 @@ choose_sign_in:
...
 
-    Registering with Government Gateway usually takes about 10 minutes. It works best if you have:
+    - your National Insurance number or UK address
+    - a recent payslip or P60 or a valid UK passport
+    
+    You can use your user ID for things like your personal or business tax account, Self Assessment, Corporation Tax, PAYE for employers and VAT.
 
-    * your National Insurance number
-    * a recent payslip or P60 or a valid UK passport
+    {button}[Choose Government Gateway](https://www.tax.service.gov.uk/gg/sign-in?continue=/check-income-tax/what-do-you-want-to-do&accountType=individual&origin=PTA-CYIT-CY){/button}
```
ChrisBAshton, line 36: I know you just moved this code a little further up, but I suspect the `origin` is wrong here - should it be `PTA-CYIT-EN`?

lib/service_sign_in/check-state-pension.en.yaml
```diff
@@ -20,34 +20,35 @@ choose_sign_in:
...
 
-    Registering with Government Gateway usually takes about 10 minutes. It works best if you have:
+    - your National Insurance number or UK address
+    - a recent payslip or P60 or a valid UK passport
+    
+    You can use your user ID for things like your personal or business tax account, Self Assessment, Corporation Tax, PAYE for employers and VAT.
 
-    * your National Insurance number
-    * a recent payslip or P60 or a valid UK passport
+    <a role="button" class="gem-c-button govuk-button" href="https://www.tax.service.gov.uk/gg/sign-in?continue=/check-your-state-pension&accountType=individual" data-module="cross-domain-tracking" data-tracking-code="UA-43414424-1" data-tracking-name="govspeakButtonTracker" data-tracking-track-event="true">Choose Government Gateway</a>
```
ChrisBAshton, line 39: I wonder why this file uses raw HTML but the previous one uses GovSpeak? 🤔

bethwalton: Yeah I thought that was strange... I've never updated these pages before and I didn't want to mess with anything I didn't need to though, so I left it alone!

lib/service_sign_in/check-income-tax.en.yaml
bethwalton: CY as in the Welsh version? I don't know enough to say for sure I'm afraid. 

I just looked at the Welsh version expecting that to have the same link, but it looks like it's the wrong one altogether - it mentions 'company car' which is a different service... 

I can go back to HMRC on this.

ChrisBAshton: No problem. Let's merge this for now, as this is the tracking link currently used on production anyway.

PR #1301, opened 2020-06-25
Delete out of office emails and send uprocessed emails count to Graphite

AlanGabbianelli:
See individual commit messages for more detail.

Trello card: https://trello.com/c/NbgFTMR4/2022-5-alert-on-fact-check-replies-that-werent-parsed

### DIFF
--- a/lib/fact_check_email_handler.rb
+++ b/lib/fact_check_email_handler.rb
@@ -39,13 +39,16 @@ class FactCheckEmailHandler
 
   # takes an optional block to call after processing each message
   def process
+    unprocessed_emails_count = 0
     Mail.all(read_only: false, delete_after_find: true) do |message|
-      message.skip_deletion unless process_message(message)
+      message.mark_for_delete = process_message(message)
+      unprocessed_emails_count += 1 unless message.is_marked_for_delete?
       begin
         yield(message) if block_given?
       rescue StandardError => e
         GovukError.notify(e)
       end
     end
+    GovukStatsd.gauge("unprocessed_emails.count", unprocessed_emails_count)
   end
 end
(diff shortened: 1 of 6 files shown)


lib/fact_check_email_handler.rb
```diff
@@ -39,13 +39,16 @@ def process_message(message)
 
   # takes an optional block to call after processing each message
   def process
+    email_count = 0
     Mail.all(read_only: false, delete_after_find: true) do |message|
-      message.skip_deletion unless process_message(message)
+      message.mark_for_delete = process_message(message)
```
benjamineskola, line 44: I wonder if these lines could be combined into a single conditional. Perhaps something like:

```ruby

if message.mark_for_delete = process_message(message)
  email_count += 1
end
```

AlanGabbianelli: That would work too, but I don't think it would add anything in terms of clarity or brevity.

lib/fact_check_email_handler.rb
```diff
@@ -39,13 +39,16 @@ def process_message(message)
...
+      email_count += 1 unless message.is_marked_for_delete?
       begin
         yield(message) if block_given?
       rescue StandardError => e
         GovukError.notify(e)
       end
     end
+    GovukStatsd.gauge("email.count", email_count)
```
thomasleese, line 52: I think it might be better if this was named something like `unprocessed_email_count` to make it clear that it's the number of emails that weren't able to be processed, i.e. deleted.

AlanGabbianelli: Technically those emails were processed but not deleted - how about this?
`GovukStatsd.gauge("undeleted_emails.count", undeleted_emails_count)`

thomasleese: Yeah, that's good too.

AlanGabbianelli: I changed the code to follow your naming suggestion. Even though we technically process the emails and "undeleted" would be more accurate, the language around this is all about "processing" the messages (method names, etc), so I think it’s better to remain consistent.

PR #1313, opened 2020-07-17
Generate an auth bypass token for each edition

thomasleese:
Currently the auth bypass token is based off the ID of the edition, this means it's not possible to invalidate old tokens, if they were leaked for example. To support this, I've create a Rake task that populates the new field based on the existing method so existing editions will continue to work, but from now onwards, the tokens will be generated from a random UUID. See commits for more details.

### DIFF
--- a/app/helpers/paths_helper.rb
+++ b/app/helpers/paths_helper.rb
@@ -7,7 +7,7 @@ module PathsHelper
     path = edition_front_end_path(edition)
 
     if should_have_auth_bypass_id?(edition)
-      token = jwt_token(sub: edition.auth_bypass_id)
+      token = jwt_token(sub: edition.temp_auth_bypass_id)
       path << "?token=#{token}"
     end
     path
@@ -46,6 +46,6 @@ protected
 private
 
   def should_have_auth_bypass_id?(edition)
-    %w[published archived].exclude?(edition.state) && edition.auth_bypass_id
+    %w[published archived].exclude?(edition.state) && edition.temp_auth_bypass_id
   end
 end
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -120,6 +120,10 @@ class Edition
     destroy_artefact
   end
 
+  after_create do
+    update(auth_bypass_id: temp_auth_bypass_id)
+  end
+
   index assigned_to_id: 1
   index({ panopticon_id: 1, version_number: 1 }, unique: true)
   index state: 1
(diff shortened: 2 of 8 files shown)


app/helpers/paths_helper.rb
```diff
@@ -46,6 +46,6 @@ def preview_url(_edition)
 private
 
   def should_have_auth_bypass_id?(edition)
-    %w[published archived].exclude?(edition.state) && edition.auth_bypass_id
+    %w[published archived].exclude?(edition.state) && edition.temp_auth_bypass_id
```
kevindew, line 49: It's perculiar how this project has conditionals based on `auth_bypass_id` (another one https://github.com/alphagov/publisher/pull/1313/commits/1ba31512ca20bb142c395b19f6371b88ad4c47dc#diff-8796fcc088e83cf48d2a897bfdc702e2R19) as I don't think it can ever be falsey.

app/models/edition.rb
```diff
@@ -33,6 +33,8 @@ class ResurrectionError < RuntimeError
   field :change_note,          type: String
   field :review_requested_at,  type: DateTime
 
+  field :auth_bypass_id,       type: String, default: -> { SecureRandom.uuid }
```
kevindew, line 36: It looks like this default is made superfluous by the subsequent commit that introduces a callback. Might be better to add this default in at the point when it will be impactful.

app/helpers/paths_helper.rb
thomasleese: Yeah, I noticed that too, I was a bit nervous about removing them in case it changed some logic I wasn't aware of...

app/models/edition.rb
thomasleese: Hmm, true... I'll remove that from this PR.

thomasleese:
@kevindew Yeah, the subsequent PR will be https://github.com/alphagov/publisher/pull/1314.

app/models/edition.rb
thomasleese: Removed this in https://github.com/alphagov/publisher/pull/1313/commits/e9b66c1a81d6e64a5aab0219fe2c120a22a01a9c

app/helpers/paths_helper.rb
kevindew: Sure, I reckon we can snipe them in the follow-up PR. I left comments to that effect.

PR #1314, opened 2020-07-17
Improvements to the auth bypass token

thomasleese:
This follows on from #1313 to make various improvements to the auth bypass token. Specifically it adds a `content_id` field to the token for auditing and adds an expiration date of one month to the token. See individual commits for more details.

### DIFF
--- a/app/helpers/paths_helper.rb
+++ b/app/helpers/paths_helper.rb
@@ -28,7 +28,12 @@ module PathsHelper
 protected
 
   def jwt_token(sub:)
-    JWT.encode({ "sub" => sub }, jwt_auth_secret, "HS256")
+    payload = {
+      "sub" => sub,
+      "iat" => Time.zone.now.to_i,
+      "exp" => 1.month.from_now.to_i,
+    }
+    JWT.encode(payload, jwt_auth_secret, "HS256")
   end
 
   def jwt_auth_secret
--- a/app/helpers/paths_helper.rb
+++ b/app/helpers/paths_helper.rb
@@ -7,9 +7,9 @@ module PathsHelper
     path = edition_front_end_path(edition)
 
     if should_have_auth_bypass_id?(edition)
-      token = jwt_token(sub: edition.auth_bypass_id)
-      path << "?token=#{token}"
+      path << "?token=#{jwt_token(edition)}"
     end
+
     path
   end
 
@@ -27,9 +27,10 @@ module PathsHelper
 
 protected
 
-  def jwt_token(sub:)
+  def jwt_token(edition)
     payload = {
-      "sub" => sub,
+      "sub" => edition.auth_bypass_id,
+      "content_id" => edition.content_id,
       "iat" => Time.zone.now.to_i,
       "exp" => 1.month.from_now.to_i,
     }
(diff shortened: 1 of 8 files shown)


app/models/edition.rb
```diff
@@ -457,8 +453,8 @@ def check_if_archived
     end
   end
 
-  def temp_auth_bypass_id
-    @temp_auth_bypass_id ||= begin
+  def auth_bypass_id
```
kevindew: You'll want to remove this method, assuming it was just an oversight 😄

app/helpers/paths_helper.rb
```diff
@@ -46,6 +46,6 @@ def preview_url(_edition)
 private
 
   def should_have_auth_bypass_id?(edition)
-    %w[published archived].exclude?(edition.state) && edition.temp_auth_bypass_id
+    %w[published archived].exclude?(edition.state) && edition.auth_bypass_id
```
kevindew, line 55: It'd be good to lose the auth_bypass_id conditional if we expect it to be always present.

app/presenters/formats/edition_format_presenter.rb
```diff
@@ -16,8 +16,8 @@ def render_for_publishing_api(republish: false)
     def optional_fields
       fields = {}
 
-      if edition.temp_auth_bypass_id
-        fields[:access_limited] = { auth_bypass_ids: [edition.temp_auth_bypass_id] }
+      if edition.auth_bypass_id
```
kevindew, line 19: We can drop this.

test/unit/edition_test.rb
```diff
@@ -27,19 +27,19 @@ class EditionTest < ActiveSupport::TestCase
     should "return a deterministic hex id if edition is in fact-check state" do
       edition = FactoryBot.create(:edition, state: "fact_check", id: 123)
       edition.artefact.update(kind: "help_page")
-      assert_equal edition.temp_auth_bypass_id, "a665a459-2042-4f9d-817e-4867efdc4fb8"
+      assert_equal edition.auth_bypass_id, "a665a459-2042-4f9d-817e-4867efdc4fb8"
```
kevindew, line 30: These tests will probably need some changing since they're using the predictable auth_bypass_id

thomasleese:
@kevindew Thanks for your review, this PR should be ready for a re-review now!

test/unit/edition_test.rb
```diff
@@ -23,23 +23,18 @@ class EditionTest < ActiveSupport::TestCase
...
 
-    should "return a deterministic hex id if edition is in ready state" do
-      edition = FactoryBot.create(:edition, state: "ready", id: 123)
-      edition.artefact.update(kind: "help_page")
-      assert_equal edition.temp_auth_bypass_id, "a665a459-2042-4f9d-817e-4867efdc4fb8"
-    end
+    second_edition = FactoryBot.create(:edition)
+    assert_is_non_empty_string second_edition.auth_bypass_id
+
+    assert_not_equal edition.auth_bypass_id, second_edition.auth_bypass_id
```
kevindew, line 38: I don't think you really need the previous assert_is_non_empty_string assertions here. By these not being equal you've asserted that at least one is not empty. 

I imagine some other parts of this are covered implicitly by other tests too (I assume schema would raise if they weren't a uuid for example).

PR #1320, opened 2020-07-22
Alert when fact check emails cannot be processed

benthorner:
https://trello.com/c/uTeN3PfS/921-fact-check-emails-not-processed

Previously we used to collect fact check processing errors and send
them to an email address. The code for this appeared to be broken
(it wasn't sending an email), and in any case, we should be sending
all errors to Sentry, so they're in one place.

### DIFF
--- a/script/mail_fetcher
+++ b/script/mail_fetcher
@@ -51,9 +51,4 @@ rescue StandardError => e
   raise
 end
 
-if handler.errors.any?
-  Rails.logger.error(handler.errors.join("\n"))
-  NoisyWorkflow.report_errors(handler.errors).deliver
-end
-
 Rails.logger.info "Finished running MailFetcher in #{Rails.env} mode - #{Time.now.utc}"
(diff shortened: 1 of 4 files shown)


script/mail_fetcher
```diff
@@ -51,9 +51,4 @@ rescue StandardError => e
   raise
 end
 
-if handler.errors.any?
-  Rails.logger.error(handler.errors.join("\n"))
-  NoisyWorkflow.report_errors(handler.errors).deliver
```
ChrisBAshton, line 56: Do we want to move these two lines to `lib/fact_check_email_handler.rb` after notifying Sentry?
Not sure what the impact is of removing them... would be a shame for NoisyWorkflow to be noisy in places and silent in others :)

benthorner: The `NoisyWorkflow` class really puzzles me. It's a mailer that sends notification emails to publisher, but also has this "make_noise" method, [added in 2011 for no documented reason](https://github.com/alphagov/publisher/commit/163b0dea17ecf66dfa2e15bc3349b0992b19f39f). Since then, we appear to have shoved all the other (useful) notifications into this class, such that its original name is inaccurate (and quite unhelpful, frankly).

Anyway, as long as we're capturing the errors (in Sentry), I don't see the value in sending ourselves an email about it - even if it did work.

PR #1333, opened 2020-09-07
Rescue bad request error and display a message

benjamineskola:
https://trello.com/c/AB5sPEMm/2121-5-handle-not-in-team-notify-error-in-mainstream-publisher

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,6 +6,7 @@ class ApplicationController < ActionController::Base
   before_action :authenticate_user!
 
   rescue_from Mongoid::Errors::DocumentNotFound, with: :record_not_found
+  rescue_from Notifications::Client::BadRequestError, with: :notify_bad_request
 
   def template_folder_for(publication)
     tmpl_folder = publication.class.to_s.underscore.pluralize.downcase.gsub("_edition", "")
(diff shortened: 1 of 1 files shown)


app/lib/govuk_content_models/action_processors/base_processor.rb
```diff
@@ -62,6 +62,13 @@ def notify_about_event(new_action, action_name)
         EventNotifierService.request_fact_check(new_action).map(&:deliver_now) if action_name.to_s == "send_fact_check"
         EventNotifierService.resend_fact_check(new_action).map(&:deliver_now) if action_name.to_s == "resend_fact_check"
         EventNotifierService.skip_review(new_action).map(&:deliver_now) if action_name.to_s == "skip_review"
+      rescue Notifications::Client::BadRequestError => e
+        if e.message == "Can’t send to this recipient using a team-only API key"
```
thomasleese, line 66: Is there a way to make this less brittle? Perhaps something like this:

```suggestion
        if e.message ~= /team-only API key/
```

benjamineskola: Actually I wondered about just redirecting to an error page for all of the Notify bad request exceptions, and displaying the error message they provide. It would make the matching less brittle but have the downside of making it harder to control the error message we display.

The error messages are [defined in the documentation](https://docs.notifications.service.gov.uk/ruby.html#error-codes) along with the error codes, but the codes aren't unique. So we could look for `e.code == 400` perhaps — that's shared with the trial-mode error, but I don't think we're likely to see that separately from the team-only error and it would at least point the user in the right direction.

thomasleese (approved):

benjamineskola:
Took a little bit of experimentation but this now works as expected.

thomasleese (approved):

PR #1343, opened 2020-09-23
Look for item ID in body as well as subject of factcheck emails

benjamineskola:
https://trello.com/c/R9ZJGaZR/2147-investigate-how-to-reduce-fact-check-email-alerts-caused-by-edition-id-being-manually-removed

### DIFF
--- a/app/lib/fact_check_config.rb
+++ b/app/lib/fact_check_config.rb
@@ -17,18 +17,30 @@ class FactCheckConfig
     @reply_to_address
   end
 
-  def valid_subject?(subject)
-    subject.scan(@subject_pattern).length == 1
+  def contains_id?(string)
+    string.scan(@subject_pattern).length == 1
   end
 
-  def item_id_from_subject(subject)
-    match = subject.scan(@subject_pattern)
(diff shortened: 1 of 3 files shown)


app/lib/fact_check_config.rb
```diff
@@ -17,18 +17,30 @@ def address
...
   end
+
+  def item_id_from_subject_or_body(subject, body = "")
+    if contains_id?(subject)
+      id = item_id_from_string(subject)
+    elsif contains_id?(body)
+      id = item_id_from_string(body)
+    else
+      raise ArgumentError, "Message does not contain any fact check ID"
+    end
```
thomasleese, line 42: It feels like `item_id_from_string` could be refactored here to return `nil`, and then we could do something like:

```suggestion
    return id if (id = item_id_from_string(subject) || item_id_from_string(body))
    raise ArgumentError, "Message does not contain any fact check ID"
```

Mostly it would avoid having to `.scan` the strings twice.

edwardkerry (approved):
LGTM

PR #1345, opened 2020-09-25
Fix type error in fact-check email handling

benjamineskola:
https://trello.com/c/R9ZJGaZR/2147-investigate-how-to-reduce-fact-check-email-alerts-caused-by-edition-id-being-manually-removed

### DIFF
--- a/test/integration/fact_check_email_test.rb
+++ b/test/integration/fact_check_email_test.rb
@@ -157,6 +157,17 @@ class FactCheckEmailTest < ActionDispatch::IntegrationTest
     assert message.is_marked_for_delete?
   end
 
+  test "should look for fact-check body field" do
+    edition = FactoryBot.create(:answer_edition, state: "fact_check")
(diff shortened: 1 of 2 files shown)


benthorner:
I've tried this and it's picking up the body correctly now 👍. Could we add a test, though?

test/integration/fact_check_email_test.rb
```diff
@@ -136,6 +136,18 @@ def assert_correct_state(key, value, state)
     edition = FactoryBot.create(:answer_edition, state: "fact_check")
     message = fact_check_mail_for(edition, subject: "Fact Checked [#{edition.id}]")
 
+    Mail.stubs(:all).yields(message)
+    pp message.body
```
thomasleese, line 140: I don't think we need this :)

edwardkerry, line 140: is this OK?

PR #1348, opened 2020-10-01
Require govuk_editor to progress editions

thomasleese:
This makes changes to the action processors to ensure that the user has the `govuk_editor` permission before performing the action. See individual commits for more details.

This is the precursor to adding a new `welsh_editor` permission which permits access to certain actions without the `govuk_editor` permission but only for Welsh language documents.

[Trello Card](https://trello.com/c/Hq2qgwiI/2155-8-give-permission-in-publisher-for-welsh-teams-in-departments-to-update-only-welsh-content)


MuriloDalRi (approved):

MuriloDalRi:
Looks good to me!

PR #1369, opened 2020-11-06
Stop logging fact check email errors to Sentry

ChrisBAshton:
We already send an Icinga alert when there is a problem processing
the email: https://github.com/alphagov/govuk-puppet/blob/ce4eb1c7acc732e920e68cc1c3acdc7e5f86f9fb/modules/govuk/manifests/apps/publisher/unprocessed_emails_count_check.pp

And we log an error in Sentry:
https://sentry.io/organizations/govuk/issues/1951927797/?project=202241&query=is%3Aunresolved

We don't need to log it in Sentry too. The question is, which is
the most appropriate vehicle for notifying developers about the
error?

I would argue that it is more important for the Icinga alert to
happen, as we have 2nd line developers actively allocated to
responding to such issues. Errors logged to Sentry (currently)
languish until a developer periodically checks Sentry. Given the
sometimes urgent nature of fact checking, it is better that we
tackle issues quickly when they arise.

However, it is still useful to log the error to the Rails logger, so that
we can search for it in Logit.

:warning: This application is Continuously Deployed: :warning:

- Merged changes are automatically deployed to staging and production.

- Make sure you follow [the guidance for deployments](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) **before** you merge.

- Check your branch is being deployed in the [Release app](https://release.publishing.service.gov.uk/applications/publisher), after merging.

### DIFF
--- a/app/lib/fact_check_email_handler.rb
+++ b/app/lib/fact_check_email_handler.rb
@@ -9,8 +9,6 @@ require_relative "fact_check_mail"
 class FactCheckEmailHandler
   attr_accessor :fact_check_config
 
-  class UnableToProcessError < StandardError; end
-
   def initialize(fact_check_config)
     @fact_check_config = fact_check_config
   end
@@ -21,8 +19,6 @@ class FactCheckEmailHandler
 
     edition_id = @fact_check_config.item_id_from_subject_or_body(message.subject, message.body.to_s)
     FactCheckMessageProcessor.process(message, edition_id)
-  rescue StandardError => e
-    raise UnableToProcessError, "Failed to process message '#{message.subject}': #{e.message}"
   end
 
   # takes an optional block to call after processing each message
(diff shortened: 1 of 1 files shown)


app/lib/fact_check_email_handler.rb
```diff
@@ -33,10 +29,9 @@ def process
       process_message(message)
       message.mark_for_delete = true
       yield(message) if block_given?
-    rescue StandardError => e
+    rescue StandardError
       message.mark_for_delete = false
       unprocessed_emails_count += 1
-      GovukError.notify(e)
```
thomasleese, line 39: I think it would be good to log this exception so we can find out about it in Logit if we're needing to debug.

app/lib/fact_check_email_handler.rb
```diff
@@ -9,8 +9,6 @@
 class FactCheckEmailHandler
   attr_accessor :fact_check_config
 
-  class UnableToProcessError < StandardError; end
```
thomasleese, line 12: If we decide to log it (see my other comment) I think it would be good to keep this exception around so we get a nicer log message and something to search for.

thomasleese (approved):

PR #1378, opened 2020-12-01
Add restrictions for user with `welsh_editor` permissions

edwardkerry:
This PR adds further functionality for users with `welsh_editor` permissions, following https://github.com/alphagov/publisher/pull/1360

Users must have correct permissions to:
*  assign other users to an edition. The assigned user must also have correct permissions. This option is hidden without correct permissions.
* destroy an edition.
* view an edition's admin page.
* see fact-check workflow buttons (the requests these buttons submit have already been addressed in the above PR)

[Trello](https://trello.com/c/Hq2qgwiI/2155-8-give-permission-in-publisher-for-welsh-teams-in-departments-to-update-only-welsh-content)

### DIFF
--- a/app/helpers/tabs_helper.rb
+++ b/app/helpers/tabs_helper.rb
@@ -22,7 +22,11 @@ module TabsHelper
     Edition::Tab.all
   end
 
-  def tabs_for(user)
-    tabs.reject { |tab| tab.name == "unpublish" unless user.govuk_editor? }
+  def tabs_for(user, resource)
+    tabs_to_remove = []
+    tabs_to_remove << "admin" unless user.has_editor_permissions?(resource)
+    tabs_to_remove << "unpublish" unless user.govuk_editor?
+
+    tabs.reject { |tab| tabs_to_remove.include?(tab.name) }
   end
 end
--- a/test/integration/edition_workflow_test.rb
+++ b/test/integration/edition_workflow_test.rb
@@ -570,6 +570,28 @@ class EditionWorkflowTest < JavascriptIntegrationTest
     save_edition_and_assert_success
   end
 
+  test "Welsh editors can assign users to Welsh editions" do
+    guide.artefact.update!(language: "cy")
+
+    login_as("WelshEditor")
+    visit edition_path(guide)
+
+    select "Bob", from: "Assigned to"
+    save_edition_and_assert_success
+    guide.reload
+
+    assert_equal guide.assigned_to, bob
+
+    save_edition_and_assert_success
+  end
+
+  test "Welsh editors cannot assign users to non-Welsh editions" do
+    login_as("WelshEditor")
+    visit edition_path(guide)
+
+    assert page.has_no_content? "Assigned to"
+  end
+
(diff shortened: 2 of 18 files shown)


app/helpers/tabs_helper.rb
```diff
@@ -22,7 +22,12 @@ def tabs
     Edition::Tab.all
   end
 
-  def tabs_for(user)
-    tabs.reject { |tab| tab.name == "unpublish" unless user.govuk_editor? }
+  def tabs_for(user, resource)
+    tabs_to_remove = [
+      user.has_editor_permissions?(resource) ? "" : "admin",
+      user.govuk_editor? ? "" : "unpublish",
+    ]
```
ChrisBAshton, line 29: This is a little bit of a smell to me, as we could end up with `tabs_to_remove == ["", ""]`, which will work but is a bit surprising. I think I'd opt for a different approach, e.g.

```ruby
tabs_to_remove = []
tabs_to_remove << "admin" unless user.has_editor_permissions?(resource)
tabs_to_remove << "unpublish" unless user.govuk_editor?
```

test/unit/tabs_helper_test.rb
```diff
@@ -40,15 +40,23 @@ class TabsHelperTest < ActionView::TestCase
...
 
     should "return all tabs except `unpublish` if user is not govuk_editor" do
-      user = FactoryBot.create(:user)
-      assert_equal %w[unpublish], (tabs - tabs_for(user)).map(&:name)
+      guide = FactoryBot.create(:guide_edition, :welsh)
+      user = FactoryBot.create(:user, :welsh_editor)
+      assert_equal %w[unpublish], (tabs - tabs_for(user, guide)).map(&:name)
+    end
+
+    should "excludes `admin` tab if a user has insufficient editor permissions" do
```
ChrisBAshton, line 56:
```suggestion
    should "exclude `admin` tab if a user has insufficient editor permissions" do
```

test/integration/edition_workflow_test.rb
```diff
@@ -592,6 +582,68 @@ class EditionWorkflowTest < JavascriptIntegrationTest
...
+    edition = FactoryBot.create(:edition, :fact_check, :welsh)
+    login_as("WelshEditor")
+
+    visit_edition edition
+
+    assert page.has_content?("We’re awaiting a response")
+    assert page.has_css?(".btn.btn-info", text: "Needs more work")
+  end
+
+  test "Welsh editors may not request more work for ready Welsh editions" do
```
ChrisBAshton, line 627:
```suggestion
  test "Welsh editors may not request more work for fact checked edition" do
```

ChrisBAshton (commented):
LGTM, just a couple of small suggestions/corrections below:

app/helpers/tabs_helper.rb
edwardkerry: I did this, and then changed my mind for some reason. thanks!

ChrisBAshton (approved):
LGTM 👍 thanks Ed!

thomasleese (approved):
Good work on this!

PR #1380, opened 2020-12-04
Remove option to sign into Verify with a European ID

idrop:
## What
Remove the option to sign into Verify with an eID from a connected European country

## Why
Due to EU Exit transition

## How to review
Searched for the `?journey_hint=eidas_sign_in` url parameter, which routes a user to the Verify country picker page.
Deleted these entries.

I was unable to get the test suite to run locally, which is why this is in draft.


idrop:
@thomasleese 👋 This is the content I mailed you about that the Verify Helping Users team would like removed, at the appropriate time

CharlesIC (approved):

thomasleese (approved):
Code changes look good. When would this need to be deployed?

PR #1417, opened 2021-03-12
Add a rake task to change a LocalService LGSL code

gclssvglx:
What

Add a rake task to change the LGSL code on a LocalService and all of its Editions.

Why

We want to be able to change an existing LocalService's LGSL code.

[Trello](https://trello.com/c/MRGj9keG/210-local-lookup-lateral-mass-test-apply-the-proper-assigned-service-code-1827-agreed-with-local-government-authority)

### DIFF
--- a/lib/tasks/local_service.rake
+++ b/lib/tasks/local_service.rake
@@ -1,0 +1,8 @@
+namespace :local_service do
+  desc "Update the temporary LGSL code of 100001"
+  task change_lgsl_code_100001: :environment do
+    draft = LocalTransactionEdition.find_by!(lgsl_code: 100_001, state: :draft)
+    LocalService.find_by!(lgsl_code: 100_001).update!(lgsl_code: 1827)
+    draft.update!(lgsl_code: 1827)
+  end
+end
--- a/test/lib/tasks/local_service_test.rb
+++ b/test/lib/tasks/local_service_test.rb
@@ -1,0 +1,12 @@
+require "test_helper"
+
+class LocalServiceTaskTest < ActiveSupport::TestCase
+  test "running the local_service:change_lgsl_code_10001 rake task changes the LGSL code" do
+    service = FactoryBot.create(:local_service, lgsl_code: 100_001)
+    edition = FactoryBot.create(:local_transaction_edition, lgsl_code: 100_001, lgil_code: 1)
+
+    Rake::Task["local_service:change_lgsl_code_100001"].execute
+    assert_equal 1827, service.reload.lgsl_code
+    assert_equal 1827, edition.reload.lgsl_code
+  end
+end


lib/tasks/local_service.rake
```diff
@@ -0,0 +1,6 @@
+namespace :local_service do
+  desc "Changes an existing local_service's LGSL code"
+  task :change_lgsl_code, %w[from_lgsl to_lgsl] => :environment do |_, args|
+    ChangeLgslCodeService.call(args.from_lgsl, args.to_lgsl)
+  end
```
kevindew, line 5:
```suggestion
  desc "Update the temporary LGSL code of 100001"
  task change_lgsl_code_10001: :environment do
    draft = LocalTransactionEdition.find_by!(lgsl_code: 100001, state: :draft)
    LocalService.find_by!(lgsl_code: 100001).update!(lgsl_code: 1827)
    draft.update!(lgsl_code: 1827)
  end
```

I think we can rather drastically simplify this and tailor this just for this specific scenario, this should probably remove the need to be so defensive in the code and make this a quicker thing to get out, run and then removed.

app/services/change_lgsl_code_service.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def self.call(from_lgsl: nil, to_lgsl: nil)
+    raise InvalidLGSLCodeError unless from_lgsl.present? && from_lgsl.is_a?(Integer)
+
+    raise InvalidLGSLCodeError unless to_lgsl.present? && to_lgsl.is_a?(Integer)
+
+    raise LocalServiceDoesNotExistError if LocalService.where(lgsl_code: from_lgsl).empty?
+
+    raise LocalServiceAlreadyExistsError if LocalService.where(lgsl_code: to_lgsl).any?
```
kevindew, line 27: This is impressively thorough but I think it's a bit too much for a temporary task, I think you can avoid the need for much of these checks through Mongoid's API and validations.

kevindew (commented):
Thanks for what you've done on this so far. Looks like you've really dug into this.

I've put a rather drastically simpler counter suggestion in to reflect that we only need this for a one off. Hope it helps.

test/lib/tasks/local_service_test.rb
```diff
@@ -0,0 +1,10 @@
+require "test_helper"
+
+class LocalServiceTaskTest < ActiveSupport::TestCase
+  context ".call" do
+    should "send the from and to LGSL codes to the change LGSL code service" do
+      ChangeLgslCodeService.expects(:call).with(1, 2)
+      ChangeLgslCodeService.call(1, 2)
```
kevindew, line 7: Oh by the way this test seems a bit bizarre I'm wondering if you didn't finish what you were doing? It doesn't seem to test a task at all and just expects a function call that you then make on the subsequent line.

gclssvglx:
@kevindew Thanks for your comments and suggestions 👍  All fixed up

kevindew (approved):

PR #1418, opened 2021-03-12
Update LocalService LGSL code

gclssvglx:
## What

Add a LocalService with an LGSL code 1827.

## Why

The service was created using a dummy LGSL code (100001).  We now have
a real LGSL code for this service (1827) and would like to use it
instead of the dummy code.

For safety and to ensure there is no interruption to the service we need
to add the 1827 service - leaving the 100001 service in-place - until
the switch is complete. At that point we can remove the 100001 service.

[Trello](https://trello.com/c/MRGj9keG/210-local-lookup-lateral-mass-test-apply-the-proper-assigned-service-code-1827-agreed-with-local-government-authority)

### DIFF
--- a/data/local_services.csv
+++ b/data/local_services.csv
@@ -126,4 +126,5 @@ LGSL,Description,Providing Tier
 1743,Find out about the Care Act 2014,county/unitary
 1788,"Find out about support for children who have difficulties with speech, language or communication",county/unitary
 1826,"Find out about the test and trace support payment scheme",district/unitary
+1827,"Find a COVID-19 rapid lateral flow test site",all
 100001,"Find a COVID-19 rapid lateral flow test site",all


kevindew:
This looks good, does it need to go in at the same time as https://github.com/alphagov/publisher/pull/1417 or can it be merged whenever?

gclssvglx:
> This looks good, does it need to go in at the same time as #1417 or can it be merged whenever?

Umm! Unless someone [runs this rake task](https://github.com/alphagov/publisher/blob/b908a0a419c03248f9cd5e123e7d113c93717e74/lib/tasks/local_transactions.rake#L3) once this PR is merged - but before the entire switch of the LGSL code is complete - nothing will happen. If they were to run it, the 100001 service would be removed and the 1827 service created - breaking the service.  Therefore, a safer approach would be to add the 1827 service - and not remove the 100001 service here.  Then removed the 100001 service in a separate PR that is merged once the switch is complete.

kevindew:
> > This looks good, does it need to go in at the same time as #1417 or can it be merged whenever?
> 
> Umm! Unless someone [runs this rake task](https://github.com/alphagov/publisher/blob/b908a0a419c03248f9cd5e123e7d113c93717e74/lib/tasks/local_transactions.rake#L3) once this PR is merged - but before the entire switch of the LGSL code is complete - nothing will happen. If they were to run it, the 100001 service would be removed and the 1827 service created - breaking the service. Therefore, a safer approach would be to add the 1827 service - and not remove the 100001 service here. Then removed the 100001 service in a separate PR that is merged once the switch is complete.
> 
> I'll update this PR and create a separate one to remove 100001.

Okey doke, it sounds good to have both and then delete 100001 afterwards. 

It sounds like doing that and using the service sync can simplify the tasks needed for: https://github.com/alphagov/publisher/pull/1417/files to not need to change the service if it can be done by the rake task to deal with this file.

kevindew (approved):

PR #1416, opened 2021-03-11
Remove \u2028 (line separator) characters on save

brucebolt:
We have found some editions contain a `\u2028` (line separator) character when the content has been copy-and-pasted from certain applications. This gets rendered in different ways in different browsers (e.g. an extra space, or sometimes a square box containing "L SEP").

This removes those characters when an edition is saved, so they never get published.

Trello card: https://trello.com/c/YQFCDuP8

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #1427, opened 2021-03-26
Modify publisher L-sep validation

callumknights:
Quick addition to the validation for publisher documents that means
documents without 'parts' that just have a 'body' are also
caught by the validation.

Related: https://trello.com/c/TGUvoJsP/2432-3-remove-l-sep-characters-from-already-published-publisher-content
Original PR: https://github.com/alphagov/publisher/pull/1416

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -463,11 +463,13 @@ class Edition
   end
 
   def remove_line_separator_character
-    return unless respond_to?(:parts)
-
     character = "\u2028"
-    parts.each do |part|
-      part.body = part.body.to_s.gsub(character, "")
+    if respond_to?(:parts)
+      parts.each do |part|
+        part.body = part.body.to_s.gsub(character, "")
+      end
+    elsif respond_to?(:body)
+      self.body = body.to_s.gsub(character, "") unless body.nil?
     end
   end
 
--- a/test/models/edition_test.rb
+++ b/test/models/edition_test.rb
@@ -1221,7 +1221,7 @@ class EditionTest < ActiveSupport::TestCase
   end
 
   context "where the body contains a line separator character" do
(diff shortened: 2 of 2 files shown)


app/models/edition.rb
```diff
@@ -463,11 +463,15 @@ def latest_link_check_report
   end
 
   def remove_line_separator_character
-    return unless respond_to?(:parts)
+    return unless respond_to?(:parts) || respond_to?(:body)
```
brucebolt, line 466: You can probably remove this line, as the `if` and `elsif` with have the same effect of returning `nil` if neither method are available.

brucebolt (commented):
Looks good to me 👍 

Happy to approve, but I think there's a test that needs uncommenting, and I've left a very minor comment on removing a line that I think is redundant (so will give a marginal performance improvement).

test/models/edition_test.rb
```diff
@@ -1221,12 +1221,19 @@ def draft_second_edition_from(published_edition)
...
   context "where the body contains a line separator character" do
-    should "remove character on save" do
+    should "remove character on save with 2 part edition" do
       edition = FactoryBot.create(:guide_edition_with_two_parts)
       edition.parts.first.body = "Some text \u2028with a line separator character"
       edition.save!
 
       assert_no_match(/\u2028/, edition.parts.first.body)
     end
+    # should "remove character on save for edition with no parts" do
```
brucebolt, line 1231: This test will need uncommenting.

callumknights: Sorted 👍

app/models/edition.rb
callumknights: Done 👍

brucebolt (approved):

PR #1456, opened 2021-05-27
Devolved administration availability modelling

Rosa-Fox:
[Trello](https://trello.com/c/Kgr3I4Tz/401-store-devolved-administration-local-transaction-data-in-model-and-govuk-content-schema)

# What?

This work is part of a task to enable publishers to specify devolved administration (DA) service availability for local transactions. This PR models the data and was created as a result of a [spike implementation in Publisher](https://github.com/alphagov/publisher/pull/1451). 

[A related content schema has also been added](https://github.com/alphagov/govuk-content-schemas/pull/1060). This PR should NOT be merged until after the content-schema PR has been merged.

# Modelling

Makes use of Mongoid’s `embeds_one`. This allows data to be stored within a hash. It can then be queried from the parent object.

```
l = LocalTransactionEdition.new
l.scotland_availability

=> #<DevolvedAdministrationAvailability _id: 60a77488a16d490001c09865, type: "devolved_administration", alternative_url: "https://www.scotland.gov/service">
```

# Presenting to the Publishing API

[Related content-schema example](https://github.com/alphagov/govuk-content-schemas/compare/national-applicability-local-transactions?expand=1).

# Naming 

Availability > applicability
DevolvedAdminstration > DevolvedNation
Unavailable > not_available
Devolved_administration_service > available_url (as a type option)

Paired with @kevindew.

### DIFF
--- a/app/models/devolved_administration_availability.rb
+++ b/app/models/devolved_administration_availability.rb
@@ -1,0 +1,7 @@
+class DevolvedAdministrationAvailability
+  include Mongoid::Document
+
+  embedded_in :local_transaction_edition
+  field :type, type: String, default: "local_authority_service"
+  field :alternative_url, type: String
+end
--- a/app/models/devolved_administration_availability.rb
+++ b/app/models/devolved_administration_availability.rb
@@ -4,4 +4,26 @@ class DevolvedAdministrationAvailability
   embedded_in :local_transaction_edition
   field :type, type: String, default: "local_authority_service"
   field :alternative_url, type: String
+
+  validates :type, inclusion: { in: %w[local_authority_service devolved_administration_service unavailable] }
+  validates :alternative_url, presence: true, if: -> { devolved_administration_service? }
+  validate :alternative_url_format
+
+  def alternative_url_format
+    return if alternative_url.blank?
+
+    uri = begin
+      URI.parse(alternative_url)
+    rescue URI::Error
+      nil
+    end
+
+    unless uri.is_a?(URI::HTTPS)
+      errors.add(:alternative_url, "must be a full URL, starting with https://")
+    end
+  end
+
+  def devolved_administration_service?
+    type == "devolved_administration_service"
+  end
 end
--- a/app/presenters/formats/local_transaction_presenter.rb
+++ b/app/presenters/formats/local_transaction_presenter.rb
@@ -28,6 +28,7 @@ module Formats
       {}.merge(introduction)
         .merge(more_information)
         .merge(need_to_know)
+        .merge(all_devolved_administration_availabilities)
     end
 
     def introduction
@@ -67,6 +68,23 @@ module Formats
           },
         ],
       }
+    end
+
+    def all_devolved_administration_availabilities
+      {
+        scotland_availability: devolved_administration_availability(edition.scotland_availability),
+        wales_availability: devolved_administration_availability(edition.wales_availability),
+        northern_ireland_availability: devolved_administration_availability(edition.northern_ireland_availability),
+      }.compact
+    end
+
+    def devolved_administration_availability(availability)
+      case availability.type
+      when "devolved_administration_service"
+        { type: "devolved_administration_service", alternative_url: availability.alternative_url }
+      when "unavailable"
+        { type: "unavailable" }
+      end
     end
 
     def service_tiers
--- a/test/models/devolved_administration_availability_test.rb
+++ b/test/models/devolved_administration_availability_test.rb
@@ -1,0 +1,65 @@
+require "test_helper"
+
+class DevolvedAdministrationAvailabilityTest < ActiveSupport::TestCase
+  test "should not be valid with an unallowed type value" do
+    devolved_administration_availability = FactoryBot.build(
+      :devolved_administration_availability,
+      type: "invalid_type",
+    )
+    assert_not devolved_administration_availability.valid?
+  end
+
+  test "should be valid with an allowed type value" do
+    devolved_administration_availability = FactoryBot.build(
+      :devolved_administration_availability,
+      type: "unavailable",
+    )
+    assert devolved_administration_availability.valid?
+  end
+
+  test "should not be valid if alternative_url is not present when devolved_administration_service is selected" do
+    devolved_administration_availability = FactoryBot.build(
+      :devolved_administration_availability,
+      type: "devolved_administration_service",
+      alternative_url: "",
+    )
+    assert_not devolved_administration_availability.valid?
+  end
+
+  test "should be valid if alternative_url is present when devolved_administration_service is selected" do
+    devolved_administration_availability = FactoryBot.build(
+      :devolved_administration_availability,
+      type: "devolved_administration_service",
+      alternative_url: "https://www.scot.gov/service",
+    )
+    assert devolved_administration_availability.valid?
+  end
+
+  test "should not be valid if alternative_url does not begin with https" do
+    devolved_administration_availability = FactoryBot.build(
+      :devolved_administration_availability,
+      type: "devolved_administration_service",
+      alternative_url: "abc",
+    )
+    assert_not devolved_administration_availability.valid?
+  end
+
+  should "should not be valid if alternative_url is not a valid URI" do
+    devolved_administration_availability = FactoryBot.build(
+      :devolved_administration_availability,
+      type: "devolved_administration_service",
+      alternative_url: " ",
+    )
+
+    assert_not devolved_administration_availability.valid?
+  end
+
+  test "should be valid if alternative_url is a valid URI starting with https" do
+    devolved_administration_availability = FactoryBot.build(
+      :devolved_administration_availability,
+      type: "devolved_administration_service",
+      alternative_url: "https://www.scot.gov/service",
+    )
+    assert devolved_administration_availability.valid?
+  end
+end
--- a/test/support/factories.rb
+++ b/test/support/factories.rb
@@ -229,6 +229,8 @@ FactoryBot.define do
     end
   end
 
+  factory :devolved_administration_availability
+
   factory :local_transaction_edition, parent: :edition, class: "LocalTransactionEdition" do
     sequence(:lgsl_code) { |nlgsl| nlgsl }
     introduction { "Test introduction" }
--- a/test/unit/presenters/formats/local_transaction_presenter_test.rb
+++ b/test/unit/presenters/formats/local_transaction_presenter_test.rb
@@ -138,5 +138,39 @@ class LocalTransactionPresenterTest < ActiveSupport::TestCase
       ]
       assert_equal expected, result[:routes]
     end
+
+    context "devolved administration availability" do
+      should "not present any devolved administration availability values by default" do
+        edition.save!(validate: false)
+        assert_not result[:details].key?(:wales_availability)
+        assert_not result[:details].key?(:scotland_availability)
+        assert_not result[:details].key?(:northern_ireland_availability)
+      end
+
+      should "not present the data if local_authority_service was selected" do
+        edition.scotland_availability = { type: "local_authority_service", alternative_url: "" }
+        edition.save!(validate: false)
+
+        assert_not result[:details].key?(:scotland_availability)
+      end
+
+      should "present the type data if unavailable was selected" do
+        edition.wales_availability.type = "unavailable"
+
+        edition.save!(validate: false)
+
+        expected = { type: "unavailable" }
+        assert_equal expected, result[:details][:wales_availability]
+      end
+
+      should "present the type and url data if devolved_administration_service was selected" do
+        edition.northern_ireland_availability = { type: "devolved_administration_service", alternative_url: "https://www.ni.gov/service" }
+
+        edition.save!(validate: false)
+
+        expected = { type: "devolved_administration_service", alternative_url: "https://www.ni.gov/service" }
+        assert_equal expected, result[:details][:northern_ireland_availability]
+      end
+    end
   end
 end
--- a/app/models/local_transaction_edition.rb
+++ b/app/models/local_transaction_edition.rb
@@ -9,6 +9,10 @@ class LocalTransactionEdition < Edition
   field :more_information, type: String
   field :need_to_know, type: String
 
+  embeds_one :scotland_availability, class_name: DevolvedAdministrationAvailability, autobuild: true
+  embeds_one :wales_availability, class_name: DevolvedAdministrationAvailability, autobuild: true
+  embeds_one :northern_ireland_availability, class_name: DevolvedAdministrationAvailability, autobuild: true
+
   GOVSPEAK_FIELDS = %i[introduction more_information need_to_know].freeze
 
   validate :valid_lgsl_code


app/models/devolved_administration_availability.rb
```diff
@@ -4,4 +4,27 @@ class DevolvedAdministrationAvailability
...
+  validate :alternative_url_format
+
+  def alternative_url_presence
+    if devolved_administration_service_selected? && alternative_url.blank?
+      errors.add(:alternative_url, "must be provided")
+    end
+  end
+
+  def alternative_url_format
+    if devolved_administration_service_selected? && alternative_url.present?
```
kevindew, line 19: You might find it's a bit cleaner to handle this with a guard clause e.g. `return if alternative_url.blank? || !develoved_administration_service_selected?`

app/models/devolved_administration_availability.rb
```diff
@@ -4,4 +4,27 @@ class DevolvedAdministrationAvailability
...
+      uri = URI.parse(alternative_url)
```
kevindew, line 20: This could raise an error (for example `URI.parse("bad uri")`) so we'll want to have a rescue to catch that.

app/models/devolved_administration_availability.rb
```diff
@@ -4,4 +4,27 @@ class DevolvedAdministrationAvailability
...
+      unless uri.is_a?(URI::HTTP) || uri.is_a?(URI::HTTPS)
```
kevindew, line 21: This looks a neat way to check it's the right type of URL 👍  - I think we'll only want users to enter a HTTPS URL

app/models/devolved_administration_availability.rb
```diff
@@ -4,4 +4,27 @@ class DevolvedAdministrationAvailability
...
+        errors.add(:alternative_url, "invalid format")
```
kevindew, line 22: It'd be good if this was a bit more helpful for a user about what is wrong. Perhaps: "must be a full URL, starting with https://"?

kevindew, line 22: It'd be good if this was a bit more helpful for a user about what is wrong. Perhaps: "must be a full URL, starting with https://"?

app/models/devolved_administration_availability.rb
```diff
@@ -4,4 +4,27 @@ class DevolvedAdministrationAvailability
...
+      end
+    end
+  end
+
+  def devolved_administration_service_selected?
```
kevindew, line 27:
```suggestion
  def devolved_administration_service?
```

Minor suggestion - it'll make this behave more similarly to ActiveRecord enums (which it's behaving like): https://api.rubyonrails.org/v5.2.5/classes/ActiveRecord/Enum.html It's a bit of a shame Mongoid doesn't support enums but doesn't seem worth installing an extra gem for.

test/models/devolved_administration_availability_test.rb
```diff
@@ -9,4 +9,55 @@ class DevolvedAdministrationAvailabilityTest < ActiveSupport::TestCase
...
+    devolved_administration_availability = FactoryBot.build(
+      :devolved_administration_availability,
+      type: "invalid_type",
+    )
+    assert_not devolved_administration_availability.valid?
+  end
+
+  test "should be valid with an allowed type value" do
+    devolved_administration_availability = FactoryBot.build(
+      :devolved_administration_availability,
```
kevindew, line 23: It'd be good if this test had a type argument provided as otherwise it looks like it's testing the factory.

app/models/devolved_administration_availability.rb
```diff
@@ -4,4 +4,27 @@ class DevolvedAdministrationAvailability
   embedded_in :local_transaction_edition
   field :type, type: String, default: "local_authority_service"
   field :alternative_url, type: String
+
+  validates :type, inclusion: { in: %w[local_authority_service devolved_administration_service unavailable], allow_blank: true }
+  validate :alternative_url_presence
+  validate :alternative_url_format
```
kevindew, line 10: With alternative_url_format we should have a think about what should happen to an alternative_url when someone changes the type. 

For example in the current situation I could set an alternative_url to be an erroneous URL but by changing the type we bypass the validation and have it be persisted. Demo:

```
irb(main):002:0> availability = DevolvedAdministrationAvailability.new(type: "devolved_administration_service", alternative_url: "/wrong-url")
=> #<DevolvedAdministrationAvailability _id: 60b7af5a39fad80caa4ef0a5, type: "devolved_administration_service", alternative_url: "/wrong-url">
irb(main):003:0> availability.valid?
=> false
irb(main):004:0> availability.type = "unavailable"
=> "unavailable"
irb(main):005:0> availability.valid?
=> true
irb(main):006:0> availability.alternative_url
=> "/wrong-url"
```

I'd imagine the simplest approach would be to always validate the alternative_url if it is present.

test/unit/presenters/formats/local_transaction_presenter_test.rb
```diff
@@ -23,6 +23,15 @@ def edition
...
       need_to_know: "for your eyes only",
+      scotland_availability: FactoryBot.build(
+        :devolved_administration_availability,
+      ),
+      wales_availability: FactoryBot.build(
+        :devolved_administration_availability,
+      ),
+      northern_ireland_availability: FactoryBot.build(
+        :devolved_administration_availability,
+      ),
```
kevindew, line 34:
```suggestion
```

I don't think you need any of these lines because of the autobuild attribute.

app/presenters/formats/local_transaction_presenter.rb
```diff
@@ -69,6 +70,22 @@ def need_to_know
...
+        northern_ireland_availability: devolved_administration_availability(edition.northern_ireland_availability),
+      }.compact
+    end
+
+    def devolved_administration_availability(availability)
+      if availability.type == "devolved_administration_service"
+        { type: "devolved_administration_service", alternative_url: availability.alternative_url }
+      elsif availability.type == "unavailable"
+        { type: "unavailable" }
+      end
```
kevindew, line 86: I reckon the linter is going to suggest you change these to:

``
case availability.type
when "devolved_administration_service"
```

once https://github.com/alphagov/publisher/pull/1457 is merged.

File: test/unit/presenters/formats/local_transaction_presenter_test.rb [hunk 10]
```diff
@@ -138,5 +147,38 @@ def result
       ]
       assert_equal expected, result[:routes]
     end
+
+    context "[:scotland_availability]" do
```
kevindew, line 151: These tests look good 👍  My only gripe is that we're only covering one administration. I'm just having a ponder what could be done to cover them all concisely.

We could change the context to be `context "devolved administration availability"` and then within that we could have a test that asserts that by default none of the nations are included (note `assert result[:details].has_key?(:wales_availability)` might be a slightly cleaner way to assert).

Then I wonder if for the next tests we could test one with Northern Ireland and keep the other as Scotland to have a little country variety.


I feel like this should be covering all of the administrations in a way

test/unit/presenters/formats/local_transaction_presenter_test.rb
```diff
@@ -138,5 +147,38 @@ def result
...
+
+        expected = nil
+        assert_equal expected, result[:details][:scotland_availability]
+      end
+
+      should "present the type data if unavailable was selected" do
+        edition.scotland_availability = FactoryBot.build(
+          :devolved_administration_availability,
+          type: "unavailable",
+        )
```
kevindew, line 164: As there is already a model we can simplify this a little:

```suggestion
        edition.scotland_availability.type = "unavailable"
```

The next test can also be simplified a little.

kevindew (commented):
This is looking great - the core of this looks good, there's a couple of snags I've caught to consider. However, most of the comments I've made are rather minor stylistic things or pointing out unnecessary code.

test/support/factories.rb
```diff
@@ -229,6 +229,11 @@
     end
   end
 
+  factory :devolved_administration_availability, class: "DevolvedAdministrationAvailability" do
```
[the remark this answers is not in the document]
Rosa-Fox: Thanks! I have updated.

test/support/factories.rb
```diff
@@ -229,6 +229,11 @@
...
+    type { "local_authority_service" }
+    alternative_url { "" }
```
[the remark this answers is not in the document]
Rosa-Fox: Thanks! I have updated.

test/models/devolved_administration_availability_test.rb
```diff
@@ -0,0 +1,12 @@
+require "test_helper"
+
+class DevolvedAdministrationAvailabilityTest < ActiveSupport::TestCase
+  test "should create a new devolved administration availability" do
```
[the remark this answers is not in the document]
Rosa-Fox: Thanks, I removed the test.

app/models/devolved_administration_availability.rb
```diff
@@ -4,4 +4,27 @@ class DevolvedAdministrationAvailability
   embedded_in :local_transaction_edition
   field :type, type: String, default: "local_authority_service"
   field :alternative_url, type: String
+
+  validates :type, inclusion: { in: %w[local_authority_service devolved_administration_service unavailable], allow_blank: true }
```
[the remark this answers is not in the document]
Rosa-Fox: Thanks! I have updated.

app/models/devolved_administration_availability.rb
```diff
@@ -4,4 +4,27 @@ class DevolvedAdministrationAvailability
...
+  validate :alternative_url_presence
```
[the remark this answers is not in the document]
Rosa-Fox: Thanks! I have updated.

app/models/devolved_administration_availability.rb
Rosa-Fox: I updated this but just the `return if alternative_url.blank?` part. Otherwise I think we could encounter the problem you mentioned above of the `alternative_url_format` not being validated if the type changes.

app/models/devolved_administration_availability.rb
Rosa-Fox: Thanks, I updated.

app/models/devolved_administration_availability.rb
Rosa-Fox: Thanks, I updated.

app/models/devolved_administration_availability.rb
Rosa-Fox: Thanks, I updated.

app/models/devolved_administration_availability.rb
Rosa-Fox: Thanks, I updated.

app/models/devolved_administration_availability.rb
Rosa-Fox: Thanks, I updated to always validate the alternative_url if it is present.

test/unit/presenters/formats/local_transaction_presenter_test.rb
Rosa-Fox: Thanks, I updated.

app/presenters/formats/local_transaction_presenter.rb
Rosa-Fox: Thanks, I updated.

test/unit/presenters/formats/local_transaction_presenter_test.rb
Rosa-Fox: ooh yeah! thanks! have updated.

test/unit/presenters/formats/local_transaction_presenter_test.rb
Rosa-Fox: Thanks, I updated.

test/models/devolved_administration_availability_test.rb
Rosa-Fox: Thanks. I updated!

Rosa-Fox:
Thanks very much for your review @kevindew. It is ready for re-review :)

test/support/factories.rb
```diff
@@ -229,6 +229,9 @@
     end
   end
 
+  factory :devolved_administration_availability do
+  end
```
kevindew, line 233:
```suggestion
  factory :devolved_administration_availability
```

I doubt you need the block here.

app/models/devolved_administration_availability.rb
```diff
@@ -4,4 +4,27 @@ class DevolvedAdministrationAvailability
...
+  validates :alternative_url, presence: true, if: -> { devolved_administration_service? }
+  validate :alternative_url_format
+
+  def alternative_url_format
+    return if alternative_url.blank?
+
+    begin
+      uri = URI.parse(alternative_url)
+    rescue URI::Error => e
+      Rails.logger.debug e.message.to_s
```
kevindew, line 18: I don't think you need to log this - I imagine it'd be a bit confusing in the log without any context.

app/models/devolved_administration_availability.rb
```diff
@@ -4,4 +4,27 @@ class DevolvedAdministrationAvailability
...
+      errors.add(:alternative_url, "must be a full URL, starting with https://")
+    end
+
+    unless uri.is_a?(URI::HTTPS)
+      errors.add(:alternative_url, "must be a full URL, starting with https://")
+    end
```
kevindew, line 24: You can probably simplify this to avoid setting the error twice

```suggestion
    uri = begin
            URI.parse(alternative_uri)
          rescue URI::Error
            nil
          end
         
    unless uri.is_a?(URI::HTTPS)
      errors.add(:alternative_url, "must be a full URL, starting with https://")
    end         
```

test/models/devolved_administration_availability_test.rb
```diff
@@ -0,0 +1,65 @@
...
+  test "should be valid if alternative_url is present when devolved_administration_service is selected" do
+    devolved_administration_availability = FactoryBot.build(
+      :devolved_administration_availability,
+      type: "devolved_administration_service",
+      alternative_url: "https://www.scot.gov/service",
+    )
+    assert devolved_administration_availability.valid?
+  end
+
+  test "should not be valid if alternative_url is not a valid URI" do
```
kevindew, line 38: This statement doesn't quite match the test. "abc" is a valid URI but fails due to not being HTTPS. You perhaps want two invalid tests, one with an invalid URI (just a space is enough for that) and another for a valid URI that isn't HTTPS.

kevindew (commented):
Nice one Rosa, this looks almost there, just a couple of minor things to resolve.

test/support/factories.rb
Rosa-Fox: 👍

app/models/devolved_administration_availability.rb
Rosa-Fox: 👍

app/models/devolved_administration_availability.rb
Rosa-Fox: 👍

test/models/devolved_administration_availability_test.rb
Rosa-Fox: 👍

Rosa-Fox:
Thanks @kevindew. It is ready for re-review.

kevindew (approved):
Thanks Rosa 👍 

You missed a small bit in https://github.com/alphagov/publisher/pull/1456#discussion_r646504283

Where it puts the assignment of uri outside the begin block:

suggestion:
```
uri = begin
        ...
      rescue 
        nil
      end
```

currently:
```
begin
  uri = ...
rescue
  nil
end
```

I don't think you'll need another review to tweak that though.

kevindew (approved):

PR #1466, opened 2021-06-15
Duplicate DA values for new editions

Rosa-Fox:
[Trello](https://trello.com/c/TsRhfzNb/448-duplicate-devolved-administration-availability-values-when-new-editions-are-created)

Draft PR at the moment... i'll create a proper PR after testing the branch out on integration tomorrow (16/06) but feel free to start reviewing.

Devolved administration availability modelling was added in a [previous PR](https://github.com/alphagov/publisher/pull/1456).

A bug was discovered, in that the embedded fields in `DevolvedAdministrationAvailability` were not being copied when a new edition was created.

The reason for this is the `scotland_availability`, `wales_availability` and `northern_ireland` availability fields were not included in the array returned by the [type_specific_field_keys](https://github.com/alphagov/publisher/blob/main/app/models/edition.rb#L499) method. The test for duplicating editions used a different edition type, which is partly why the bug was missed.

This PR overrides the `build_clone` method to ensure that the devolved administration availability values are copied over when a new edition is created.

### DIFF
--- a/app/models/local_transaction_edition.rb
+++ b/app/models/local_transaction_edition.rb
@@ -39,4 +39,14 @@ class LocalTransactionEdition < Edition
   def whole_body
     introduction
   end
+
+  def build_clone(target_class = nil)
+    new_edition = super
+    if new_edition.is_a?(LocalTransactionEdition)
+      new_edition.scotland_availability = scotland_availability.clone
+      new_edition.wales_availability = wales_availability.clone
+      new_edition.northern_ireland_availability = northern_ireland_availability.clone
+    end
+    new_edition
+  end
 end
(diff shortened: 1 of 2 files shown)


kevindew:
Just a quick comment as I'm running out of time before a meeting. Thanks for working this out 👍 

I think there's some improvements we can make to this, so it's a bit of a cleaner implementation. I don't think the Edition class should know about devolved administration stuff at all, that should be entirely within the LocalTransactionEdition model.

There is some prior art in https://github.com/alphagov/publisher/blob/45182956f5bd16eb0c6ae1b88ed9783d60d5efbe/app/models/simple_smart_answer_edition.rb#L27-L35

I imagine it'd be better still if the copying of fields was able to copy over the embedded fields implicitly as per fields, I'm not sure how much of a pain that is but could probably be investigated by looking into `edition_class.embedded` and having something to remove actions and link_check_reports.

Hope this helps, happy to chat verbally about it.

Rosa-Fox:
Thanks @kevindew i'll look into the suggested approach. I'll message you if any questions.

kevindew (approved):
Looks good to me 👍  Thanks

PR #1473, opened 2021-06-28
Update the user interface to store, validate and retrieve the devolved administration data

gclssvglx:
## What

In a [previous PR](https://github.com/alphagov/publisher/pull/1456/) we have adapted the [`LocalTransactionEdition`](https://github.com/alphagov/publisher/blob/3e083c9998692847cbb7325c0e9234684dc1be35/app/models/local_transaction_edition.rb) and [`DevolvedAdministrationAvailability`](https://github.com/alphagov/publisher/blob/3e083c9998692847cbb7325c0e9234684dc1be35/app/models/devolved_administration_availability.rb) models so that each `LocalTransactionEdition` has a 1:1 relationship with a Scotland, Wales & Northern Ireland `DevolvedAdministrationAvailability` instance.

We now need to update the user interface so that these devolved administration availability entries can be created & maintained.

## Why

We want to provide publishers with the means to specify whether a Local Transaction is only available in particular devolved administrations and to allow them to control the devolved administration availability themselves.

There is an [MVP UI design available](https://trello-attachments.s3.amazonaws.com/5ea1b10bd5e5988b8e0a8690/607d981400f41b805ffe9f5e/7ba499a85ec9eb441cb82bce9a7d0c5d/MVP_implementation.png) that the UI changes should match - although [the actual content has changed from the original design](https://docs.google.com/document/d/1pT0hglrcisw0NH8M3aEOo-LQuBd9x6IjIb7Cx__TqDg/edit) so the final design should also reflect these changes.
 
[Trello](https://trello.com/c/Ip1IP9YU/403-update-publisher-interface-to-store-devolved-administration-data-for-a-local-transaction)

### Before

screencapture-publisher-dev-gov-uk-editions-60cb5ebe39b2d700017cd974-2021-06-29-10_00_32

### After

screencapture-publisher-dev-gov-uk-editions-60cb5ebe39b2d700017cd974-2021-07-02-10_52_54

### After - Published Edition

screencapture-publisher-dev-gov-uk-editions-60cb5ebe39b2d700017cd974-2021-07-02-11_17_23

### After - Creation of New Edition

screencapture-publisher-dev-gov-uk-editions-60dee7c139b2d700014b9141-2021-07-02-11_17_55

### DIFF
--- a/app/assets/stylesheets/forms.scss
+++ b/app/assets/stylesheets/forms.scss
@@ -67,3 +67,13 @@
 .vertical-drag {
   cursor: ns-resize;
 }
+
+.inset-text-field {
+  border-left: 5px solid #bcbcbc;
+  padding: 6px 0 6px 10px;
+  margin-left: 25px;
+
+  .inset-text-field-label {
+    font-weight: normal;
+  }
+}
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -259,12 +259,15 @@ protected
         licence_overview
       ]
     when :local_transaction_edition
-      %i[
-        lgsl_code
-        lgil_code
-        introduction
-        more_information
-        need_to_know
+      [
+        :lgsl_code,
+        :lgil_code,
+        :introduction,
+        :more_information,
+        :need_to_know,
+        { scotland_availability_attributes: %i[type alternative_url] },
+        { wales_availability_attributes: %i[type alternative_url] },
+        { northern_ireland_availability_attributes: %i[type alternative_url] },
       ]
     when :place_edition
       %i[
--- a/app/models/local_transaction_edition.rb
+++ b/app/models/local_transaction_edition.rb
@@ -13,6 +13,10 @@ class LocalTransactionEdition < Edition
   embeds_one :wales_availability, class_name: DevolvedAdministrationAvailability, autobuild: true
   embeds_one :northern_ireland_availability, class_name: DevolvedAdministrationAvailability, autobuild: true
 
+  accepts_nested_attributes_for :scotland_availability
+  accepts_nested_attributes_for :wales_availability
+  accepts_nested_attributes_for :northern_ireland_availability
+
   GOVSPEAK_FIELDS = %i[introduction more_information need_to_know].freeze
 
   validate :valid_lgsl_code
--- a/app/views/local_transactions/_fields.html.erb
+++ b/app/views/local_transactions/_fields.html.erb
@@ -48,6 +48,49 @@
           <%= f.text_area :need_to_know, rows: 4, disabled: @resource.locked_for_edits?, class: "input-md-7 form-control" %>
         </span>
       </div>
+
+      <% administrations = {
+        "Scotland" => :scotland_availability,
+        "Wales" => :wales_availability,
+        "Northern Ireland" => :northern_ireland_availability
+      } %>
+      <% administrations.each do |title, field_name| %>
+        <%= f.fields_for field_name do |f| %>
+          <p class="h4"><%= title %></p>
+          <div class="form-group">
+            <div class="radio">
+              <%= f.label :type_local_authority_service, class: "control-label" do %>
+                <%= f.radio_button :type, 'local_authority_service', disabled: @resource.locked_for_edits? %>
+                Service available from local council
+              <% end %>
+            </div>
+          </div>
+          <div class="form-group">
+            <div class="radio">
+              <%= f.label :type_devolved_administration_service, class: "control-label" do %>
+                <%= f.radio_button :type, 'devolved_administration_service', disabled: @resource.locked_for_edits? %>
+                Service available from devolved administration (or a similar service is available)
+              <% end %>
+            </div>
+          </div>
+          <div class="form-group inset-text-field">
+            <div class="form-label">
+              <%= f.label :alternative_url, "Enter the URL of the devolved administration website page", class: "inset-text-field-label" %>
+            </div>
+            <div class="form-wrapper">
+              <%= f.text_field :alternative_url, disabled: @resource.locked_for_edits?, class: "input-md-7 form-control" %>
+            </div>
+          </div>
+          <div class="form-group">
+            <div class="radio">
+              <%= f.label :type_unavailable, class: "control-label" do %>
+                <%= f.radio_button :type, 'unavailable', disabled: @resource.locked_for_edits? %>
+                Service not available
+              <% end %>
+            </div>
+          </div>
+        <% end %>
+      <% end %>
     </fieldset>
   </div>
 </div>


app/views/local_transactions/_fields.html.erb
```diff
@@ -48,6 +48,43 @@
...
+        "Scotland" => :scotland_availability,
+        "Wales" => :wales_availability,
+        "Northern Ireland" => :northern_ireland_availability
+      } %>
+      <div class="devolved-administrations">
+        <% administrations.each do |title, field_name| %>
+          <%= f.fields_for field_name do |f| %>
+            <div class="form-group">
+              <div class="form-label">
+                <%= f.label :type, title %>
```
kevindew, line 62: This shouldn't be a label as there is id="input" for it to link to it. I think a header would be the appropriate choice for it. You could have a look for any similar precedents, most semantically I imagine this would be a legend inside a fieldset but that might be a whole can of worms without precedent.

app/views/local_transactions/_fields.html.erb
```diff
@@ -48,6 +48,43 @@
...
+              </div>
+              <div class="devolved-administration-fields">
+                <div class="form-wrapper">
+                  <%= f.radio_button :type, 'local_authority_service', disabled: @resource.locked_for_edits? %>
+                  <%= f.label :type, "Service available from local council", value: 'local_authority_service' %>
```
kevindew, line 67: The label looks a little fishy with the use of value, I imagine we might want to adjust the first argument of label.

You can wrap these labels around the element too which seems to be [an existing pattern](https://github.com/alphagov/publisher/blob/1a974da8b8ea342effc55ed5db78a906d59e5f1e/app/views/editions/_major_change_fields.html.erb#L19-L22) and will resolve the need for additional styling to tell it not to be a normal font size. E.g:


```suggestion
                  <%= f.label :type_local_authority_service, class: "control-label" do %>
                    <%= f.radio_button :type, 'local_authority_service', disabled: @resource.locked_for_edits? %>
                    Service available from local council
                  <% end %>
```

kevindew (commented):
Thanks for putting this together @gclssvglx - all looks very promising. Thanks for flexing some CSS muscles 💪.

There's a few things I've flagged where we've drifted away from some precedents a bit. If I understand right we are also caught behind some snagging bugs with Publisher, is that the case?

gclssvglx:
> Thanks for putting this together @gclssvglx - all looks very promising. Thanks for flexing some CSS muscles 💪.
> 
> There's a few things I've flagged where we've drifted away from some precedents a bit. If I understand right we are also caught behind some snagging bugs with Publisher, is that the case?

👍  Very much the case!  There are some very odd things going on.

app/controllers/editions_controller.rb
```diff
@@ -405,6 +408,11 @@ def tagging_update_form_params
   end
 
   def format_failure_message(resource)
+    if resource.instance_of?(LocalTransactionEdition) && resource.invalid?
+      resource.errors[:base] << resource.errors.full_messages
+      resource.errors[:base].flatten!
+    end
```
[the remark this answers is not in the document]
gclssvglx: It would be good to discuss this.  What I found was that model validation errors weren't being displayed at all - the field is highlighted (on the JS disabled version - but not on the JS enabled version) and there aren't messages displayed in the context of the label/field itself.

app/models/local_transaction_edition.rb
```diff
@@ -49,4 +55,12 @@ def build_clone(target_class = nil)
...
     new_edition
   end
+
+private
+
+  def merge_embedded_devolved_administration_availability_errors
+    [scotland_availability, wales_availability, northern_ireland_availability].each do |administration|
+      errors.merge! administration.errors
+    end
+  end
```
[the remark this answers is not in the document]
gclssvglx: > I wondered if something like this might needed for JS errors (perhaps keys like wales_availability_alternative_url => []) but when I tried that it seemed like it's rather broken across the app.

Yeah, I found this too.  Not the only broken thing either - the JS version (Ajax) simply doesn't display any messages except a sort of catch-all `We had some problems saving. Please check the form below.` in a bar along the bottom of the page - nothing seems to be highlighted and I can't see any meaningful error messages.  Would be good to get an approach for this PR as I suspect a proper fix needs to be done as another story.

app/views/layouts/application.html.erb
```diff
@@ -38,7 +38,7 @@
       data-module="auto-track-event"
       data-track-action="alert-<%= alert_class %>"
       data-track-label="<%= strip_tags(flash[k]) %>">
-        <%= flash[k] %>
+        <%= sanitize flash[k] %>
```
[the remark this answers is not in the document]
gclssvglx: Yup, the line breaks.  Which is odd as they were working before I added more errors to the hash.  It's worth noting that I've not changed the way these line breaks are rendered see [this line](https://github.com/alphagov/publisher/blob/1a974da8b8ea342effc55ed5db78a906d59e5f1e/app/controllers/editions_controller.rb#L409), but without the `sanitize` or (shudder) `html_safe` we see "blah blah blah<br />blah blah blah..."

app/models/local_transaction_edition.rb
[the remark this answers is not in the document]
gclssvglx: The way things are structured currently makes a nice `wales_availability: { alternative_url: "can't be blank" }` type hash difficult to achieve.

app/assets/stylesheets/application.scss
```diff
@@ -14,6 +14,7 @@
 @import 'notes';
 @import 'workflow';
 @import 'broken_links_report';
+@import 'devolved_administrations';
```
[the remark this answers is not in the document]
gclssvglx: I'll re-work the CSS.  Let you know when it's done.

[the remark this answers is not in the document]
gclssvglx: I've re-worked this now - let me know what you think.

app/views/local_transactions/_fields.html.erb
gclssvglx: Fixed up - is now a govuk heading

app/views/local_transactions/_fields.html.erb
gclssvglx: Fixed up - except for the text field (URL) and label - these are all now label with form field embedded.

app/assets/stylesheets/forms.scss
```diff
@@ -67,3 +67,14 @@
...
   cursor: ns-resize;
 }
+
+.inset-text-field {
+  border-left: 5px solid #bcbcbc;
+  padding: 6px 0 6px 10px;
+  position: relative;
+  left: 25px;
+
+  label {
```
andysellick, line 77: We generally avoid element based selectors for styling, like `label`, preferring a class e.g. `.inset-text-field__label` but I'm not sure of the conventions here.

app/assets/stylesheets/forms.scss
```diff
@@ -67,3 +67,14 @@
 .vertical-drag {
   cursor: ns-resize;
 }
+
+.inset-text-field {
+  border-left: 5px solid #bcbcbc;
+  padding: 6px 0 6px 10px;
+  position: relative;
+  left: 25px;
```
andysellick, line 75: `left` is usually reserved for absolutely positioned elements, although it does work on others. Could this be achieved using `margin-left` instead? Or possibly even by extending the padding on line 73?

app/views/local_transactions/_fields.html.erb
```diff
@@ -48,6 +48,41 @@
...
       </div>
+
+      <% administrations = {
+        "Scotland" => :scotland_availability,
+        "Wales" => :wales_availability,
+        "Northern Ireland" => :northern_ireland_availability
+      } %>
+      <% administrations.each do |title, field_name| %>
+        <%= f.fields_for field_name do |f| %>
+          <h3 class="govuk-heading-s"><%= f.label :type, title %></h3>
```
andysellick, line 59: `govuk-heading-s` is a class from `govuk-frontend`, but I don't think Publisher uses that. Does this class do anything?

app/views/local_transactions/_fields.html.erb
```diff
@@ -48,6 +48,41 @@
...
+          <div class="form-group">
```
andysellick, line 60: (trying to do this as a separate comment but github won't let me do that on the same line) 

This seems to be the only heading element in this template. I don't know the context - can you confirm that a `h3` here fits into the overall heading structure of the page?

Also just curious, what does it output as the content for this element? It looks like it might be a label element, which doesn't belong in a heading.

andysellick (commented):
Looks good, my comments have been addressed. Happy with the frontend on this (although it looks like Publisher generally needs a lot of work).

app/assets/stylesheets/forms.scss
gclssvglx: Thanks Andy - added class

app/assets/stylesheets/forms.scss
gclssvglx: Thanks Andy - fixed with `margin-left`

app/views/local_transactions/_fields.html.erb
gclssvglx: Thanks Andy - updated to `<p>` with `h4` class.

app/views/local_transactions/_fields.html.erb
gclssvglx: Thanks Andy - fixed up

app/assets/stylesheets/forms.scss
```diff
@@ -67,3 +67,14 @@
 .vertical-drag {
   cursor: ns-resize;
 }
+
+.inset-text-field {
+  border-left: 5px solid #bcbcbc;
+  padding: 6px 0 6px 10px;
+  position: relative;
```
kevindew, line 74: You shouldn't need this line now that you've got a margin-left on this.

app/assets/stylesheets/forms.scss
```diff
@@ -67,3 +67,14 @@
...
+  margin-left: 25px;
+
+  .non-bold-label {
```
kevindew, line 77: I'd suggest naming this to `.inset-text-field-label` as it seems a bit surprising that this class would only work in a nesting situation (and the naming is perhaps too specific about what the style does rather than what the element is)

app/views/local_transactions/_fields.html.erb
```diff
@@ -48,6 +48,41 @@
...
+      <% administrations = {
+        "Scotland" => :scotland_availability,
+        "Wales" => :wales_availability,
+        "Northern Ireland" => :northern_ireland_availability
+      } %>
+      <% administrations.each do |title, field_name| %>
+        <%= f.fields_for field_name do |f| %>
+          <p class="h4"><%= title %></p>
+          <div class="form-group">
+            <div class="form-wrapper radio">
```
kevindew, line 61:
```suggestion
            <div class="radio">
```

I don't think these form-wrapper classes do anything for us here and they seem inconsistently applied compared to how they're applied on other elements (they seem to be use to wrap a control where label is a separate element).

Same for the other `.form-wrapper .radio` selectors.

app/views/local_transactions/_fields.html.erb
```diff
@@ -48,6 +48,41 @@
...
+              <%= f.label :type_devolved_administration_service, class: "control-label" do %>
+                <%= f.radio_button :type, 'devolved_administration_service', disabled: @resource.locked_for_edits? %>
+                Service available from devolved administration (or a similar service is available)
+              <% end %>
+            </div>
+            <div class="form-wrapper inset-text-field">
+              <%= f.label :alternative_url, "Enter the URL of the devolved administration website page", class: "non-bold-label" %>
+              <%= f.text_field :alternative_url, disabled: @resource.locked_for_edits?, class: "input-md-7 form-control" %>
+            </div>
```
kevindew, line 76:
```suggestion
            <div class="form-group inset-text-field">
              <span class="form-label">
                <%= f.label :alternative_url, "Enter the URL of the devolved administration website page", class: "non-bold-label" %>
              </span>
              <span class="form-wrapper">
                <%= f.text_field :alternative_url, disabled: @resource.locked_for_edits?, class: "input-md-7 form-control" %>
              </span>
            </div>
```

Just a suggestion of how to make this field closer in element structure to existing fields. I'm anticipating we'll want this consistency when it comes to fixing error handling.

kevindew (commented):
Looks great, think there's just a couple of small things to fix and then this looks ready to merge.

app/assets/stylesheets/forms.scss
gclssvglx: Good spot - thanks - fixed up.

app/assets/stylesheets/forms.scss
gclssvglx: Ah! Ok, I tend to read nested classes in context. So this would be `.inset-text-field-non-text-field` - in my head. Happy to change it.  Fixed up.

app/views/local_transactions/_fields.html.erb
gclssvglx: No, I don't think `form-wrapper` adds much here. This was an attempt to be consistent with [this instance](https://github.com/alphagov/publisher/blob/3e083c9998692847cbb7325c0e9234684dc1be35/app/views/shared/_common_edition_attributes.html.erb#L21). Have removed `form-wrapper`. But, we may need/want to revisit these classes for consistency when dealing with the error handling.

app/views/local_transactions/_fields.html.erb
gclssvglx: Totally agree - a consistent approach would be the best approach for error handling. Fixed up as per your suggestion. I've also added the missing `<div class="form-group">`'s around each form element - for consistency.

kevindew (approved):
Awesome, well done

PR #1478, opened 2021-07-02
Validation errors not being displayed in the user interface

gclssvglx:
## What

Model validation error messages in user interface are currently not being displayed to users.

Tests should exist that catch any change in the desired behaviour.

## Why

The current user experience is sub-optimal. The user might experience a record that fails to save for example, but no error is shown and no indication is given as to which field generated the error. This gives a somewhat "trial and error" experience for the user and is generally disappointing for them.

Error messaging should be made consistent across the application with all models and formats behaving in the same way. 

This should include: 

- devolved administrations on local transactions, 
- parts on various formats, 
- steps on simple smart answers, and 
- any other quirky fields that may exist

[Trello](https://trello.com/c/I5aCsCPB/474-validation-errors-not-being-displayed-in-publisher)

Co-authored-by: Alex Newton <@GDSNewt>
Co-authored-by: Jonathan Hallam <@JonathanHallam>

## Screenshots - with JavaScript enabled

answer
licence
local-transaction

## Screenshots - without JavaScript enabled

answer
licence
local-transaction

### DIFF
--- a/app/assets/javascripts/modules/ajax_save.js
+++ b/app/assets/javascripts/modules/ajax_save.js
@@ -117,15 +117,14 @@
 
       function showErrors (errors) {
         $.each(errors, function (errorKey, errorMessages) {
-          var errorElement = element.find('#edition_' + errorKey + '_input')
-          var $list = $('<ul class="help-block js-error"></ul>')
+          var errorElement = element.find('#edition_' + errorKey)
+          var parents = errorElement.parents('.form-group')
+          parents.addClass('has-error')
 
-          errorElement.addClass('has-error')
+          var list = parents.find('.error-block')
           for (var j = 0, m = errorMessages.length; j < m; j++) {
-            $list.append('<li>' + errorMessages[j] + '</li>')
+            list.append('<li>' + errorMessages[j] + '</li>')
           }
-
-          errorElement.append($list)
         })
       }
 
@@ -136,6 +135,7 @@
       }
 
       function hideErrors () {
+        element.find('.error-block').empty()
         element.find('.js-error').remove()
         element.find('.has-error').removeClass('has-error')
       }
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -1,0 +1,32 @@
+module FormHelper
+  def form_errors(errors)
+    tag.ul(class: %w[help-block error-block]) do
+      safe_join(errors.map { |e| tag.li(e) })
+    end
+  end
+
+  def form_group(form, field_name, label: nil, help: nil, attributes: {}, &block)
+    raise "No input field given" unless block_given?
+
+    errors = form.object.errors[field_name]
+    attributes[:class] = Array(attributes[:class]) << "form-group"
+    attributes[:class] << "has-error" if errors.any?
+
+    tag.div(attributes) do
+      wrapped_label = tag.div(class: "form-label") { form_label_element(form, field_name, label) }
+      wrapped_field = tag.div(class: "form-wrapper", &block)
+      errors = form_errors(errors)
+      help = tag.div(class: "help-block") { help } if help
+
+      safe_join([wrapped_label, wrapped_field, errors, help])
+    end
+  end
+
+private
+
+  def form_label_element(form, field_name, label)
+    return label if label&.match?(/\A<label/) # already a label element
+
+    form.label(field_name, label)
+  end
+end
--- a/app/models/local_transaction_edition.rb
+++ b/app/models/local_transaction_edition.rb
@@ -17,6 +17,8 @@ class LocalTransactionEdition < Edition
   accepts_nested_attributes_for :wales_availability
   accepts_nested_attributes_for :northern_ireland_availability
 
+  after_validation :merge_embedded_errors
+
   GOVSPEAK_FIELDS = %i[introduction more_information need_to_know].freeze
 
   validate :valid_lgsl_code
@@ -52,5 +54,16 @@ class LocalTransactionEdition < Edition
       new_edition.northern_ireland_availability = northern_ireland_availability.clone
     end
     new_edition
+  end
+
+private
+
+  def merge_embedded_errors
+    %i[scotland_availability wales_availability northern_ireland_availability].each do |availability|
+      nested_errors = public_send(availability).errors
+      nested_errors.each do |key, value|
+        errors.add("#{availability}_attributes_#{key}", value)
+      end
+    end
   end
 end
--- a/app/views/editions/_reviewer_field.html.erb
+++ b/app/views/editions/_reviewer_field.html.erb
@@ -1,12 +1,3 @@
-<% error_class = @resource.errors.has_key?(:reviewer) ? ' error has-error' : '' -%>
-<div id="edition_reviewer_input" class="form-group<%= error_class %>">
-  <label for="edition_reviewer" class="control-label">Reviewer</label>
-  <%= f.select :reviewer, User.enabled.order_by([[:name, :asc]]).collect{ |p| [p.name, p.name] }, { :include_blank => true }, { :class => "form-control input-md-3 #{error_class}", :disabled => @resource.locked_for_edits?, "data-module" => 'assignee-select'} %>
-  <% if @resource.errors.has_key?(:reviewer) -%>
-    <ul class="help-block js-error">
-    <% @resource.errors[:reviewer].each do |error| -%>
-      <li><%= error %></li>
-    <% end -%>
-    </ul>
-  <% end -%>
-</div>
+<%= form_group(f, :reviewer) do %>
+  <%= f.select :reviewer, User.enabled.order_by([[:name, :asc]]).collect{ |p| [p.name, p.name] }, { :include_blank => true }, { :class => "form-control input-md-3", :disabled => @resource.locked_for_edits?, "data-module" => 'assignee-select'} %>
+<% end %>
--- a/app/views/licences/_fields.html.erb
+++ b/app/views/licences/_fields.html.erb
@@ -3,14 +3,9 @@
     <fieldset class="inputs">
       <%= render :partial => 'shared/common_edition_attributes', :locals => {:f => f} %>
 
-      <div class="form-group">
-        <span class="form-label">
-          <%= f.label :licence_identifier %>
-        </span>
-        <span class="form-wrapper">
-          <%= f.text_field :licence_identifier, disabled: @resource.locked_for_edits?, class: "input-md-3 form-control" %>
-        </span>
-      </div>
+      <%= form_group(f, :licence_identifier, label: "Licence identifier") do %>
+        <%= f.text_field :licence_identifier, disabled: @resource.locked_for_edits?, class: "input-md-3 form-control" %>
+      <% end %>
 
       <div class="form-group">
         <span class="form-label">
@@ -22,15 +17,9 @@
         </span>
       </div>
 
-      <div class="form-group">
-        <span class="form-label">
-          <%= f.label :continuation_link, "Link to competent authority" %>
-        </span>
-        <span class="form-wrapper">
-          <%= f.text_field :continuation_link, disabled: @resource.locked_for_edits?, class: "input-md-7 form-control" %>
-          <span class="help-block">Link as deep as possible.</span>
-        </span>
-      </div>
+      <%= form_group(f, :continuation_link, label: "Link to competent authority", help: "Link as deep as possible.") do %>
+        <%= f.text_field :continuation_link, disabled: @resource.locked_for_edits?, class: "input-md-7 form-control" %>
+      <% end %>
 
       <div class="form-group">
         <span class="form-label">
--- a/app/views/local_transactions/_devolved_administrations.html.erb
+++ b/app/views/local_transactions/_devolved_administrations.html.erb
@@ -1,0 +1,41 @@
+<% administrations = {
+  "Scotland" => :scotland_availability,
+  "Wales" => :wales_availability,
+  "Northern Ireland" => :northern_ireland_availability
+} %>
+<% administrations.each do |title, field_name| %>
+  <%= f.fields_for field_name do |f| %>
+    <p class="h4 add-top-margin"><%= title %></p>
+
+    <div class="form-group">
+      <div class="radio">
+        <%= f.label :type_local_authority_service, class: "control-label" do %>
+          <%= f.radio_button :type, 'local_authority_service', disabled: @resource.locked_for_edits? %>
+          Service available from local council
+        <% end %>
+      </div>
+    </div>
+
+    <div class="form-group">
+      <div class="radio">
+        <%= f.label :type_devolved_administration_service, class: "control-label" do %>
+          <%= f.radio_button :type, 'devolved_administration_service', disabled: @resource.locked_for_edits? %>
+          Service available from devolved administration (or a similar service is available)
+        <% end %>
+      </div>
+    </div>
+
+    <%= form_group(f, :alternative_url, label: "Enter the URL of the devolved administration website page", attributes: { class: %w[inset-text-field] }) do %>
+      <%= f.text_field :alternative_url, disabled: @resource.locked_for_edits?, class: "input-md-7 form-control" %>
+    <% end %>
+
+    <div class="form-group">
+      <div class="radio">
+        <%= f.label :type_unavailable, class: "control-label" do %>
+          <%= f.radio_button :type, 'unavailable', disabled: @resource.locked_for_edits? %>
+          Service not available
+        <% end %>
+      </div>
+    </div>
+  <% end %>
+<% end %>
--- a/app/views/simple_smart_answers/_node.html.erb
+++ b/app/views/simple_smart_answers/_node.html.erb
@@ -33,7 +33,7 @@
           <%= o.link_to_remove "<i class=\"glyphicon glyphicon-remove glyphicon-smaller-than-text add-right-margin\"></i> Remove option".html_safe,
             :class => "btn btn-default add-left-margin remove-option" %>
           <% if o.object.errors.has_key?(:next_node) -%>
-            <span class="help-inline"><%= o.object.errors[:next_node].join(', ') %></span>
+            <span class="has-error help-inline"><%= o.object.errors[:next_node].join(', ') %></span>
           <% end -%>
         </div>
       <% end %>
--- a/test/functional/editions_controller_test.rb
+++ b/test/functional/editions_controller_test.rb
@@ -331,7 +331,7 @@ class EditionsControllerTest < ActionController::TestCase
     should "show the edit page again if updating fails" do
       Edition.expects(:find).returns(@guide)
       @guide.stubs(:update).returns(false)
-      @guide.expects(:errors).at_least_once.returns(title: %w[values])
+      @guide.errors.add(:title, "values")
 
       post :update,
            params: {
@@ -344,7 +344,7 @@ class EditionsControllerTest < ActionController::TestCase
     should "show the resource base errors if present" do
       Edition.expects(:find).returns(@guide)
       @guide.stubs(:update).returns(false)
-      @guide.expects(:errors).at_least_once.returns(base: ["Editions scheduled for publishing can't be edited"])
+      @guide.errors.add(:base, "Editions scheduled for publishing can't be edited")
 
       post :update, params: { id: @guide.id, edition: {} }
 
--- a/test/integration/edition_major_change_test.rb
+++ b/test/integration/edition_major_change_test.rb
@@ -45,11 +45,7 @@ class EditionMajorChangeTest < JavascriptIntegrationTest
         should "validate that the change note is present for a major change" do
           visit_edition @second_edition
           check("edition_major_change")
-          save_edition_and_assert_error
-
-          within("#edition_change_note_input .help-block") do
-            assert page.has_content?("can't be blank")
-          end
+          save_edition_and_assert_error("can't be blank")
 
           fill_in "edition_change_note", with: "Something changed"
           save_edition_and_assert_success
--- a/test/integration_test_helper.rb
+++ b/test/integration_test_helper.rb
@@ -203,9 +203,10 @@ class JavascriptIntegrationTest < ActionDispatch::IntegrationTest
     end
   end
 
-  def save_edition_and_assert_error
+  def save_edition_and_assert_error(error = nil)
     save_edition
     assert page.has_content? "We had some problems saving"
+    assert page.has_content? error if error.present?
   end
 
   def save_tags_and_assert_success
--- a/test/unit/helpers/form_helper_test.rb
+++ b/test/unit/helpers/form_helper_test.rb
@@ -1,0 +1,70 @@
+require "test_helper"
+
+class FormHelperTest < ActionView::TestCase
+  context "form_errors" do
+    should "return an unordered list with errors when there are errors present" do
+      expected = '<ul class="help-block error-block"><li>One</li><li>Two</li><li>Three</li></ul>'
+
+      assert_equal expected, form_errors(%w[One Two Three])
+    end
+
+    should "return an empty unordered list when there are no errors present" do
+      expected = '<ul class="help-block error-block"></ul>'
+
+      assert_equal expected, form_errors([])
+    end
+  end
+
+  context "form_group" do
+    setup do
+      object = stub(field_name: "", errors: Hash.new([]))
+      @form = FormBuilder.new("edition", object, self, {})
+    end
+
+    should "return the input wrapped within a form group" do
+      output = form_group(@form, :field_name) { @form.text_field(:field_name) }
+      label = '<div class="form-label"><label for="edition_field_name">Field name</label></div>'
+      wrapped_field = '<div class="form-wrapper"><input type="text" value="" name="edition[field_name]" id="edition_field_name" /></div>'
+      error_block = '<ul class="help-block error-block"></ul>'
+
+      assert_equal %(<div class="form-group">#{label}#{wrapped_field}#{error_block}</div>), output
+    end
+
+    should "include help text if that is provided" do
+      output = form_group(@form, :field_name, help: "Help block text") { @form.text_field(:field_name) }
+      help_block = '<div class="help-block">Help block text</div>'
+
+      assert_match help_block, output
+    end
+
+    should "accept, and not wrap, a label element passed in" do
+      output = form_group(@form, :field_name, label: "Custom label text") { @form.text_field(:field_name) }
+      custom_label = '<div class="form-label"><label for="edition_field_name">Custom label text</label></div>'
+
+      assert_match custom_label, output
+    end
+
+    should "return a form-group html structure with a custom attributes" do
+      output = form_group(@form, :field_name, attributes: { id: :field_id, class: %i[one two] }) { @form.text_field(:field_name) }
+      custom_attributes = '<div id="field_id" class="one two form-group">'
+
+      assert_match custom_attributes, output
+    end
+
+    should "return a form-group html structure with a has-error class when any errors are present" do
+      object = stub(field_name: "field_name", errors: Hash.new(["cannot be blank", "must be unique"]))
+      form = FormBuilder.new("edition", object, self, {})
+      output = form_group(form, :field_name) { form.text_field(:field_name) }
+      form_group_with_errors = '<div class="form-group has-error">'
+
+      assert_match form_group_with_errors, output
+      assert_match "cannot be blank", output
+    end
+
+    should "raise an exception if there is no input field given as a block" do
+      assert_raises(RuntimeError, "No input field given") do
+        form_group(@form, :test_field)
+      end
+    end
+  end
+end
--- a/app/assets/stylesheets/forms.scss
+++ b/app/assets/stylesheets/forms.scss
@@ -77,3 +77,7 @@
     font-weight: normal;
   }
 }
+
+.has-error {
+  color: #a94442;
+}
--- a/app/views/local_transactions/_fields.html.erb
+++ b/app/views/local_transactions/_fields.html.erb
@@ -10,14 +10,9 @@
         </span>
       </div>
 
-      <div class="form-group">
-        <span class="form-label">
-          <%= f.label :lgil_code, "LGIL code" %>
-        </span>
-        <span class="form-wrapper">
-          <%= f.text_field :lgil_code, disabled: @resource.locked_for_edits?, class: "input-md-4 form-control" %>
-        </span>
-      </div>
+      <%= form_group(f, :lgil_code, label: "LGIL code") do %>
+        <%= f.text_field :lgil_code, disabled: @resource.locked_for_edits?, class: "input-md-4 form-control" %>
+      <% end %>
 
       <%= render :partial => 'shared/common_edition_attributes', :locals => {:f => f} %>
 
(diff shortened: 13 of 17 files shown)


app/views/licences/_fields.html.erb
```diff
@@ -3,14 +3,13 @@
...
-          <%= f.text_field :licence_identifier, disabled: @resource.locked_for_edits?, class: "input-md-3 form-control" %>
-        </span>
-      </div>
+      <%= render partial: 'shared/validated_text_field', locals: {
+        f: f,
+        resource: resource,
+        id: "licence_identifier",
+        label: "Licence identifier",
+        input_classes: "input-md-3",
+      } %>
```
kevindew, line 12: I'm not convinced this partial based approach is quite wise, it seems like you'll have to create a lot of partials and think of lots of potential options.

It might be better to do things with some helper methods.

You could have a helper for just adding the error block:

something like below? with perhaps an attribute that can be used as a selector - could be used pretty much anywhere.
```
def form_field_errors(errors)
  tag.ul(class: %w[help-block js-error]) do
    safe_join errors.map { |e| tag.li(e) }
  end
end
```

Could then have a helper function that goes further for the most generic cases - which yields the field rather than needing to create it.

```
def form_group(form, field_name, label: nil, help: nil) 
  tag.div(class: "form_group) do
    label = #...
    field = tag.span(class: "form-wrapper") { yield }
    errors = form_field_errors(form.object.errors[field_name])
    help = #...
    safe_join(label, field, errors, help)
  end
end
```
which, in this instance could be called like this:

```
<%= form_group(f, :licence_identifier) do %>
  <%= f.text_field :licence_identifier, disabled: @resource.locked_for_edits?, class: "input-md-3 form-control" %>
<% end %>
```

Not 100% sure this would work mind.

kevindew (commented):
Thanks @gclssvglx, nice work digging into all this. It looks pretty hairy! 

I've tried to put a few suggestions in that could focus this PR a bit more. There's a lot going on so I found it quite tough to pick-up where to review. I think there's some of the foundational stuff that needs to be sussed out before taking the next step.

Happy to chat about this further.

app/assets/javascripts/modules/ajax_save.js
```diff
@@ -117,15 +117,20 @@
...
+          if (errorKey.endsWith('_availability')) {
+            errorElement = element.find('#edition_' + errorKey + '_attributes_alternative_url')
+          } else {
+            errorElement = element.find('#edition_' + errorKey)
           }
+          var parents = errorElement.parents('.form-group')
+          parents.addClass('has-error')
 
-          errorElement.append($list)
+          var list = parents.find('.error-block')
```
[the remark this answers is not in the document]
gclssvglx: Not all fields - no.  But all validated fields will.  That's the idea behind using partials for field types (e.g. `_validated_text_field.html.erb`) so that they have this same basic structure...

```
<div class="form-group">
  <div class="form-label">...</div>
  <div class="form-wrapper">...</div>
  <div class="help-block">...</div>
  <ul class="error-block">...</ul>
</div>
```

I'm thinking of introducing a `_validated_element.html.erb` partial if possible - that has this structure yet yields to the field specific partials as required.  The only thing that really needs to differ is the actual input element.  

Haven't tried doing this so far as I'm not quite sure how to actually do it yet.

app/assets/javascripts/modules/ajax_save.js
```diff
@@ -157,6 +163,14 @@
           return (css.match(/(^|\s)workflow-message-\S+/g) || []).join(' ')
         })
       }
+
+      function humanize (string) {
+        var humanizedString = string.charAt(0).toUpperCase() + string.substr(1).toLowerCase()
+        humanizedString = humanizedString.replaceAll('_', ' ')
+        humanizedString = humanizedString.replaceAll('ireland', 'Ireland')
+        humanizedString = humanizedString.replaceAll('url', 'URL')
```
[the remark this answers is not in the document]
gclssvglx: Yes, it is a bit hacky!  And it does depend on the specific error.  The problem comes when we get errors like this one coming back from the Ajax save response...

```
{ "northern_ireland_availability": { "alternative_url": [ "can't be blank"] } }
```

Resolving [this comment](https://github.com/alphagov/publisher/pull/1478#discussion_r672408263) may help solve this one too.

app/controllers/editions_controller.rb
```diff
@@ -408,10 +408,7 @@ def tagging_update_form_params
   end
 
   def format_failure_message(resource)
-    resource_base_errors = resource.errors[:base]
-    return resource.errors[:base].join("<br />") if resource_base_errors.present?
-
-    "We had some problems saving. Please check the form below."
+    resource.errors.any? ? resource.errors.full_messages : ["We had some problems saving. Please check the form below."]
```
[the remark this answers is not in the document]
gclssvglx: Ah!  Undoubtedly my fault apologies.  I was having wifi issues while we were discussing this and I thought I heard that you wanted to keep (not loose) the `flash` messages.  I'll refactor these out.  Just to be clear, are you happy to remove all `flash` messages?  They are really only used for non-JavaScript responses anyway and it means we can also remove [this block](https://github.com/alphagov/publisher/blob/5fda3e9e6746d7bac1166efef8e7f369b2edbd0a/app/views/layouts/application.html.erb#L26) too, or is it just the ones injected into the `flash` hash via the `format_failure_message` method?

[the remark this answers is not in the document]
kevindew: No problem, for clarity: keep the flash messages as they are currently live. Where they show either the "We had some problems saving. Please check the form below." or any base errors, but don't show any further errors.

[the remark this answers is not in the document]
gclssvglx: Gotcha!  Thanks 👍

app/views/licences/_fields.html.erb
gclssvglx: Thanks @kevindew - I have refactored into helpers as per your suggestion.

app/views/editions/_reviewer_field.html.erb
```diff
@@ -1,12 +1,18 @@
 <% error_class = @resource.errors.has_key?(:reviewer) ? ' error has-error' : '' -%>
 <div id="edition_reviewer_input" class="form-group<%= error_class %>">
-  <label for="edition_reviewer" class="control-label">Reviewer</label>
-  <%= f.select :reviewer, User.enabled.order_by([[:name, :asc]]).collect{ |p| [p.name, p.name] }, { :include_blank => true }, { :class => "form-control input-md-3 #{error_class}", :disabled => @resource.locked_for_edits?, "data-module" => 'assignee-select'} %>
-  <% if @resource.errors.has_key?(:reviewer) -%>
-    <ul class="help-block js-error">
-    <% @resource.errors[:reviewer].each do |error| -%>
```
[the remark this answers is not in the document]
gclssvglx: 👍

app/assets/javascripts/modules/ajax_save.js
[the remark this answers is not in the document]
gclssvglx: I have remove the need for this function by making the with JS version consistent with the without JS version, i.e. errors are presented as `can't be blank` and not `Title can't be blank`.

app/assets/javascripts/modules/ajax_save.js
```diff
@@ -117,15 +117,20 @@
...
         $.each(errors, function (errorKey, errorMessages) {
-          var errorElement = element.find('#edition_' + errorKey + '_input')
-          var $list = $('<ul class="help-block js-error"></ul>')
+          var errorElement = null
 
-          errorElement.addClass('has-error')
-          for (var j = 0, m = errorMessages.length; j < m; j++) {
-            $list.append('<li>' + errorMessages[j] + '</li>')
+          if (errorKey.endsWith('_availability')) {
+            errorElement = element.find('#edition_' + errorKey + '_attributes_alternative_url')
```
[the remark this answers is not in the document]
gclssvglx: Have removed the need for this fishyness.

app/models/local_transaction_edition.rb
```diff
@@ -53,4 +55,18 @@ def build_clone(target_class = nil)
...
+
+  def merge_embedded_errors
+    [scotland_availability, wales_availability, northern_ireland_availability].each do |administration|
+      instance_name = administration.instance_values["_association"].name
+      next unless errors.delete(instance_name) == ["is invalid"]
+
+      da_errors = administration.errors.map do |key, value|
+        "#{key} #{value}".humanize
+      end
+      errors.add(instance_name, da_errors)
```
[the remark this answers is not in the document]
gclssvglx: Have fixed this up based on your suggestion and removed the `SimpleSmartAnswerEdition` version as I don't think it's needed.

app/models/simple_smart_answer_edition.rb
```diff
@@ -73,4 +75,26 @@ def initial_node
...
+      end
+
+      option_errors = node.options.map do |options|
+        options.errors.map do |key, value|
+          "#{key} #{value}".humanize
+        end
+      end
+
+      errors.add(instance_name, (node_errors + option_errors).flatten)
+    end
```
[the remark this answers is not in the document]
gclssvglx: I don't believe we need to merge in the child model errors for a `SimpleSmartAnswerEdition` so I've removed this code.

app/helpers/validated_field_helper.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    tag.div(class: klasses) do
+      label = form_label(field_name, label, classes: label_classes)
+      field = form_input(&block)
+      help = form_help(help)
+      errors = form_errors(errors[field_name])
+
+      safe_join([label, field, help, errors])
+    end
+  end
```
kevindew, line 42: This stuff looks great 👍 

I've had a play around and have some suggestions to make for it.

Suggested approach: 

```ruby
module FormHelper
  def form_errors(errors)
    tag.ul(class: %w[help-block error-block]) do
      safe_join(errors.map { |e| tag.li(e) }) if errors.present?
    end
  end

  def form_group(form, field_name, label: nil, help: nil, attributes: {}, &block)
    errors = form.object.errors[field_name]
    attributes[:class] = Array(attributes[:class]) << "form-group"
    attributes[:class] << "has-error" if errors.any?

    tag.div(attributes) do
      wrapped_label = tag.div(class: "form-label") { form_label_element(form, field_name, label) }
      wrapped_field = tag.div(class: "form-wrapper", &block)
      errors = form_errors(errors)
      help = tag.div(class: "help-block") { help } if help

      safe_join([wrapped_label, wrapped_field, errors, help])
    end
  end

private

  def form_label_element(form, field_name, label)
    return label if label&.match?(/\A<label/) # already a label element

    form.label(field_name, label)
  end
end
```

Differences and why:

- form_group receives a form object rather than an errors block, we can use the form to create the label and access the errors
- label argument can be a HTML label, it means we can just create the label in the code if we need to rather than needing to pass attributes separately
- attributes argument, this can be used to pass through any HTML attribute tweaks we want to make to form group, saving a need for specific id argument too
- no label_classes argument, not necessary if label can be given classes directly
- no form_label helper, this helper seemed a bit confusing since it didn't return a label but a label wrapped in a div, also seemed coupled to editions
- no form_input or form_help helpers, these seemed pretty trivial and input seemed a little misleadingly named
- this will always generate a label element as it'll fall back to field name without them; this seems closer to how Rails behaves and I think fields that are complex enough to do label differently should be handled separately - this stuff should only be for the more conventional fields
- errors before help, this seemed to look nicer to me when testing, it seemed confusing to have field, help text then error

Applied to some examples this comes out like:

```erb
# simple
<%= form_group(f, :licence_identifier) do %>
  <%= f.text_field :licence_identifier, disabled: @resource.locked_for_edits?, class: "input-md-3 form-control" %>
<% end %>

# with help
<%= form_group(f, :continuation_link, label: "Link to competent authority", help: "Link as deep as possible.") do %>
  <%= f.text_field :continuation_link, disabled: @resource.locked_for_edits?, class: "input-md-7 form-control" %>
<% end %>

# complex
<%= form_group(
  f,
  :alternative_url,
  label: f.label(:alternative_url, "Enter the URL of the devolved administration website page", class: "inset-text-field-label"),
  attributes: { class: "inset-text-field" }
) do %>
  <%= f.text_field :alternative_url, disabled: @resource.locked_for_edits?, class: "input-md-7 form-control" %>
<% end %>
```

app/views/editions/_reviewer_field.html.erb
```diff
@@ -1,12 +1,18 @@
 <% error_class = @resource.errors.has_key?(:reviewer) ? ' error has-error' : '' -%>
```
kevindew, line 1: Do we know why this has both error and has-error? I'm guessing we only need has-error in this new world?

app/views/local_transactions/_devolved_administrations.html.erb
```diff
@@ -0,0 +1,35 @@
+<% administrations = {
```
kevindew, line 1: Good shout on putting this into a partial

app/views/local_transactions/_devolved_administrations.html.erb
```diff
@@ -0,0 +1,35 @@
...
+<% administrations.each do |title, field_name| %>
+  <%= f.fields_for field_name do |f| %>
+    <p class="h4 add-top-margin"><%= title %></p>
+
+    <%= form_group(@resource.errors, field_name, classes: %w[radio]) do %>
+      <%= f.label :type_local_authority_service, class: "control-label" do %>
+        <%= f.radio_button :type, 'local_authority_service', disabled: @resource.locked_for_edits? %>
+        Service available from local council
+      <% end %>
+    <% end %>
```
kevindew, line 15: I don't think it's worth putting these into the form group since they're already pretty custom and there couldn't be an error on them (as the radio has a value that isn't a field). I'd suggest going back to:

```html
<div class="form-group">
  <div class="radio">
    <%= f.label :type_local_authority_service, class: "control-label" do %>
    <%= f.radio_button :type, 'local_authority_service', disabled: @resource.locked_for_edits? %>
          Service available from local council
    <% end %>
  </div>
</div>
```

test/integration/transaction_create_edit_test.rb
```diff
@@ -75,7 +75,7 @@ class TransactionCreateEditTest < JavascriptIntegrationTest
       visit_edition transaction
 
       fill_in "Service analytics profile", with: "UA-INVALID-SPACE-FLIGHT"
-      save_edition_and_assert_error
```
kevindew, line 78: As per suggestion earlier I think it's good to keep the `and_assert_error` suffix, it seems a logical counter to the next assertion of `save_edition_and_assert_success`. It'd be a good idea to also allow an error to be passed in which could be asserted against too (e.g. "can't be blank")

kevindew (commented):
Looks like you've made great progress since we last spoke. I've made a few suggestions on how to improve the underlying approach.

test/integration_test_helper.rb
```diff
@@ -203,11 +203,6 @@ def save_edition_and_assert_success
     end
   end
 
-  def save_edition_and_assert_error
-    save_edition
-    assert page.has_content? "We had some problems saving"
```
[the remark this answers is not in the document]
gclssvglx: Ah - great spot.  Personally, I still think the message is kinda meaningless. As per [your suggestion](https://github.com/alphagov/publisher/pull/1478#discussion_r678577979) I'll update this - keeping method and allowing a string param for the error to be checked.  Will keep the `We had some problems saving` check too - now we know why it works for with- and without-JS.  Thanks

app/views/local_transactions/_devolved_administrations.html.erb
gclssvglx: Pragmatically I totally agree and I'm happy to change these back.  But just for completeness... Technically, there can be an error generated here as the [model validates the type here](https://github.com/alphagov/publisher/blob/3e928a81d339387f91699ba6fba8c4a4d1eb5231/app/models/devolved_administration_availability.rb#L8).  In order to fire this validation error the user needs to 'change' the `type` value on the selected radio via dev tools or similar, then submit the form.  Of course this is not an easy (or obvious) thing to do and I don't think we'd expect a user of an internal application like this to actually do it.

app/models/parted.rb
```diff
@@ -39,6 +39,7 @@ def merge_embedded_parts_errors
 
     if errors.delete(:parts) == ["is invalid"]
       parts_errors = parts.each_with_object({}) do |part, result|
+        part.errors.messages.each { |_k, v| v.delete("is invalid") }
```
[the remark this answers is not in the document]
gclssvglx: We've created [this PR](https://github.com/alphagov/publisher/pull/1485) to handle the `parts` and `variants` messaging.  Thanks

app/views/local_transactions/_devolved_administrations.html.erb
kevindew: Yup, I was thinking (incorrectly) that the form group was looking at a field name of `:type_local_authority_service` which can't have an error (since any errors would be on `:type` as you suggest). I see in the code though that the field name for these groups is `:wales_availability` etc so I think that would cause a strange situation where you could any errors for `errors[:wales_availability]` repeated on every single radio button? That'd not make sense as we'd only want an error specific to that radio button - we'd probably want a spot below the radio buttons to show any errors on `:type` if we thought they'd add value.

This perhaps explains to me why you were deleting "is invalid" as I guess you'd get that under every radio button with this code?

gclssvglx: Yes, I've just tried it - with- and without-JS - not removing `is invalid` and changing the type value on Northern Ireland to be `unknown` via dev tools.  With-JS is ok, but without-JS...

### With-JS
with-js

### Without-JS
without-js

app/views/editions/_reviewer_field.html.erb
gclssvglx: Ideally, I'd like just one and `has-error` seems the best option as it'll be used by both with- and without-JS.  But, there are quite a few places to change, so I suggest a separate PR for this too.

app/helpers/validated_field_helper.rb
gclssvglx: Have update as per your suggestion (with a couple of minor tweaks).  Thanks

app/views/editions/_reviewer_field.html.erb
gclssvglx: Have created a separate card for this story

app/views/local_transactions/_devolved_administrations.html.erb
gclssvglx: Have reverted the radio button inputs as suggested.

test/integration/transaction_create_edit_test.rb
gclssvglx: Updated as per your suggestion.

app/models/local_transaction_edition.rb
```diff
@@ -53,4 +55,17 @@ def build_clone(target_class = nil)
     end
     new_edition
   end
+
+private
+
+  def merge_embedded_errors
+    %i[scotland_availability wales_availability northern_ireland_availability].each do |availability|
+      next unless errors.delete(availability) == ["is invalid"]
```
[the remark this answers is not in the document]
gclssvglx: Have remove this line - cannot see `is invalid` in either with- or without-JS version.

app/views/simple_smart_answers/_node.html.erb
```diff
@@ -33,7 +35,7 @@
           <%= o.link_to_remove "<i class=\"glyphicon glyphicon-remove glyphicon-smaller-than-text add-right-margin\"></i> Remove option".html_safe,
             :class => "btn btn-default add-left-margin remove-option" %>
           <% if o.object.errors.has_key?(:next_node) -%>
-            <span class="help-inline"><%= o.object.errors[:next_node].join(', ') %></span>
+            <span class="has-error help-inline"><%= o.object.errors[:next_node].join(', ') %></span>
```
kevindew, line 38: Does this render properly? it looked a bit funky when I tried.

(as an aside we probably should have had this bit as a separate PR too)

test/integration_test_helper.rb
```diff
@@ -203,9 +203,10 @@ def save_edition_and_assert_success
     end
   end
 
-  def save_edition_and_assert_error
+  def save_edition_and_assert_error(error: nil)
```
kevindew, line 206:
```suggestion
  def save_edition_and_assert_error(error = nil)
```

I think this will read better this way, compare: `save_edition_and_assert_error("can't be blank")` with `save_edition_and_assert_error(error: "can't be blank")`

test/integration/edition_major_change_test.rb
```diff
@@ -45,11 +45,7 @@ class EditionMajorChangeTest < JavascriptIntegrationTest
         should "validate that the change note is present for a major change" do
           visit_edition @second_edition
           check("edition_major_change")
-          save_edition_and_assert_error
-
-          within("#edition_change_note_input .help-block") do
-            assert page.has_content?("can't be blank")
-          end
```
kevindew, line 52: Nice to clean this up 👍

test/unit/helpers/form_helper_test.rb
```diff
@@ -0,0 +1,78 @@
...
+    should "return an empty unordered list when there are no errors present" do
+      expected = "<ul class=\"help-block error-block\"></ul>"
+
+      assert_equal expected, form_errors
+    end
+  end
+
+  context "form_group" do
+    setup do
+      object = stub(field_name: "", errors: Hash.new([]))
```
kevindew, line 20: It'd be worth explaining this cryptic `Hash.new([])` that we came up with in the commit message.

app/helpers/form_helper.rb
```diff
@@ -0,0 +1,33 @@
+module FormHelper
+  def form_errors(errors = [])
+    tag.ul(class: %w[help-block error-block]) do
+      safe_join(errors.map { |e| tag.li(e) }) if errors.present?
+    end
+  end
+
+  def form_group(form, field_name, label: nil, help: nil, attributes: {}, &block)
+    raise StandardError, "Form cannot be nil" if form.nil?
+    raise StandardError, "No input field given" unless block_given?
```
kevindew, line 10:
```suggestion
    raise RuntimeError, "No input field given" unless block_given?
```

RuntimeError is more appropriate here

test/unit/helpers/form_helper_test.rb
```diff
@@ -0,0 +1,78 @@
...
+      @form = FormBuilder.new("edition", object, self, {})
+      @label = '<div class="form-label"><label for="edition_field_name">Field name</label></div>'
+      @wrapped_field = '<div class="form-wrapper"><input type="text" value="" name="edition[field_name]" id="edition_field_name" /></div>'
+      @error_block = '<ul class="help-block error-block"></ul>'
+    end
+
+    should "return a form-group html structure without: errors, help block, custom label or custom attributes" do
```
kevindew, line 27:
```suggestion
    should "return the input wrapped within a form group" do
```

test/unit/helpers/form_helper_test.rb
```diff
@@ -0,0 +1,78 @@
...
+      output = form_group(@form, :field_name) { @form.text_field(:field_name) }
+
+      assert_equal %(<div class="form-group">#{@label}#{@wrapped_field}#{@error_block}</div>), output
+    end
+
+    should "return a form-group html structure with a help block when help text is supplied" do
```
kevindew, line 33:
```suggestion
    should "include help text if that is provided" do
```

test/unit/helpers/form_helper_test.rb
```diff
@@ -0,0 +1,78 @@
...
+      output = form_group(@form, :field_name, help: "Help block text") { @form.text_field(:field_name) }
+      help_block = '<div class="help-block">Help block text</div>'
+
+      assert_equal %(<div class="form-group">#{@label}#{@wrapped_field}#{@error_block}#{help_block}</div>), output
```
kevindew, line 37: I don't think we need to test the whole output again. I imagine `assert_match help_block, output` will suffice.

Same comment applies to following tests

test/unit/helpers/form_helper_test.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    should "return a form-group html structure with a custom label" do
```
kevindew, line 40:
```suggestion
    should "accept, and not wrap, a label element passed in" do
```

test/unit/helpers/form_helper_test.rb
```diff
@@ -0,0 +1,78 @@
...
+      output = form_group(@form, :field_name, attributes: { id: :field_id, class: %i[one two] }) { @form.text_field(:field_name) }
+
+      assert_equal %(<div id="field_id" class="one two form-group">#{@label}#{@wrapped_field}#{@error_block}</div>), output
+    end
+
+    should "return a form-group html structure with a has-error class when any errors are present" do
+      object = stub(field_name: "field_name", errors: Hash.new(["cannot be blank", "must be unique"]))
+      form = FormBuilder.new("edition", object, self, {})
+      label = '<div class="form-label"><div class="field_with_errors"><label for="edition_field_name">Field name</label></div></div>'
+      wrapped_field = '<div class="form-wrapper"><div class="field_with_errors"><input type="text" value="field_name" name="edition[field_name]" id="edition_field_name" /></div></div>'
```
kevindew, line 57: We shouldn't be testing these `field_with_errors` classes as their implicit Rails ones that are applied outside our unit of code. I think the only thing we need to test error wise is that form-group gets the extra class. We could also check the errors are displayed, but I don't think we need to do it at a granular HTML level. We can get away with `assert_match "cannot be blank", output`

kevindew (commented):
This looks just about there, thanks for all your work on it. There's a few coding redundancies to fix and a visual problem with Simple Smart Answers.

app/views/simple_smart_answers/_node.html.erb
```diff
@@ -22,7 +21,10 @@
     <div class="options">
       <%= f.fields_for :options, :wrapper_class => "outcome-wrap option" do |o| %>
         <div class="form-inline form-group<%= o.object.errors.empty? ? '' : ' error' %>">
-          <%= o.text_field :label, placeholder: "Label", class: "option-label form-control" %>
+          <%= form_group(o, :label) do %>
+            <%= o.text_field :label, placeholder: "Label", class: "option-label form-control" %>
+          <% end %>
```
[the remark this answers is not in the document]
gclssvglx: I've resolved this, but... I could not find a way to turn off the label generation without introducing a new param `has_label` on the `form_group` method.

Screenshot 2021-08-09 at 12 22 23

app/views/simple_smart_answers/_node.html.erb
gclssvglx: No, it doesn't. To be fair this isn't something we've introduced - it is broken in `main` and the `has-error` class has no other effect than turning the `can't be blank` message red. I will raise another PR and fix the issue there so as not to block this one any further.

app/views/simple_smart_answers/_node.html.erb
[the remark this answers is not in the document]
kevindew: I'm not sure why this needs to use the form_group helper? Can it just be `<%= o.text_field :label, placeholder: "Label", class: "option-label form-control" %>` as it was before?

app/helpers/form_helper.rb
```diff
@@ -0,0 +1,32 @@
+module FormHelper
+  def form_errors(errors)
+    tag.ul(class: %w[help-block error-block]) do
+      safe_join(errors.map { |e| tag.li(e) })
+    end
+  end
+
+  def form_group(form, field_name, has_label: true, label: nil, help: nil, attributes: {}, &block)
```
kevindew, line 8: Let's drop the has_label field. We should expect that if users have a case that doesn't need a label that they can then render their form without the helper.

app/helpers/form_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+  end
+
+  def form_group(form, field_name, label: nil, help: nil, attributes: {}, &block)
+    raise StandardError, "Form cannot be nil" if form.nil?
+    raise StandardError, "No input field given" unless block_given?
+
+    errors = form.object.errors[field_name]
+    attributes[:class] = Array(attributes[:class]) << "form-group"
+    attributes[:class] << "has-error" if errors && errors.any?
```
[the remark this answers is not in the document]
gclssvglx: Hi @kevindew - I've discovered an issue with this. There are 2 tests (e.g. [this one](https://github.com/alphagov/publisher/blob/3e928a81d339387f91699ba6fba8c4a4d1eb5231/test/functional/editions_controller_test.rb#L351)) that send a `Hash` through. They are not all `ActiveModel::Errors` objects as you would expect.

The original line 14 (above) sort of masked the issue and the tests passed. But I'm not 100% convinced it's the best approach. I've tried using the safe navigation operator `errors = form&.object&.errors[field_name] || []` which works, but needs the `|| []`, we could convert the tests to be instances of `ActiveModel::Errors`... But is there a better way?

Please let me know your preferred approach on this.  Thanks.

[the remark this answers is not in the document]
kevindew: Thanks,  I expect the problem in the example you linked to is this line:
https://github.com/alphagov/publisher/blob/3e928a81d339387f91699ba6fba8c4a4d1eb5231/test/functional/editions_controller_test.rb#L347

Where it mocks fake errors on the model

I'm thinking that rather than writing this code in order to satisfy mocking, we should make mocks more realistic. As an opening gambit I'm wondering if we can replace the stub of errors to be:

```
@guide.errors.add(:base, "Editions scheduled for publishing can't be edited")
```

app/views/simple_smart_answers/_node.html.erb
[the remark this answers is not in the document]
gclssvglx: Reverted to original code.  Thanks

app/helpers/form_helper.rb
gclssvglx: removed `has_label`.  Thanks.

app/helpers/form_helper.rb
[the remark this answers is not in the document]
gclssvglx: Yup, this works great - thanks 👍   Both instances updated.

test/unit/helpers/form_helper_test.rb
```diff
@@ -0,0 +1,78 @@
...
+      assert_match "cannot be blank", output
+    end
+
+    should "raise an exception if the input form is nil" do
+      assert_raises(StandardError, "Form cannot be nil") do
+        form_group(nil, :test_field) do
+          nil
+        end
+      end
+    end
```
kevindew, line 70:
```suggestion
```

I'm a bit confused by this test as we don't have code that checks for this or raises this message.

test/unit/helpers/form_helper_test.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    should "raise an exception if there is no input field given as a block" do
+      assert_raises(StandardError, "No input field given") do
```
kevindew, line 73:
```suggestion
      assert_raises(RuntimeError, "No input field given") do
```

minor tweak

kevindew (commented):
This looks just about done, I've just flagged a few minor things

test/functional/editions_controller_test.rb
```diff
@@ -331,7 +331,7 @@ class EditionsControllerTest < ActionController::TestCase
     should "show the edit page again if updating fails" do
       Edition.expects(:find).returns(@guide)
       @guide.stubs(:update).returns(false)
-      @guide.expects(:errors).at_least_once.returns(title: %w[values])
+      @guide.errors.add(:title, %w[values])
```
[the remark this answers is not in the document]
gclssvglx: Good spot - thanks.  Updated as per your suggestion.

test/unit/helpers/form_helper_test.rb
gclssvglx: Yeah, we removed that in an earlier review.  Test removed.

test/unit/helpers/form_helper_test.rb
gclssvglx: Update as per your suggestion - thanks.

gclssvglx:
Hi @kevindew - I've removed the `Node` `title` field conversion in `SimpleSmartAnswer` to `FormHelper` as it added a label that wasn't in the `main` branch.  Also removed the screenshots for `SimpleSmartAnswer`.  This means the `SimpleSmartAnswer` UI is essentially unchanged - but we intend to fix this particular UI in [this PR](https://github.com/alphagov/publisher/pull/1488).

kevindew (approved):
Great stuff 👍

PR #1486, opened 2021-08-04
Update Part model validation errors

gclssvglx:
## What

Model validation error messages in Publisher are currently not being displayed to users.

This PR re-enables the model validation messages for the GuildeEdition format - Part child records.

## Why

[Trello](https://trello.com/c/SvOnnXeB/530-update-parts-and-variants-validation-errors-in-publisher)

Co-authored-by: Alex Newton <@GDSNewt>

### DIFF
--- a/app/views/shared/_part.html.erb
+++ b/app/views/shared/_part.html.erb
@@ -17,6 +17,7 @@
           <span class="form-wrapper">
             <%= f.text_field :title, disabled: @resource.locked_for_edits?, class: "title form-control" %>
           </span>
+          <%= form_errors(f.object.errors[:title]) %>
         </div>
 
         <%
@@ -36,6 +37,7 @@
               for example, title-of-part (no spaces, apostrophes or acronyms)
             </span>
           </span>
+          <%= form_errors(f.object.errors[:slug]) %>
         </div>
 
         <div class="form-group">


app/models/parted.rb
```diff
@@ -39,6 +39,7 @@ def merge_embedded_parts_errors
 
     if errors.delete(:parts) == ["is invalid"]
       parts_errors = parts.each_with_object({}) do |part, result|
+        part.errors.messages.each { |_k, v| v.delete("is invalid") }
```
kevindew, line 42: I think this is a mistake as it's removing "is invalid" validation from fields.

For example if you put in an invalid slug you'd not expect this error to be removed.

test/models/parted_test.rb
```diff
@@ -12,7 +12,7 @@ class PartedTest < ActiveSupport::TestCase
     assert_not edition.valid?
 
     assert_equal({ title: ["can't be blank"] }, edition.errors[:parts][0]["54c10d4d759b743528000010:1"])
-    assert_equal({ slug: ["can't be blank", "is invalid"] }, edition.errors[:parts][0]["54c10d4d759b743528000011:2"])
+    assert_equal({ slug: ["can't be blank"] }, edition.errors[:parts][0]["54c10d4d759b743528000011:2"])
```
kevindew, line 15: It seems strange that this error is gone (to support first comment)

app/views/shared/_part.html.erb
```diff
@@ -17,6 +17,13 @@
           <span class="form-wrapper">
             <%= f.text_field :title, disabled: @resource.locked_for_edits?, class: "title form-control" %>
           </span>
+          <% if f.object.errors[:title].any? %>
+            <ul class="help-block error-block">
+              <% f.object.errors[:title].each do |error| %>
+                <li><%= error %></li>
+              <% end %>
+            </ul>
+          <% end %>
```
kevindew, line 26: Should this be using the new helper?

kevindew: Also will the JS be able to clear these?

gclssvglx: Rebased and updated to use the `FormHelper`

app/models/parted.rb
gclssvglx: While I'm happy to keep `is invalid` if you feel it needs to be there, I personally do not feel it adds anything or helps the user diagnose the issue.  If there are other reasons that I'm unaware of as to why you feel this error is useful, I'd be grateful if you could help me understand.

I would point out that this error will never be shown alone - it'll always be accompanied by a more meaningful and helpful message (e.g. `can't be blank`).  And having both `is invalid` and `can't be blank` does offer any further assistance to the user in my view as they can see it can't be blank, so do not also need to be told it's invalid.

Sorry if I'm missing something.

test/models/parted_test.rb
gclssvglx: See https://github.com/alphagov/publisher/pull/1486#discussion_r688477271

app/models/parted.rb
gclssvglx: Have removed this commit - thanks.

app/views/shared/_part.html.erb
```diff
@@ -17,6 +17,9 @@
           <span class="form-wrapper">
             <%= f.text_field :title, disabled: @resource.locked_for_edits?, class: "title form-control" %>
           </span>
+          <% if f.object.errors[:title].any? %>
+            <%= form_errors(f.object.errors[:title]) %>
+          <% end %>
```
kevindew, line 22: As per https://github.com/alphagov/publisher/pull/1488#discussion_r689405171 convention dictates that this shouldn't have the conditional check

gclssvglx: Fixed - thanks

kevindew (approved):

PR #1488, opened 2021-08-09
Fix SimpleSmartAnswer UI issues in Publisher

gclssvglx:
## What

Model validation error messages in Publisher are currently not being displayed to users.

This PR improves the `SimpleSmartAnswer` format's `Node` UI by showing model validation error messaging, and fixing style/formatting issues with `Options`. 

## Why

[Trello](https://trello.com/c/HTrLSx0G/549-fix-simplesmartanswer-ui-issues-in-publisher)

### Before
before

### After

Screenshot 2021-08-11 at 12 07 46

### DIFF
--- a/app/views/simple_smart_answers/_node.html.erb
+++ b/app/views/simple_smart_answers/_node.html.erb
@@ -4,10 +4,12 @@
     <%= f.link_to_remove "<i class=\"glyphicon glyphicon-remove glyphicon-smaller-than-text add-right-margin\"></i> Remove node".html_safe, :class => "btn btn-default pull-right remove-node" %>
   </legend>
   <%= f.hidden_field :order, class: 'node-order' %>
+
   <div class="form-group">
-    <span class="form-wrapper">
+    <div class="form-wrapper">
       <%= f.text_field :title, placeholder: "The title of the outcome", class: "node-title form-control" %>
-    </span>
+    </div>
+    <%= form_errors(f.object.errors[:title]) %>
   </div>
 
   <div class="form-group">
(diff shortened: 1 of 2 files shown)


kevindew (commented):
Looks great, thanks for this.

My only comment is that we should replace the 3 chunks of error lists with the `form_errors` helper once we've merged https://github.com/alphagov/publisher/pull/1478 and rebased this on it.

gclssvglx:
> Looks great, thanks for this.
> 
> My only comment is that we should replace the 3 chunks of error lists with the `form_errors` helper once we've merged #1478 and rebased this on it.

Rebased and updated to use the `FormHelper` - thanks

app/views/simple_smart_answers/_node.html.erb
```diff
@@ -4,10 +4,14 @@
...
+
   <div class="form-group">
-    <span class="form-wrapper">
+    <div class="form-wrapper">
       <%= f.text_field :title, placeholder: "The title of the outcome", class: "node-title form-control" %>
-    </span>
+    </div>
+    <% if f.object.errors.has_key?(:title) %>
+      <%= form_errors(f.object.errors[:title]) %>
+    <% end %>
```
kevindew, line 14:
```suggestion
    <%= form_errors(f.object.errors[:title]) %>
```

It looks like [we have a convention already](https://github.com/alphagov/publisher/blob/a6438219a710f8dc6ddb23fd59eb9db0f7c71000/app/helpers/form_helper.rb#L18) to call this helper whether there are errors or not.

kevindew: Same thing on the other two instances.

kevindew (commented):
Thanks

app/views/simple_smart_answers/_node.html.erb
gclssvglx: Yes, sorry.  Fixed

kevindew (approved):

PR #1492, opened 2021-08-16
Improve invalid slug validation messages

gclssvglx:
## What 

Currently, the messaging around invalid slug's for GuideEdition, TransactionEdition, and SimpleSmartAnswerEdition simply states that the slug `is invalid`. This does not help the user understand why the slug is invalid, so diagnosing the issue takes longer than it would if a more meaningful message were output, and correcting the issue is a somewhat trial-and-error experience - especially for the novice user.

This pull request adds more helpful and specific messages when the user experiences an invalid slug.

## Why

[Trello](https://trello.com/c/SvOnnXeB/530-update-parts-and-variants-validation-errors-in-publisher)

### DIFF
--- a/app/models/part.rb
+++ b/app/models/part.rb
@@ -19,7 +19,7 @@ class Part
   validates :title, presence: true
   validates :slug, presence: true
   validates :slug, exclusion: { in: %w[video], message: "Can not be video" }
-  validates :slug, format: { with: /\A[a-z0-9\-]+\Z/i }
+  validates :slug, format: { with: /\A[a-z0-9\-]+\Z/i, message: "can only consist of lower case characters, numbers and hyphens" }
   validates_with SafeHtml
   validates_with LinkValidator
 end
--- a/app/models/simple_smart_answer_edition/node/option.rb
+++ b/app/models/simple_smart_answer_edition/node/option.rb
@@ -15,7 +15,7 @@ class SimpleSmartAnswerEdition < Edition
       default_scope -> { order_by(order: :asc) }
 
       validates :label, :next_node, presence: true
-      validates :slug, format: { with: /\A[a-z0-9-]+\z/ }
+      validates :slug, format: { with: /\A[a-z0-9-]+\z/, message: "can only consist of lower case characters, numbers and hyphens" }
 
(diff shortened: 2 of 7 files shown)


kevindew (commented):
Nice one, this looks a much improved user experience. I've just added a comment to resolve.

I've ummed and ahhed about whether it's better to have a SlugValidator since we repeat this configuration 3 times but I think it's good as it is, maybe something to consider if we find ourselves having to add a fourth.

app/models/part.rb
```diff
@@ -19,7 +19,7 @@ class Part
   validates :title, presence: true
   validates :slug, presence: true
   validates :slug, exclusion: { in: %w[video], message: "Can not be video" }
-  validates :slug, format: { with: /\A[a-z0-9\-]+\Z/i }
+  validates :slug, format: { with: /\A[a-z0-9\-]+\Z/i, message: "Only lower case characters, numbers and hyphens are allowed" }
```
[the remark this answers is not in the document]
gclssvglx: Ah cool - thanks for pointing that out 👍   I've updated as your suggestion.  Thanks.

app/models/part.rb
```diff
@@ -19,7 +19,7 @@ class Part
   validates :title, presence: true
   validates :slug, presence: true
   validates :slug, exclusion: { in: %w[video], message: "Can not be video" }
-  validates :slug, format: { with: /\A[a-z0-9\-]+\Z/i }
+  validates :slug, format: { with: /\A[a-z0-9\-]+\Z/i, message: "only lower case characters, numbers and hyphens are allowed" }
```
kevindew, line 22: This still won't read correctly as a full message and should start with a verb. For example:

```
irb(main):008:0> p = Part.new(slug: "not valid"); p.valid?; p.errors.full_messages
=> ["Title can't be blank", "Slug only lower case characters, numbers and hyphens are allowed"]
```

gclssvglx: Ah!  Yes, I see the problem - thanks.  How about this...?

```
Slug can only consist of lower case characters, numbers and hyphens
```

kevindew (approved):

PR #1494, opened 2021-08-16
Convert input fields to use FormHelper in Publisher

gclssvglx:
## What

A `FormHelper` module was introduced into Publisher via [this PR](https://github.com/alphagov/publisher/pull/1478).

This helper creates a consistent HTML structure, classes, help and error blocks etc for input fields (see example below). But it has so far only been applied to fields that can generate model validation errors.

Ideally, we want a consistent approach and structure across as many fields as possible - wether or not they generate model validation errors.

This change converts all fields - where it is sensible and possible to do so - to use the `FormHelper` module. 

We ignore the following - retired - formats:

- CampaignEdition
- ProgrammeEdition
- VideoEdition

### Example

This `FormHelper` entry...

```ruby
form_group(f, :field_id, label: "Custom field label", help: "Some helpful text") do
  form.text_field :field_id, class: "form-control"
end
```

Generates this HTML structure...

```html
<div class="form-group">
  <div class="form-label">
    <label for="field_id">Custom field label</label>
  </div>
  <div class="form-wrapper">
    <input type="text" name="field_id" id="field_id" class="form-control">
  </div>
  <div class="help-block">Some helpful text</div>
  <ul class="help-block error-block"></ul>
</div>
```

## Why

A consistent approach to the way input fields are constructed will help those working with and updating the codebase, and simplify the user interface. It should also help if (the rather old version of) Bootstrap were to be updated or removed.

[Trello](https://trello.com/c/QnOEgTxW/531-convert-all-input-fields-to-use-formhelper-in-publisher)

### DIFF
--- a/app/views/answers/_fields.html.erb
+++ b/app/views/answers/_fields.html.erb
@@ -3,14 +3,9 @@
     <fieldset class="inputs">
       <%= render :partial => 'shared/common_edition_attributes', :locals => {:f => f} %>
 
-      <div class="form-group">
-        <span class="form-label">
-          <%= f.label :body %>
-        </span>
-        <span class="form-wrapper">
-          <%= f.text_area :body, rows: 20, disabled: @resource.locked_for_edits?, class: "input-md-10 form-control" %>
-        </span>
-      </div>
+      <%= form_group(f, :body) do %>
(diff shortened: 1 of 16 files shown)


JonathanHallam (approved):
We've had a chat offline and Graham's shown me this running locally. Everything looks the way it should and updates the DB in the way I expected. Happy with this going in! 🎉

PR #1501, opened 2021-09-14
Clean up inconsistencies between Parts and Variants

gclssvglx:
## What

The `Part` and `Variant` models both have `title` and `slug` attributes. However, there are differences between the HTML currently used for these input fields in the `Part` and `Variant` view partials. Notably, the errors list is within the `form-wrapper` for `Variant`s and outside for `Part`s.  

#### Variants (title and slug)...

```
<div class="form-group" id="edition_parts_title_input">
  <span class="form-label">
    <%= f.label :title %><abbr title="required">*</abbr>
  </span>
  <span class="form-wrapper">
    <%= f.text_field :title, disabled: @resource.locked_for_edits?, class: "title form-control" %>
    <%= form_errors(f.object.errors[:title]) %>
  </span>
</div>
```

#### Parts (title and slug)...

```
<div class="form-group" id="edition_parts_title_input">
  <span class="form-label">
    <%= f.label :title %><abbr title="required">*</abbr>
  </span>
  <span class="form-wrapper">
    <%= f.text_field :title, disabled: @resource.locked_for_edits?, class: "title form-control" %>
  </span>
  <%= form_errors(f.object.errors[:title]) %>
</div>
```

Much of the rest of the page structure is also the same.  Differing only in the additional model specific fields that each model has and the use of the words `part` or `variant` in some on-screen wording.  They even share the same JavaScript file used for updating the UI via Ajax (`ajax_save_with_parts.js`). 

We should simplify these view into a single common partial - this will make the HTML consistent - and convert `title` and `slug` to use the `FormHelper` introduced in [this PR](https://github.com/alphagov/publisher/pull/1478). We can also remove the `required` HTML abbreviation on these fields as the model validation error messages is now being shown to the user. 

## Why

These changes will make future updates and changes simpler and easier to make, and make this aspect of the app much easier to understand for new developers.

[Trello](https://trello.com/c/pK4RLqgV/580-clean-up-inconsistencies-between-parts-and-variants-in-publisher)


Rosa-Fox (commented):
Hi @gclssvglx,

Thanks for the descriptive PR message. I have checked that the HTML is in the correct form when rendering the partials and all looks really good. I also liked how it now makes use of the `form_group` method 👍 

A nitpick I have is that some of the code to renders the partials uses hash rockets `=>` and others use colons `child_record_type: 'part'` so it would be good to make them consistent.

The thing I am confused about is that when I try to create a guide or transaction locally and save a form with blank fields, I don't see any error messages displaying, despite there being presence validation in the `variant` and `part` models... the draft is saved so I wonder if this is a separate problem to this task, in that validations are not happening so it never displays on the view? Was this something you encountered?

gclssvglx:
Thanks @Rosa-Fox - I'll fix up the `:` verses `=>` issue into a new commit.

On the creation - yes, there is some odd behaviour going on. `Guide` will automatically create a slug from the title if you don't supply one.an odd difference that I've noticed between `Variant` and `Guide`. Frankly, I'm hoping not to fix this issue as it'll have consequences I think on the parent models. What happens with `Guide` is 

> Hi @gclssvglx,
> 
> Thanks for the descriptive PR message. I have checked that the HTML is in the correct form when rendering the partials and all looks really good. I also liked how it now makes use of the `form_group` method 👍
> 
> A nitpick I have is that some of the code to renders the partials uses hash rockets `=>` and others use colons `child_record_type: 'part'` so it would be good to make them consistent.
> 
> The thing I am confused about is that when I try to create a guide or transaction locally and save a form with blank fields, I don't see any error messages displaying, despite there being presence validation in the `variant` and `part` models... the draft is saved so I wonder if this is a separate problem to this task, in that validations are not happening so it never displays on the view? Was this something you encountered?

Thanks @Rosa-Fox - I've updated the `=>` to `:` to make them consistent. Perhaps we could have a chat about the `Guide` and `Variant` views as I am aware of some issues there.

Rosa-Fox (approved):
Thanks for updating the syntax @gclssvglx. Thank also for adding [the card to fix the underlying parts and variant validation issues](https://trello.com/c/fPn5QTnc/640-fix-publisher-validation-display-issues-with-guide-and-transaction-formats). Looks like a mountain to be climbed! 

I followed your instructions in the card to display an error. Looks good to me:
Screenshot 2021-09-22 at 15 09 05

gclssvglx:
Thanks @Rosa-Fox 👍

PR #1523, opened 2021-11-30
Set RAILS_LOG_TO_STDOUT in Dockerfile.

sengi:
This always needs to be set when a Rails app is running in a container, so it makes sense to set it in the Dockerfile rather than making the user remember to set it.


sengi:
Turns out this isn't needed; see alphagov/frontend#3055. Sorry for the noise.

PR #1549, opened 2022-02-22
Update downtime message inputs to include labels

davidgisbey:
### Description

This is part of the work to make our applications more accessible.

The `date_select` and `time_select` methods do not allow you to pass a label in as on option. This means that the fields are inaccessible for screen readers. I've used it to use more granular select fields (date, month, year etc.) and added labels.

### Screenshots

|Before|After|
|---------------|---------------|
|image|image|

### Trello card 

https://trello.com/c/1lMcyxjF/245-upgrade-orphaned-labels-mainstream

### DIFF
--- a/app/assets/stylesheets/downtime.scss
+++ b/app/assets/stylesheets/downtime.scss
@@ -7,3 +7,18 @@
     margin-right: $default-horizontal-margin;
   }
 }
+
+.downtime-dates p {
+  font-weight: bold;
+  font-size: 16px;
+}
+
+.date {
+  width: auto;
+}
+
+.joining-on {
+  @media screen and (min-width: 600px) {
+    margin-top: 40px;
+  }
+}
--- a/app/views/downtimes/_form.html.erb
+++ b/app/views/downtimes/_form.html.erb
@@ -6,27 +6,79 @@
    </p>
 <% end %>
 
-<div class="form-group form-inline js-start-time">
-  <%= f.label :start_time, 'This service will be unavailable from' %><br />
-  <span>
-  <%= f.time_select :start_time, { default: 1.hour.from_now.beginning_of_hour, ignore_date: true }, { class: 'date form-control' } %>
-</span>
-  <b class="add-left-margin add-right-margin">on</b>
-  <%= f.date_select :start_time, { order: [:day, :month, :year], default: Date.tomorrow,
-    start_year: Date.today.year, end_year: Date.today.year.next }, { class: 'date form-control' } %>
+<div class="downtime-dates">
+  <div class="form-group">
+    <p>This service will be unavailable from</p>
+  </div>
+
+  <div class="row js-start-time">
+    <div class="form-group date col-sm-1">
+      <%= f.label :start_time, "Hour", for: "downtime_start_time_4i" %>
+      <%= select_hour @downtime.start_time&.hour || 1.hour.from_now.beginning_of_hour, { field_name: "start_time(4i)", prefix: "downtime" }, { class: "form-control date" } %>
+    </div>
+
+    <div class="form-group date col-sm-1">
+      <%= f.label :start_time, "Minute", for: "downtime_start_time_5i" %>
+      <%= select_minute @downtime.start_time&.minute, { field_name: "start_time(5i)", prefix: "downtime" }, { class: "form-control date" } %>
+    </div>
+
+    <div class="form-group date col-sm-1">
+      <p class="joining-on">on</p>
+    </div>
+
+    <div class="form-group date col-sm-1">
+      <%= f.label :start_time, "Day", for: "downtime_start_time_3i" %>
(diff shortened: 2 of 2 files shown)


app/controllers/downtimes_controller.rb
```diff
@@ -48,6 +48,11 @@ def downtime_params
     params[:downtime].permit([
       "artefact_id",
       "message",
+    ]).merge!(date_params)
```
davidgisbey, line 51: I don't like this at all. As the fields are now built in a more granular way (hour, min & day, month, year) rather than using the `time_select` and `date_select` helper methods, they need to be given unique field names. 

This means that they don't correspond with the `start_time` and `end_time` attrs on the `Downtime` object. As a result you can't call `f.select_hour` etc.

I've got round this by merging the params in the controller. A more optimal solution would be to create an intermediary object like a FormObject with the correct attrs, but i'm not sure if it's worth introducing a new pattern just for this. 

Wdyt?

app/views/downtimes/_form.html.erb
```diff
@@ -7,26 +7,67 @@
...
 <div class="form-group form-inline js-stop-time">
-  <%= f.label :end_time, 'to' %><br />
-  <span>
-  <%= f.time_select :end_time, { default: 1.hour.from_now.beginning_of_hour, ignore_date: true }, { class: 'date form-control' } %>
-  </span>
+  <p>to</p>
+
+  <div class="date-label">
+    <%= f.label :end_time, "Hour", for: "date_end_time_4i" %>
+    <%= select_hour @downtime.end_time&.hour || 1.hour.from_now.beginning_of_hour, { field_name: "end_time(4i)" }, { class: "form-control date" } %>
```
davidgisbey, line 44: Having a form object would also help with setting these values as the form could have a default value on initialisation which would be called in the new controller action 
```
def new 
  @downtime = DowntimeForm.new 
end
```

and then a build_from_downtime method for edit 

```
def edit
  @downtime = DowntimeForm.build_from_downtime(Downtime.find(params[:id]
end 

class DowntimeForm

attr_accessor :all_the_accessors

def build_from_downtime(downtime)
  new(
    message: downtime.message
    etc.
  )
end
```

app/assets/stylesheets/downtime.scss
```diff
@@ -7,3 +7,26 @@
...
   }
 }
+
+.date-label label {
+  display: block;
+}
+
+.date-label {
+  display: inline-block;
+}
```
DilwoarH, line 17: are these needed still @davidgisbey ?

davidgisbey: Oh good shout! Will pull them out now

app/views/downtimes/_form.html.erb
```diff
@@ -6,27 +6,77 @@
...
+    <p class="joining-on">on</p>
+  </div>
+
+  <div class="form-group date col-sm-1">
+    <%= f.label :start_time, "Day", for: "date_start_time_3i" %>
+    <%= select_day @downtime.start_time&.day || Date.tomorrow.day, { field_name: "start_time(3i)" }, { class: "form-control date" } %>
+  </div>
+  <div class="form-group date col-sm-1">
+    <%= f.label :start_time, "Month", for: "date_start_time_2i" %>
+    <%= select_month@downtime.start_time&.month || Date.tomorrow.month, { field_name: "start_time(2i)" }, { class: "form-control date" } %>
```
brucebolt, line 34:
```suggestion
    <%= select_month @downtime.start_time&.month || Date.tomorrow.month, { field_name: "start_time(2i)" }, { class: "form-control date" } %>
```

app/controllers/downtimes_controller.rb
brucebolt: Could you specify a `prefix` of `downtime` for the `select_day` (etc) helpers (i.e. override the default `date` prefix)?  It also means none of the tests need updating.

I've put some suggested code for this in a branch: https://github.com/alphagov/publisher/compare/downtime-page-fixes-bruce?expand=1

davidgisbey:
Amazing. I had no idea you could prefix the attr's. Will apply these changes to my code 👍 Thanks Bruce

app/controllers/downtimes_controller.rb
davidgisbey: Amazing. I had no idea you could prefix the attr's. Will apply these changes to my code 👍 Thanks Bruce

brucebolt (approved):

app/assets/stylesheets/downtime.scss
```diff
@@ -7,3 +7,18 @@
     margin-right: $default-horizontal-margin;
   }
 }
+
+p {
+  font-weight: bold;
+  font-size: 16px;
+}
```
kevindew, line 14: Sorry I'm late to the party here. I was going to just click merge but this caught my eye. Won't this set the styling of paragraphs for the entire app?

davidgisbey: Good spot. Have pushed a fix

PR #1552, opened 2022-03-01
Use internal Frontend API

bilbof:
Publisher uses the public bank holidays API rather than the internal endpoint.

Using internal APIs is preferred - we won't need to give publisher credentials for this API in environments where this is necessary, and requests that stay in the VPC may be faster too.


fredericfran-gds (approved):
cool

PR #1553, opened 2022-03-01
Use GOVUK_ENVIRONMENT_NAME in place of SENTRY_CURRENT_ENV

bilbof:
SENTRY_CURRENT_ENV is set to integration-blue-aws in
integration whereas GOVUK_ENVIRONMENT_NAME is set to
integration.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -32,7 +32,8 @@ class ApplicationController < ActionController::Base
   end
 
   def notify_bad_request(exception)
-    if %w[integration staging].include?(ENV["SENTRY_CURRENT_ENV"]) && exception.message =~ /team-only API key/
+    # TODO: control this via the log level rather than an environment variable
+    if %w[integration staging].include?(ENV["GOVUK_ENVIRONMENT_NAME"]) && exception.message =~ /team-only API key/
       # in production we care about all errors
       # in staging and integration the team-only error may be encountered by
       # end-users who should see a more helpful error message


app/controllers/application_controller.rb
```diff
@@ -32,7 +32,7 @@ def squash_multiparameter_datetime_attributes(params, attribute_names)
   end
 
   def notify_bad_request(exception)
-    if %w[integration staging].include?(ENV["SENTRY_CURRENT_ENV"]) && exception.message =~ /team-only API key/
+    if %w[integration staging].include?(ENV["GOVUK_ENVIRONMENT_NAME"]) && exception.message =~ /team-only API key/
```
sengi, line 35: Ideally we'd control this via log level rather than having the app change its behaviour based on what environment it's being told it's in, but 🤷

sengi: A less yak-shavy compromise might be to add a `LOG_SENSITIVE_INFO` flag that defaults off and is enabled in non-prod, but that would still mean faffing about with govuk-puppet so 🤷

sengi (approved):

app/controllers/application_controller.rb
sengi: Goes without saying that this isn't your fault - the problem was basically already there so it's not really getting any worse. I'm just fretting about the precedent that it sets. It's fine though 😅

sengi: I'm honestly ok with this as is. Just don't want people thinking it's a good idea to add more logic like this 😅

PR #1603, opened 2022-05-06
Honour `RAILS_LOG_TO_STDOUT` for Sidekiq and GdsApi client logs.

sengi:
All of our logs need to go to stdout/stderr when running on Kubernetes.

- Honour the `RAILS_LOG_TO_STDOUT` environment variable for GdsApi client logs.
- Have Sidekiq log to stdout by default and make the config forward-compatible with Sidekiq 6.

See commit messages for details.

https://trello.com/c/Jef1t4xU

Pair: @nsabri1

### DIFF
--- a/config/initializers/gds_api.rb
+++ b/config/initializers/gds_api.rb
@@ -1,8 +1,5 @@
-require "gds_api/base"
 require "gds_api/asset_manager"
 require "plek"
-
-GdsApi::Base.logger = Logger.new(Rails.root.join("log/#{Rails.env}.api_client.log"))
 
 # Need to provide a bearer token to authenticate to the asset manager.
 # This can be loaded from an environment variable, or this file can
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -1,9 +1,10 @@
-# Sample configuration file for Sidekiq.
+# Configuration file for Sidekiq.
 # Options here can still be overridden by cmd line args.
-#   sidekiq -C config.yml
 ---
 :verbose: true
-:concurrency:  2
-:logfile: ./log/sidekiq.log
+:concurrency: 2
+<% if ENV.key?('SIDEKIQ_LOGFILE') %>
+:logfile: <%= ENV['SIDEKIQ_LOGFILE'] %>
+<% end %>
 :queues:
   - default


config/initializers/gds_api.rb
```diff
@@ -2,7 +2,13 @@
...
 require "plek"
 
-GdsApi::Base.logger = Logger.new(Rails.root.join("log/#{Rails.env}.api_client.log"))
+GdsApi::Base.default_options = {
+  logger: if ENV["RAILS_LOG_TO_STDOUT"].present?
+            Logger.new($stdout)
+          else
+            Logger.new(Rails.root.join("log/#{Rails.env}.api_client.log"))
+          end,
+}
```
kevindew, line 11: I've no idea why this application sets up a dedicated log file for GDS API Adapters, it looks very inconsitent with the rest of the stack. I'd suggest deleting this config.

```suggestion
```

config/sidekiq.yml
```diff
@@ -1,6 +1,8 @@
-# Sample configuration file for Sidekiq.
+# Configuration file for Sidekiq (but not when running in Kubernetes).
+#
+# This file is NOT used in EKS/Replatforming. See sidekiq_k8s.yml.
```
kevindew, line 3: I think we should try avoid having a separate file for this for simplicity.

We can actually use embedded Ruby in this file to reference environment variables.

What I'd actually suggest for this is that we set up an environment variable in puppet to specify the logfile and then have everywhere else log to STDOUT. This is because it's mostly annoying having the logfile in dev (where we've fought it in the past: https://github.com/alphagov/content-publisher/blob/3e2ad9d23852bbc691d52922283a33d2dbc9acb9/config/sidekiq.yml#L5-L7)

So I'd suggest having the file set-up like:

```
:logfile: <%= ENV.fetch("SIDEKIQ_LOGFILE", "~") %> # default to stdout, unless logfile specified
```

and then adding that env var to: https://github.com/alphagov/govuk-puppet/blob/5e3be50fee026d9d33d9c7f68c2251262573b48c/modules/govuk/manifests/deploy/config.pp#L121-L134 so all apps can pick it up (note: this app seems to be a little unconventional by not having ./log/sidekiq.json.log as a name which is the convention: https://github.com/alphagov/govuk-puppet/blob/5b5e249600eeddbd78e5292bdf4a8999c8b35a26/modules/govuk/manifests/app.pp#L442-L447)

kevindew (commented):
Thanks @sengi, I've made some counter suggestions to see if things can be a bit cleaner.

config/initializers/gds_api.rb
sengi: Ooh cool, thanks! There are a couple of others which do the same thing ([smart-answers and service-manual-publisher](https://sourcegraph.com/search?q=context:global+repo:alphagov+api_client.log&patternType=literal)). Happy to ditch it from those too if you are.

config/sidekiq.yml
sengi: Ooh ok, hadn't spotted that it puts the file through ERB :D Sounds great - let's go with that.

config/initializers/gds_api.rb
kevindew: Ah nice spot - that sourcegraph website looks really helpful!

Yes, they can go too. Feel free to tag me on reviews since I have the context.

config/sidekiq.yml
sengi: RFAL. I wrapped it in a conditional, for forward-compatibility with Sidekiq 6. Puppet PR on the way.

sengi: (I considered renaming the file to make it play nice with editor syntax plugins, but thought it safer not to for now since the filename is part of the command in the Procfile etc.)

sengi: Puppet PR: https://github.com/alphagov/govuk-puppet/pull/11673

PR #1608, opened 2022-05-13
Add paste html to govspeak

DilwoarH:
## What
Converts pasted formatted content to GovSpeak

## Why
This work will reduce the effort and time for publishers to create new content, including HTML attachments and as a result, increase interest in creating HTML attachments by default

https://trello.com/c/20l0GFJl/440-add-paste-html-to-govspeak-to-mainstream-publisher

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,13 +4,13 @@ ARG base_image=ruby:2.7.6-slim-buster
 FROM $base_image AS builder
 
 # TODO: have a separate build image which already contains the build-only deps.
-RUN apt-get update -qy && apt-get upgrade -y
+RUN apt-get update -qy && apt-get upgrade -y && apt-get install -y build-essential curl git
 
-# TODO: Look to remove Node v9 (below) (E2E tests currently fail without it).
-RUN curl -sL https://deb.nodesource.com/setup_9.x | bash -
-RUN apt-get update && apt-get install -y build-essential nodejs git npm && \
-    npm install -g yarn && \
-    npm install -g phantomjs-prebuilt@2 --unsafe-perm
+RUN curl -sL https://deb.nodesource.com/setup_lts.x | bash -
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
+RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+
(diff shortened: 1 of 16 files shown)


Dockerfile
```diff
@@ -2,31 +2,41 @@
...
 
 FROM $base_image
-RUN apt-get update -qy && apt-get upgrade -y && \
-    apt-get install -y nodejs
+
 # TODO: MONGODB_URI shouldn't be set here but seems to be required by E2E tests, figure out why.
 # TODO: Can ASSETS_PREFIX default to `/assets/publisher` within Publisher?
 ENV GOVUK_PROMETHEUS_EXPORTER=true RAILS_ENV=production GOVUK_APP_NAME=publisher ASSETS_PREFIX=/assets/publisher MONGODB_URI=mongodb://mongo/govuk_content_development
+
+COPY --from=builder /usr/bin/node* /usr/bin/
```
ollietreend, line 37: @kevindew does the runner image _need_ node installed? Just wondering why we don't also need yarn, in that case. Sprockets assets are compiled within the builder image and copied across – so I suspect node might be unnecessary. 🤔

kevindew: Good shout. I'd like to think it doesn't, but often it seems apps have a surprising need for a node runtime. I reckon, with the e2e tests, that this image is tested well enough that I can try remove it and we'll find out if it's a problem.

davidgisbey: Looks like it's not happy 

image

davidgisbey:
The code in the first 3 commits lgtm. Once the build is passing i'll deploy out and test it on integration.

Dockerfile
ollietreend: Yep – it looks like the `uglifier` gem, which is included in production, depends on `execjs` gem, which in turn depends on Node.js 😔

ExecJS::RuntimeUnavailable: Could not find a JavaScript runtime.

As for whether `uglifier` is really needed in production... I don't know? But since it's already there, I guess the simplest thing is just to give the runtime image a copy of Node.js

kevindew: Thanks for digging out that detail. 

That is a bit annoying. You do need uglifier for production, so that you can run assets:precompile in production - however it seems a bit frustrating that it demands it outside of runtime usage.

davidgisbey (approved):
Have manually tested a bunch of the fields and it's working nicely 👍 🚢

PR #1620, opened 2022-06-09
Render govspeak help for editions with fields that can use it and ensure it works on dynamically added fields

davidgisbey:
## Description 

We've done some work to allow paste to govspeak. As part of this work we've updated the govspeak guidance. However, currently, it's only rendered on Guides and Programmes.

We should expose it on any edition type that has fields that can use govspeak.

image

Additionally, it is not working for fields that are dynamically created using JavaScript. A list of the affected fields are:

 - The body fields for Parts (GuidEdition)
- The body fields for nodes (questions and answers for  SimpleSmartAnswerEdition)

## Trello card

https://trello.com/c/20l0GFJl/440-add-paste-html-to-govspeak-to-mainstream-publisher

### DIFF
--- a/app/views/editions/show.html.erb
+++ b/app/views/editions/show.html.erb
@@ -28,6 +28,14 @@
         <div role="tabpanel" class="tab-pane <% if active_tab.name == 'edit'%>active<% end %>" id="edit">
           <div class="link-check-report col-md-4">
             <%= render 'link_check_reports/link_check_report', edition: @edition, report: @edition.latest_link_check_report %>
+
+            <% if @edition.class.to_s.in?(Edition::HAS_GOVSPEAK_FIELDS) %>
+              <h3 class="remove-top-margin add-bottom-margin">Govspeak help</h3>
+
+              <div class="well">
+                <%= render "shared/govspeak_help" %>
+              </div>
+            <% end %>
           </div>
 
           <%= resource_form(@resource) do |f| %>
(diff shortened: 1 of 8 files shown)


app/views/editions/show.html.erb
```diff
@@ -28,6 +28,14 @@
         <div role="tabpanel" class="tab-pane <% if active_tab.name == 'edit'%>active<% end %>" id="edit">
           <div class="link-check-report col-md-4">
             <%= render 'link_check_reports/link_check_report', edition: @edition, report: @edition.latest_link_check_report %>
+
+            <% if [AnswerEdition, GuideEdition, HelpPageEdition, LicenceEdition, LocalTransactionEdition, PlaceEdition, ProgrammeEdition, SimpleSmartAnswerEdition, TransactionEdition].include?(@edition.class) %>
```
ollietreend, line 32: Do we know for sure if there are any types of edition which _don't include_ Govspeak fields? I've just had a quick look and can't seem to find any (but I could have missed something).

<details>
  <summary><strong>Helpful Terminal command</strong></summary>

> List models including `GOVSPEAK_FIELDS`:
> ```
> $ grep -rl GOVSPEAK_FIELDS app/models
> ```
> 
> And models without (use `-L` instead of `-l`):
> ```
> $ grep -rL GOVSPEAK_FIELDS app/models
> ```
</details>

I'm just thinking, this seems like a very long and specific looking list. And the code itself doesn't make it obvious why these ones are included and others aren't.

If there _are_ types of edition that we'd want to hide this for, I wonder if we could introspect the model to see if it has Govspeak fields. I think this could be a bit more intention revealing rather than relying on a hardcoded array.

```suggestion
            <% if @edition.class.const_defined?(:GOVSPEAK_FIELDS) %>
```

davidgisbey: yea good shout. Have added a constant to Editions

PR #1658, opened 2022-08-24
Fix HTML email conversion

benjamineskola:
Character encoding needs to be handled before decoding the HTML, otherwise potentially fails on unexpectedly-encoded messages.

Also now using a real email rather than a synthetic one to test with, as the previous PR assumed no charset errors. (I'm not really sure why the previous version didn't trigger the same error but this is the easiest way to be sure.)

This means potentially decoding twice which isn't quite ideal but restructuring it seemed like it would create further complexity — however might be worth attempting to find a better way anyway.


ollietreend:
There's probably not much else we can do until we have more examples of 'broken' emails. This seems to answer the immediate problem.

PR #1659, opened 2022-08-25
Avoid double-decoding unicode

benjamineskola:
It appears that `@message.body.charset` is always ASCII even when another encoding is used; `@message.charset` on the other hand appears to contain the correct value. Using this correctly flags the body as UTF-8 and avoids trying to decode it twice and mangling non-ASCII Unicode characters; the second pass through `try_decode` for an HTML-only message is therefore a no-op.


ollietreend (approved):
Nice work ⭐

PR #1679, opened 2022-09-21
Remove references to alphagov.co.uk

kevindew:

This removes the references to alphagov.co.uk principally as a way to indicate that it isn't used as part of the email fact-checking process. This was done to resolve confusion as to whether there is still a need for the Google account that powered this fact-checking process.

alphagov.co.uk was the original hostname that GOV.UK launched on before moving to alpha.gov.uk and finally www.gov.uk.


ollietreend (approved):
Nice work ✨🧹

PR #1682, opened 2022-09-23
Base the container image on govuk-ruby-base.

sengi:
This only affects Replatforming and publishing-e2e-tests. It should also reduce build times by at least few seconds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,38 +1,27 @@
-# (unless we decide to use Bitnami instead)
-ARG base_image=ruby:2.7.6-slim-buster
+ARG base_image=ghcr.io/alphagov/govuk-ruby-base:2.7.6
+ARG builder_image=ghcr.io/alphagov/govuk-ruby-builder:2.7.6
 
-FROM $base_image AS builder
-
-# TODO: have a separate build image which already contains the build-only deps.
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,38 +1,27 @@
...
-# TODO: We probably don't want assets in the image; remove this once we have a proper deployment process which uploads to (e.g.) S3.
 RUN GOVUK_APP_DOMAIN=www.gov.uk GOVUK_WEBSITE_ROOT=www.gov.uk bundle exec rails assets:precompile
 
+
 FROM $base_image
 
-# TODO: MONGODB_URI shouldn't be set here but seems to be required by E2E tests, figure out why.
-# TODO: Can ASSETS_PREFIX default to `/assets/publisher` within Publisher?
-ENV GOVUK_PROMETHEUS_EXPORTER=true RAILS_ENV=production GOVUK_APP_NAME=publisher ASSETS_PREFIX=/assets/publisher MONGODB_URI=mongodb://mongo/govuk_content_development
+# TODO: don't set MONGODB_URI here. Move it to publishing-e2e-tests.
```
nsabri1, line 21: Ok from it will be done

sengi: (discussed out-of-band - no action needed for now)

PR #649, opened 2017-10-17
Remove business finance support code

h-lame:
For: https://trello.com/c/RRJOTFQo/252-remove-business-support-code-that-still-exist-in-publisher

This is part 2 of the work for this card to remove business support code and data from publisher and make sure they're all redirected properly in publishing-api instead of router-data. Part 1 does the redirects (see: #648), part 3 removes the entries from router-data (see: alphagov/router-data#5), and part 4 removes the schema from content-schemas (see: alphagov/govuk-content-schemas#653).

We've left the `Artefacts` for business support documents but we've deleted all the business support editions and code that supports those editions.

PR #1686, opened 2022-09-28
Remove obsolete Area/AreaHelper code

KludgeKML:
Area is only used by AreaHelper, and the AreaHelper methods are not used anywhere AFAICT. Removing them allows publisher to be decoupled from Imminence completely (it's already not required to run publisher in govuk-docker, another hint it's not really used).

- remove code, tests, and test support.
- remove mapit initialiser
- remove imminence requirement mention from readme.

Following the merge of this PR, we can retire the /areas/ endpoint in Imminence, which is a blocker to retiring MapIt.


kevindew (approved):
Great find 💯 🥇 and excellent news to not have another unusual dependency.

As far as I can tell this code hasn't been utilised since 2017, with the last view removal in https://github.com/alphagov/publisher/pull/649. Quite amazing given some of the files have been amended multiple times in the subsequent 5 years.

PR #1697, opened 2022-10-06
Use Redis.new rather than Redis.current

edwardkerry:
Redis.current will be deprecated in Redis 5.0.

As Redis.current ultimately calls Redis.new we can migrate this method now to pave the way for future upgrades.

From Redis changelog from [5.0][1]:

‘Removed Redis.current. You shouldn't assume there is a single global Redis connection, use a connection pool instead, and libaries using Redis should accept a Redis instance (or connection pool) as a config. E.g. MyLibrary.redis = Redis.new(…).’

[Redis.current][2] just calls Redis.new so should just be able to swap methods.

[trello](https://trello.com/c/oGFWhrCO/1551-migrate-away-from-deprecated-redis-method)

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -29,6 +29,9 @@
       },
       "user_input": "params[:list].presence",
       "confidence": "High",
+      "cwe_id": [
+        22
+      ],
       "note": ""
     },
     {
@@ -45,6 +48,10 @@
       "location": null,
       "user_input": null,
       "confidence": "Medium",
+      "cwe_id": [
+        565,
+        502
+      ],
       "note": "The params[:list] argument is already checked before the template gets rendered by @presenter.acceptable_list?(@list)."
     },
     {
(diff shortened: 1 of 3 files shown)


config/brakeman.ignore
```diff
@@ -29,6 +29,9 @@
       },
       "user_input": "params[:list].presence",
       "confidence": "High",
+      "cwe_id": [
```
1pretz1, line 32: Are these brakeman changes related?

edwardkerry: These changes are not directly related but an artefact of re-running Brakeman to update the fingerprint of our ignoring the use of `Redis.current|new`. I'm happy to remove them if they are confusing?

config/brakeman.ignore
```diff
@@ -76,18 +83,21 @@
...
       "warning_code": 0,
-      "fingerprint": "d3d42c84d0b0acf57ccb1c9d1133ca43389fd725d0471016013e52bf59269b83",
+      "fingerprint": "814bc0a997725602aa2913dd902c8d77c28de9044d4562f60a898ad250d04b5a",
       "check_name": "SQL",
       "message": "Possible SQL injection",
       "file": "lib/csv_report_generator.rb",
       "line": 11,
       "link": "https://brakemanscanner.org/docs/warning_types/sql_injection/",
-      "code": "Redis.current.lock(\"publisher:#{Rails.env}:report_generation_lock\", :life => 15.minutes)",
+      "code": "Redis.new.lock(\"publisher:#{Rails.env}:report_generation_lock\", :life => 15.minutes)",
```
kevindew, line 100: Shall we change the code so these no longer produce warnings. 

I can't see there being any need for us to have Rails.env inside these locking keys - it's hard to think of a scenario where you'd need that distinction. So we could just simplify them to get rid of needing to update this further.

edwardkerry: Would you object to my making this a new card on our backlog? There are some similar/identical brakeman warnings in other apps, and it's not really connected to the intention of this PR, so might make sense to complete the Redis work and then tackle the Brakeman work en masse after.

config/brakeman.ignore
1pretz1: Sorry for the delayed response! I'm happy for the changes to be part of this PR, I just wondered why they are present. Thanks for explaining!

config/brakeman.ignore
KludgeKML: @kevindew just trying to clear up some of Ed's cards while he's off - are you okay with his suggestion (make this change a new one in our backlog separate from the redis fix)?

KludgeKML: I see Kev's away, I'm going to say we can expedite this and catch the other work Ed mentions in this card: https://trello.com/c/84yWQ0NJ/1662-clean-up-brakeman-errors-following-on-from-rediscurrent-redisnew-updates

ollietreend: Yep – that sounds good to me @KludgeKML 👍🏻

PR #1740, opened 2022-12-08
Update application content

johnpauldickie:
This PR includes several commits updating content in Mainstream Publisher to make actions and steps in the workflow clearer for content designers using the application.

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -187,7 +187,7 @@ class EditionsController < InheritedResources::Base
   def destroy
     if resource.can_destroy?
       destroy! do
-        flash[:success] = "#{description(resource)} destroyed"
+        flash[:success] = "Edition deleted"
         redirect_to root_url
         return
       end
--- a/app/views/shared/_review.html.erb
+++ b/app/views/shared/_review.html.erb
@@ -6,7 +6,7 @@
         <p><%= build_review_button(@resource, "skip_review", "Skip review").html_safe %></p>
       <% end %>
     <% else %>
-    <p><%= @resource.latest_status_action(Action::REQUEST_REVIEW).requester.name %> has sent this edition to be reviewed.</p>
+      <p><%= @resource.latest_status_action(Action::REQUEST_REVIEW).requester.name %> has sent this edition to be reviewed.</p>
       <div class="workflow-buttons">
         <%= review_buttons(@resource)%>
       </div>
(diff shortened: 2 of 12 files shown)


app/controllers/editions_controller.rb
```diff
@@ -187,7 +187,7 @@ def review
   def destroy
     if resource.can_destroy?
       destroy! do
-        flash[:success] = "#{description(resource)} destroyed"
+        flash[:success] = "Edition deleted"
```
ollietreend, line 190: Other flash messages in this file tend to show the format of the edition – i.e. the `#{description(resource)}` bit that you removed.

For example, when a new Edition is created:

https://github.com/alphagov/publisher/blob/572213a6f5a595c00546ce185b4b6367c938159b/app/controllers/editions_controller.rb#L54

So when creating a new help page edition, the flash message would read:
**Help page successfully created**

But this change means the delete action will always just read:
**Edition deleted**

regardless of the type of edition being deleted.

I'm just flagging it as I believe it introduces an inconsistency into the flash messages shown to users when working on editions.

app/views/shared/_review.html.erb
```diff
@@ -1,12 +1,12 @@
...
   <% if @resource.latest_status_action(Action::REQUEST_REVIEW) %>
-    <p><%= @resource.latest_status_action(Action::REQUEST_REVIEW).requester.name %> has requested this edition be reviewed.</p>
     <% if @resource.latest_status_action.requester == current_user %>
-      <p>That&rsquo;s you! You'll get an email when it has been reviewed.</p>
+      <p>You’ve sent this edition to be reviewed.</p>
       <% if skip_review? %>
         <p><%= build_review_button(@resource, "skip_review", "Skip review").html_safe %></p>
       <% end %>
     <% else %>
+    <p><%= @resource.latest_status_action(Action::REQUEST_REVIEW).requester.name %> has sent this edition to be reviewed.</p>
```
ollietreend, line 9: Code formatting nitpick: this line should be indented 2 spaces more now that it's inside the "else" block.

```suggestion
      <p><%= @resource.latest_status_action(Action::REQUEST_REVIEW).requester.name %> has sent this edition to be reviewed.</p>
```

ollietreend (commented):
@johnpauldickie this looks good to me!

Just one question, and a tiny nitpick about spacing.

app/controllers/editions_controller.rb
johnpauldickie: Thanks @ollietreend. We think this change is fine because, in your example of creating a new artefact, you are creating a new thing (for example, a guide) in comparison to deleting an edition (of a thing).

ollietreend (approved):
Looks good to me 👍🏻

PR #1797, opened 2023-03-31
Include locale in call to publishing api to discard draft edition

z-gooch:
There is a bug whereby when an edition with a non-english locale is deleted in Mainstream Publisher, it attempts to discard the draft from the Publishing API. However, this fails because the 'locale' option is not provided, so the Publishing API assumes its an english edition. This commit fixes that by ensuring that the locale is passed to Publishing API for this discard message.

[Trello ticket](https://trello.com/c/1aKN17NQ/1086-mainstream-publisher-doesnt-discard-non-english-editions-properly)

## What
- Adds locale to call to Publishing API to discard draft edition

## Why
At the moment there is a bug where because this locale is not provided, the publishing API assumes its an English edition and then fails to find and discard the appropriate edition.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -434,7 +434,7 @@ class Edition
   def destroy_publishing_api_draft
     return unless can_destroy?
 
-    Services.publishing_api.discard_draft(content_id)
+    Services.publishing_api.discard_draft(content_id, locale: artefact.language)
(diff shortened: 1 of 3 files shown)


kevindew (approved):
Nice one - this looks good to me.

Should this also be tested in a unit test for `destroy_publishing_api_draft`, if there is a test for that?

PR #1818, opened 2023-05-22
Retire licence format and remove code pertaining to creating new licences

kashifatcha:
## Dependancies

Before we freeze the old licence document, we want to ensure all licences are either published or archived. Currently [the only licences not in these states](https://publisher.publishing.service.gov.uk/?format_filter=licence&list=drafts&string_filter=&user_filter=all) haven’t been updated in at least 18 months, so we can probably just delete them.

## What

We'll want to freeze the creation and modification of licences in Mainstream Publisher before we import them.

There are PRs where previous document types have been retired https://github.com/alphagov/publisher/pull/586 . This format has been followed for licences and highlights that whilst the licences have been retired on Mainstream Publisher, they have been migrated to Specialist Licence Finder. This has been made clear in the README.

We don’t want to merge this until we’re ready to run the migration in production.

## Why

To prevent Licences in Publisher being created or edited after they've been imported to Specialist Publisher.

https://trello.com/c/QQJ0ajaZ/1661-disable-the-creation-and-modification-of-licences-in-mainstream-publisher


README.md
```diff
@@ -7,7 +7,7 @@ Publisher is sometimes referred to as "mainstream publisher".
 - [Completed transaction](https://www.gov.uk/done/make-lpa)
 - [Guide](https://www.gov.uk/council-tax-appeals)
 - [Help page](https://www.gov.uk/help/accessibility)
-- [Licence](https://www.gov.uk/day-nurseries-wales)
+<!-- - [Licence](https://www.gov.uk/day-nurseries-wales) -->
```
KludgeKML, line 10: I think this could just be deleted (if for no other reason than aesthetically it breaks the list rendering in markdown slightly).

test/models/licence_edition_test.rb
```diff
@@ -49,16 +49,6 @@ def setup
       end
     end
```
KludgeKML, line 51: Could this entire file be removed?

test/models/edition_test.rb
```diff
@@ -267,29 +267,6 @@ def draft_second_edition_from(published_edition)
     assert_equal edition.whole_body, new_edition.whole_body
   end
 
-  test "Cloning from LicenceEdition into AnswerEdition" do
```
KludgeKML, line 270: Hmm, might be worth double-checking if there's code to do this - maybe it's not needed? What did they do with the other retirements, did any of these have AnswerEdition cloning?

1pretz1 (commented):
Changes look good, nice work! There is more code in the repo that should be removed when we actually retire the licence document type, but that isn't really a concern of this PR as it is more of a cleanup task. 

Have you run this locally / integration? It would also be good / helpful to provide screenshots so reviewers can see the changes, especially as we don't have a bespoke test for a retired licence.

PR #1826, opened 2023-06-01
Add task to archive and unpublish licences after migration

KludgeKML:
Licences are to be migrated to Specialist Publisher. The import will be handled from that side. After import and after confirmation that everything has been carried across successfully, it will be necessary to archive/unpublish/redirect the licences.

This rake task uses a copy of the code from UnpublishService (slightly modified to allow a named task as the user) to perform this task for each published licence, although there are a list of licences for which the redirect is unnecessary (in this case they will just be archived and unpublished).

https://trello.com/c/9qWrtRto/2038-create-migration-to-archive-and-redirect-old-licences

### DIFF
--- a/lib/tasks/once_off/unpublish_licences.rake
+++ b/lib/tasks/once_off/unpublish_licences.rake
@@ -1,0 +1,62 @@
+PATHS_NOT_TO_MIGRATE = [
+  "hazardous-waste-producer-registration-wales",
+  "licence-to-photograph-wildlife-northern-ireland",
+  "auctioneer-s-permit-firearms-and-ammunition-northern-ireland",
+  "petroleum-exploration-licence-northern-ireland",
+  "chaperone-licence-northern-ireland",
+  "voluntary-registration-as-state-registered-hairdresser-england-scotland-wales",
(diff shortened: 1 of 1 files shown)


1pretz1 (approved):
Looks good to me! I had a query about draft handling but as we'll probably manually remove them before the import so it shouldn't be a concern of the migration 👍

lib/tasks/once_off/unpublish_licences.rake
```diff
@@ -0,0 +1,62 @@
...
+      if licence_edition.exact_route?
+        puts("WARNING: #{licence_edition.slug} skipped as it unexpectedly has an exact route")
+        next
+      end
+
+      redirect_url = PATHS_NOT_TO_MIGRATE.include?(licence_edition.slug) ? nil : "/find-licences/#{licence_edition.slug}"
+      artefact = licence_edition.artefact
+
+      artefact.assign_attributes(state: "archived", redirect_url:)
+      artefact.save_as_task!("once_off:unpublish_licences")
```
1pretz1, line 33: Nice catch!

PR #1846, opened 2023-06-29
Disable document actions after marked as retired

1pretz1:
Previously, retiring a document only disabled creating a brand new
edition and creating a new edition of an existing document. However,
there were a number of POST actions that a user could perform, like
update tagging, change format + saving the document.

If we mark the document type as 'retired' then we don't expect the
inidvidual documents to be modified. All the documents data can be
displayed but not edited + internal notes + unpublish feature (incase a
document needs unpublishing).

Trello:
https://trello.com/c/85Kl4zO7/2092-disable-editing-of-tags-when-marking-publisher-licences-as-retired

## Before

Screenshot 2023-06-29 at 13 00 17
Screenshot 2023-06-29 at 12 59 12
Screenshot 2023-06-29 at 12 59 59
Screenshot 2023-06-29 at 13 00 09

## After

Screenshot 2023-07-10 at 13 39 20
Screenshot 2023-07-10 at 12 58 45
Screenshot 2023-07-10 at 12 58 51
Screenshot 2023-07-10 at 12 59 00

### DIFF
--- a/app/helpers/tabs_helper.rb
+++ b/app/helpers/tabs_helper.rb
@@ -25,8 +25,9 @@ module TabsHelper
   def tabs_for(user, resource)
     tabs_to_remove = []
     tabs_to_remove << "admin" unless user.has_editor_permissions?(resource)
+    tabs_to_remove << "admin" if resource.retired_format?
     tabs_to_remove << "unpublish" unless user.govuk_editor?
 
-    tabs.reject { |tab| tabs_to_remove.include?(tab.name) }
+    tabs.reject { |tab| tabs_to_remove.uniq.include?(tab.name) }
   end
 end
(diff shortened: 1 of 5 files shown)


app/helpers/tabs_helper.rb
```diff
@@ -25,8 +25,9 @@ def tabs
   def tabs_for(user, resource)
     tabs_to_remove = []
     tabs_to_remove << "admin" unless user.has_editor_permissions?(resource)
+    tabs_to_remove << "admin" if resource.retired_format?
     tabs_to_remove << "unpublish" unless user.govuk_editor?
 
-    tabs.reject { |tab| tabs_to_remove.include?(tab.name) }
+    tabs.reject { |tab| tabs_to_remove.uniq.include?(tab.name) }
```
KludgeKML, line 31: You probably don't need the uniq here, include? doesn't care if a thing appears twice in a list. But no harm.

KludgeKML (approved):
LGTM

app/helpers/tabs_helper.rb
1pretz1: You're correct, after some brief thought, I think I'm going to keep the `uniq` call so the array isn't populated with duplicates which may be surprising

PR #1903, opened 2023-08-21
fix(editions) prevent calling method on nil

FelixMarcusMillne:
- when a request is sent without an assigned_to param to update an edition, an error was being thrown as the update_assignment function expected a value to always be passed


anatron (approved):
Looks good to me! 👍

PR #1917, opened 2023-09-06
add all edition reports for content workflow and edition churn

FelixMarcusMillne:
- also with additional time and date created fields split from editioned_at to allow for easier data parsing

Making these changes based on a request by the content improvement team to better understand the rate of change of editions in publisher - they need the historical data, not just that of all current editions, so I'm adding in a couple of extra reports for them.

### DIFF
--- a/app/presenters/all_content_workflow_presenter.rb
+++ b/app/presenters/all_content_workflow_presenter.rb
@@ -1,0 +1,39 @@
+class AllContentWorkflowPresenter < CSVPresenter
+  def initialize(scope = Edition.all)
+    super(scope)
+    self.column_headings = %i[
+      content_title
+      content_slug
+      content_url
+      current_status
+      stage
+      format
+      current_assignee
+      created_at
+      date_created
+      time_created
+    ]
+  end
+
+private
+
+  def build_csv(csv)
+    csv << column_headings.collect { |ch| ch.to_s.humanize }
+    scope.each do |item|
+      item.actions.each do |action|
+        csv << [
+          item.title,
+          item.slug,
+          "#{Plek.website_root}/#{item.slug}",
+          item.state,
+          action.request_type,
+          item.format,
+          item.assignee,
+          action.created_at.to_fs(:db),
+          action.created_at.to_date.to_s,
+          action.created_at.to_fs(:time),
+        ]
+      end
+    end
+  end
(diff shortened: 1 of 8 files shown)


config/brakeman.ignore
```diff
@@ -107,7 +153,7 @@
       "check_name": "Redirect",
       "message": "Possible unprotected redirect",
       "file": "app/controllers/reports_controller.rb",
-      "line": 21,
+      "line": 25,
```
anatron, line 156: Is it worth keeping these ordered by line number?

app/presenters/all_content_workflow_presenter.rb
```diff
@@ -0,0 +1,39 @@
...
+    super(scope)
+    self.column_headings = %i[
+      content_title
+      content_slug
+      content_url
+      current_status
+      stage
+      format
+      current_assignee
+      created_at
```
anatron, line 12: Do we need to have `date_created` and `time_created` in the `initialize` function when we have `created_at` (assuming we keep all 3 in the `build_csv` method)?

config/brakeman.ignore
FelixMarcusMillne: This is autogenerated by brakeman - I suppose we could, but it'll mean constantly fighting against the tool to do it...

FelixMarcusMillne: Given it's not really for human consumption I'd prefer to leave it. Thoughts?

app/presenters/all_content_workflow_presenter.rb
FelixMarcusMillne: This was explicitly requested by those who are consuming the data - the _better_ thing to do in my mind would be to remove the `created_at` field, but I left it in there for consistency with the other reports 🤔

FelixMarcusMillne: Just to be _overbearingly_ explicit, the `column_headings` array in the `initialize` function need to match perfectly with what we create in the `build_csv` function (as this all needs to line up properly in the right order on the csv file). So, if we add or remove something from the first array then we need to do the same in the one below, yeah? 😄

sivsubra-tw (approved):
Thanks for the clarification, Felix! Changes look good 👍

PR #1942, opened 2023-10-19
Add ADR for using mermaid to visualise simple smart answers

cynthia-anya:
## Description
This PR will add an ADR, that proposes we should use mermaid.js to add visualisations of simple smart answers.

This was following a review spike into the best library to adopt. 

This PR does not cover the implementation of the required behaviour, which will be raised in future PRs.

## Trello
[Trello card](https://trello.com/c/FtSuDOJa/545-spike-for-decision-tree-visualisation)

### DIFF
--- a/docs/arch/adr-006-use-mermaid-for-simple-smart-answer-visualisation.md
+++ b/docs/arch/adr-006-use-mermaid-for-simple-smart-answer-visualisation.md
@@ -4,9 +4,19 @@
 Proposed
 
 ## Context
-Following user research it was identified that Content Designers (CDs) were struggling to understand the flow of a smart answer, communicate this with colleagues and confidently edit it. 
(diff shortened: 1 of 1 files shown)


docs/arch/adr-006-use-mermaid-for-simple-smart-answer-visualisation.md
```diff
@@ -0,0 +1,35 @@
+# ADR 006 - Use Mermaid for Simple Smart Answer Visualisation
+
+## Status
+Proposed
+
+## Context
+Following user research it was identified that Content Designers (CDs) were struggling to understand the flow of a smart answer, communicate this with colleagues and confidently edit it.
```
andy-collon, line 7: Was it mentioned that the CDs have arrived at solutions outside of the system to visualise the more complex flows? And the ask is to bring that into the system to reduce context switching and extra overhead in converting formats. Feels like there's a sentence or two missing between the stated problem and the evaluation as well, what it is we're looking to make a decision on.

andy-collon (approved):
LGTM

PR #1952, opened 2023-10-30
Revert "Bump mongoid from 8.1.2 to 8.1.3"

andy-collon:
Reverts alphagov/publisher#1948
Reason being, the release didn't get through to production and while we haven't figured out why right now, removing this as a variable saves us worrying about it.


anatron (approved):
LGTM

PR #1975, opened 2023-12-05
Handle errors connecting to fact check mailbox

mtaylorgds:
Occasionally, there is an error when connecting to the mailbox in production. It seems a very transient error, and since the job is run every few minutes isn't really a problem, but if the exception is left unhandled a Sentry alert is raised. Log the error and move on.

We can't determine whether there are any unprocessed emails if we couldn't connect to the mailbox, so don't send any stats to Graphite.

[Trello card](https://trello.com/c/E6NTvtxY)


cynthia-anya (approved):
LGTM

PR #2027, opened 2024-01-29
Initial work for Downtime Index page transition to Design System

patrickpatrickpatrick:
## What

- Add new feature `design_system_downtime_pages` to the FlipFlop config.
- Update the routes with new constraint.
- Duplicate the controllers, views and test of Downtimes and append the `legacy_` prefix to them.
- Add new test in the routes integration test for the new feature.
- Add new test for legacy views in integration for Downtimes.
- Make the new Downtimes index page use the Design System layout.
- Add flash back to the Design System layout for integration tests (and also for future work to flash success/failure messages).

## Why

As part of the work to transition Publisher to the Design System from Bootstrap.

### Before

Screenshot 2024-01-29 at 16 16 37

### After

Screenshot 2024-01-29 at 16 15 42

[Relevant Trello Card](https://trello.com/c/c1023Flb/620-transition-downtime-hub)

### DIFF
--- a/app/views/layouts/design_system.html.erb
+++ b/app/views/layouts/design_system.html.erb
@@ -17,8 +17,26 @@
   } %>
 
   <div class="govuk-width-container">
-    <main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">
 
+    <main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">
+      <% [:success, :info, :warning, :danger, :notice, :alert].select { |k| flash[k].present? }.each do |k| %>
+        <%
+          case k
+          when :notice
+            alert_class = "success"
+          when :alert
+            alert_class = "danger"
+          else
+            alert_class = k
+          end
+        %>
+        <div class="alert alert-<%= alert_class %>"
+          data-module="auto-track-event"
+          data-track-action="alert-<%= alert_class %>"
+          data-track-label="<%= strip_tags(flash[k]) %>">
+            <%= flash[k] %>
+        </div>
+      <% end %>      
       <% if yield(:title).present? %>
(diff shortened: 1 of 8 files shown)


app/views/layouts/design_system.html.erb
```diff
@@ -18,6 +18,19 @@
 
   <div class="govuk-width-container">
     <main class="govuk-main-wrapper" id="main-content" role="main">
+      <% [:success, :info, :warning, :danger, :notice, :alert].select { |k| flash[k].present? }.each do |k| %>
```
mtaylorgds, line 21: Should we be using the design system components ([Error Alert](https://components.publishing.service.gov.uk/component-guide/error_alert), [Success Alert](https://components.publishing.service.gov.uk/component-guide/success_alert), [Notice](https://components.publishing.service.gov.uk/component-guide/notice)), rather than copying this pattern from the legacy layout template?

patrickpatrickpatrick: I thought that implementing the design system component would be a separate task in of itself. I think the flash message probably shouldn't have been removed initially, which is why I'm putting it back for now.

mtaylorgds: Maybe we need a team discussion around this, as I would expect that without using the components from the design system, the look of the flash messages won't be acceptable for releasing these pages, in which case we need to get a story for doing that done ASAP.

I think we probably need a discussion around how we're going to release pages (i.e. turn the default state of the toggle to on) anyway.

patrickpatrickpatrick: I have implemented the [flash messages as part of the actual page content itself](https://github.com/alphagov/publisher/pull/2031).

cynthia-anya (approved):
LGTM

PR #2047, opened 2024-02-06
Fix randomly failing Downtime integration test

patrickpatrickpatrick:
## What

- Add feature toggle to downtime_with_invalid_dates_test
- Create new legacy_downtimes_with_invalid_dates_test
- Add feature toggle to legacy_downtime_integration_test
- Update downtime_with_invalid_dates_test link name
 
## Why

Prevent randomly failing test.

### DIFF
--- a/test/integration/legacy_downtime_integration_test.rb
+++ b/test/integration/legacy_downtime_integration_test.rb
@@ -14,6 +14,9 @@ class LegacyDowntimeIntegrationTest < JavascriptIntegrationTest
     WebMock.reset!
     stub_any_publishing_api_put_content
     stub_any_publishing_api_publish
+
+    test_strategy = Flipflop::FeatureSet.current.test!
+    test_strategy.switch!(:design_system_downtime_index_page, false)
   end
 
   test "Scheduling new downtime" do
--- a/test/integration/legacy_downtime_with_invalid_dates_test.rb
+++ b/test/integration/legacy_downtime_with_invalid_dates_test.rb
@@ -1,6 +1,6 @@
 require "integration_test_helper"
 
-class DowntimeWithInvalidDates < ActionDispatch::IntegrationTest
+class LegacyDowntimeWithInvalidDates < ActionDispatch::IntegrationTest
   setup do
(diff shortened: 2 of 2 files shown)


test/integration/downtime_integration_test.rb
```diff
@@ -19,6 +19,11 @@ class DowntimeIntegrationTest < JavascriptIntegrationTest
     test_strategy.switch!(:design_system_downtime_index_page, true)
   end
 
+  teardown do
+    test_strategy = Flipflop::FeatureSet.current.test!
+    test_strategy.switch!(:design_system_downtime_index_page, false)
```
mtaylorgds, line 24: I don't think this will have any effect?

`Flipflop::FeatureSet.current.test!` [replaces the existing strategies](https://github.com/voormedia/flipflop/blob/0d70d8e33483a9c0282ed8d6bca9c5ccd61e61e8/lib/flipflop/feature_set.rb#L56C26-L56C50) (i.e. the previous test strategy) with a new `Strategies::TestStrategy`, which won't have any features set.

I think if we want to set the flag to false again we'd need to save the `test_strategy` from the setup into a field and use that in the teardown (though I'm not convinced that will necessarily have the desired effect either).

mtaylorgds: Having investigated this a bit more, it's line 23 that is having an effect (`test_strategy = Flipflop::FeatureSet.current.test!`) by overwriting the existing test strategy with a new one.

The intermittent test failure can be triggered by running `rails test test/integration/downtime_integration_test.rb test/integration/downtime_with_invalid_dates_test.rb --seed 37215`.

I think we should probably have a `LegacyDowntimeWithInvalidDates` which sets the flag to `false` (which we should probably also do in `LegacyDowntimeIntegrationTest`), and a `DowntimeWithInvalidDates` where we set the flag to `true` (and update the tests to work with it being `true`).

mtaylorgds (changes requested):

mtaylorgds (approved):
LGTM

PR #2064, opened 2024-02-15
Transition downtime message Javascript

mtaylorgds:
## What

Create a new Javascript module that generates the downtime message when adding/editing downtime for a publication, that works with the new add/edit downtime form components.

## Why

The old page used dropdown lists for selecting the date components, whereas the transitioned page uses text inputs.

The old code disabled the form, which is not accessible, so the new code does not do this.

Furthermore, the old code generated two messages that appeared in slightly different places on the form, whereas the new page only has one message to display.

## Trello ticket

[Trello card](https://trello.com/c/0J3X18AA).

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,2 +1,4 @@
+//= require_directory ./modules
+
 //= require govuk_publishing_components/dependencies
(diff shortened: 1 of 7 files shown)


anatron (approved):
Looks good! :)

PR #2067, opened 2024-02-16
Transition add downtime page/rename datetime fields

mtaylorgds:
## What

Rename the datetime inputs on the new "Add downtime" page to match those of the original page.

Extracts the fields to a partial to reduce duplication.

Also renames the error summary in preparation for replacement with a new version using GOV.UK Design System components.

## Why

There is some code in `ApplicationController#squash_multiparameter_datetime_attributes` that combines the date time form parameters used by the legacy view into a single parameter for use within the controller. This change enables making use of this code without having to modify it.

## Trello

[Trello ticket](https://trello.com/c/0J3X18AA)


anatron (approved):
looks good!

PR #2079, opened 2024-02-29
Transition add downtime page/field level validation messages

mtaylorgds:
## What

- Update validations on the "Add downtime" page now that we're using text fields for all of the input fields.
- When there are validation errors, populate the fields with the values that were submitted.
- Make model validation error messages consistent (so the `message` does not include the field name at the start, which is prepended when using `full_message`).

## Why

Part of the transition of Mainstream Publisher to the GOV.UK Design System.

## Trello

[Trello ticket](https://trello.com/c/0J3X18AA)

### DIFF
--- a/app/controllers/downtimes_controller.rb
+++ b/app/controllers/downtimes_controller.rb
@@ -47,19 +47,11 @@ class DowntimesController < ApplicationController
 private
 
   def downtime_params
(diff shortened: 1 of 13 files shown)


anatron (approved):
Looks good!

PR #2087, opened 2024-03-11
Transition edit downtime and destroy downtime to use the design system

anatron:
[Trello card](https://trello.com/c/i3Jjd1sr/157-upgrade-edit-downtime)

This PR is to upgrade the edit downtime page to use the design system. There is also a new delete downtime page too that uses the design system.

Edit down page before:
Screenshot 2024-03-11 at 15 41 10

Edit downtime page after:
Screenshot 2024-03-11 at 15 38 46

Delete downtime page:

Screenshot 2024-03-11 at 16 00 01

Changes also include:
- Adding a new style in the downtime.scss to make the downtime remove-link red.
- The destroy method in the downtimes_controller opens the delete page. (Previously this did not exist as there was no delete page as it was on the edit page).
- Adding delete to the routes and added it to the design_system_downtime_edit flipflop switch.

### DIFF
--- a/app/controllers/downtimes_controller.rb
+++ b/app/controllers/downtimes_controller.rb
@@ -39,7 +39,7 @@ class DowntimesController < ApplicationController
   def update
     @downtime = Downtime.for(@edition.artefact)
 
-    if params["commit"] == "Cancel downtime"
+    if !params[:downtime]
       DowntimeRemover.destroy_immediately(@downtime)
       flash[:success] = "#{edition_link} downtime message cancelled".html_safe
       redirect_to downtimes_path
@@ -50,6 +50,11 @@ class DowntimesController < ApplicationController
     else
       render :edit
     end
+  end
+
+  def destroy
+    @downtime = Downtime.for(@edition.artefact)
+    render :delete
   end
 
 private
@@ -68,7 +73,9 @@ private
   end
 
   def process_params
-    squash_multiparameter_datetime_attributes(downtime_params, %w[start_time end_time])
+    if params[:downtime]
+      squash_multiparameter_datetime_attributes(downtime_params, %w[start_time end_time])
+    end
   end
 
   def edition_link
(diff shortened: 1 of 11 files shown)


app/controllers/downtimes_controller.rb
```diff
@@ -39,7 +39,7 @@ def edit
   def update
     @downtime = Downtime.for(@edition.artefact)
 
-    if params["commit"] == "Cancel downtime"
+    if !params[:downtime]
```
mtaylorgds, line 42: We're going to need to change the `update` method to do the same extra validation that is done in the `create` method now.

(We should probably add some test coverage around this.)

app/views/downtimes/_form.html.erb
```diff
@@ -7,32 +7,32 @@
   year: {
     name: "downtime[start_time(1i)]",
     label: "Year",
-    value: params.dig("downtime", "start_time(1i)"),
+    value: @downtime.start_time&.year || params.dig("downtime", "start_time(1i)"),
```
mtaylorgds, line 10: The changes in here to set the value (I think it's these changes) seems to have had the side-effect of meaning that when a value is omitted from one of the text boxes it gets the value "1" put into it when the validation errors are shown.

(We should probably add some test coverage around this.)

mtaylorgds (changes requested):
Noticed a few minor issues, predominantly around validation cases.

app/views/downtimes/_form.html.erb
mtaylorgds, line 10: I'm wondering whether we should reverse the order of these since, as it currently is, when trying to set an invalid value (e.g. omitting a day value) and submitting, the page renders with a validation error message, but the value in the text field will be the previously valid value, not the submitted invalid value, which might be confusing for users.

mtaylorgds (commented):
Just one more issue I noticed when submitting invalid date/times.

mtaylorgds (approved):
LGTM

PR #2128, opened 2024-04-05
Adds a new ADR that documents our decision to move to Postgres

andy-collon:
https://trello.com/c/S1BbvoQN/262-document-our-decision-to-replatform-our-db-to-postgresql

### DIFF
--- a/docs/arch/adr-009-replace-mongodb-with-postgres.md
+++ b/docs/arch/adr-009-replace-mongodb-with-postgres.md
@@ -1,0 +1,31 @@
+# ADR009 - Replace MongoDb with Postgres
+
+Date 2024-04-03
+
+## Status
+Accepted
+
+## Context
+
+Publisher uses MongoDb for persistence and we use Amazon DocumentDB in our deployed environments. As described by Amazon, 
+```Amazon DocumentDB (with MongoDB compatibility) is a fast, scalable, highly available, and fully managed document database service that supports MongoDB workloads.```
+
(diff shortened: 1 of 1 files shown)


mtaylorgds (approved):
LGTM, but worth other devs having a look first too.

richardTowers (approved):
Hello! Thanks for this!

Happy to approve this as an initial statement of intent. We'll eventually need a bit more detail on how we're going to structure things in the new postgres world, but I imagine we'll want to do some spikes to inform those decisions before we write too much in an ADR.

We've got a few different approaches to relational database schemas for publishing apps:
- [Whitehall](https://github.com/alphagov/whitehall/blob/main/db/schema.rb)
- [collections](https://github.com/alphagov/collections-publisher/blob/main/db/schema.rb)
- [content-publisher](https://github.com/alphagov/content-publisher/blob/main/db/schema.rb)
- [publishing-api](https://github.com/alphagov/publishing-api/blob/main/db/schema.rb) (not technically a publishing app, but some of the same data)

... and I guess we've got an option here to try to be consistent with one of those, or to introduce another model.

I think it's totally fine if we end up making the call to build a schema that's very similar to the structure in mongo at the moment - e.g. if that makes migration easier, or reduces the changes we have to make to the code so the project gets done sooner. Still, if we think it's doable without disrupting the project too much, it would be good to aim for a schema we're happy with. Personally, I like content-publisher and publishing-api, because they've abstracted the individual content types away from the database schema - they just have `documents` and `editions` etc.

... anyway, thanks for this! ✅

PR #2150, opened 2024-04-30
Upgrade Ruby to 3.3.1

Tetrino:
Needed for upcoming changes in Rails 7.1.x and 7.2. Also solves issues with running the suite locally on Apple Silicon laptops.


mtaylorgds (commented):
Should we update the `.ruby-version` file at the same time?

I _think_ this file is only used during Github builds, and the `ruby-version` file is used when running locally using `govuk-docker`, but I could well be wrong about that.

mtaylorgds (approved):
LGTM

PR #2152, opened 2024-05-01
Fix scheduled last updated at date

anatron:
[Trello card](https://trello.com/c/z6XJSVge/765-scheduled-help-pages-dont-have-their-last-updated-value-set-mainstream-publisher)

This will send the latest edition payload to the publishing api before asking it to publish, which will update the last updated at date when publish is scheduled.

### DIFF
--- a/test/unit/scheduled_publisher_test.rb
+++ b/test/unit/scheduled_publisher_test.rb
@@ -32,8 +32,12 @@ class ScheduledPublisherTest < ActiveSupport::TestCase
       assert @edition.reload.published?
     end
 
-    should "call downstream publish service" do
-      PublishService.expects(:call).with(@edition)
+    should "update publishing api with latest edition payload before publishing" do
+      sequence = sequence(:task_order)
+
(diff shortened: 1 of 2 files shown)


test/unit/scheduled_publisher_test.rb
```diff
@@ -32,6 +32,11 @@ class ScheduledPublisherTest < ActiveSupport::TestCase
       assert @edition.reload.published?
     end
 
+    should "should update publishing api with latest edition payload before publishing" do
```
mtaylorgds, line 35: The test name states  that it is checking that the publishing API is updated with the latest edition _before_ publishing, but the test doesn't actually check any ordering. If the order is important, it would be nice to check that in the test (or rename the test to not mislead readers).

syed-ali-tw: fair point, though ordering is important and thats what we wanted to convey through test, I am not sure if there is a way to test ordering.

mtaylorgds (approved):

PR #2166, opened 2024-05-13
224 create new table component

davidtrussler:
[Trello](https://trello.com/c/9EQaeTRw/224-create-new-table-component-for-publication-page-in-design-system)

This is a first iteration of the new table component for the Publisher landing page. It’s still a work in progress but at a useful stage for an initial review. It still needs some custom styling but seems to work pretty much as expected. There are currently no other components associated with this repo so this creates a new component guide section which can be viewed locally at this URL: http://publisher.dev.gov.uk/component-guide/table

I had a few concerns emerge in my mind as I was working on it, as detailed below. Apart from that the code is just good enough to make it work and will need some tidying plus I’d need to add a Jasmine test for the JS.  

Concerns referred to are: 
- Is the creation of a new component really the best way to achieve what is required here? To my mind the use of a component would suggest that we are introducing a reusable pattern but during its development this component has become ever more tightly tied to the specific requirement such that it would never be used for anything else.
- The actual implementation here rolls together a number of different existing components (table, accordion and summary list) and sprinkles some custom stuff on top to make it work as required. This feels a little bit inelegant but might  not be a huge issue. 
- I had some concerns about the accessibility of the component though I haven’t looked into that fully yet. Mostly this is just a feeling that the structure is a little bit overly complex.


mtaylorgds:
I agree that this does seem very specific to the one use-case, in which case we should probably make that clear from the naming.

I think it makes some sense to have the code for this pulled out from the page itself (to avoid the page being massive), but maybe just using a partial would suffice?

I can't comment so much on the accessibility aspects (maybe we can ask about that on Slack), but it does at least seem usable via keyboard.

mtaylorgds:
We don't need this any more; closing PR.

PR #2187, opened 2024-05-29
Tidy up legacy downtime to remove new and create endpoints

anatron:
[Trello card](https://trello.com/c/94QrMjUT/49-tidy-up-new-downtime)

Task is to remove the toggle and legacy code created for new and create endpoints.


config/features.rb
```diff
@@ -9,10 +9,6 @@
             description: "A feature only used by tests; not to be used for any actual features."
   end
 
-  feature :design_system_downtime_new,
```
mtaylorgds, line 12: There's still a reference to this feature in `test/integration/downtime_integration_test.rb`.

mtaylorgds (approved):
Looks good—just one minor comment, as there is a reference to set the value of the removed toggle in one of the integration tests (which has no effect now, but we should clean it up anyway).

PR #2200, opened 2024-06-10
Show popular links

anatron:
[Trello card](https://trello.com/c/6xb4YYE7/299-create-a-page-to-show-popular-links)

We want the ability to view the latest published popular links within Mainstream Publisher.

In order to be able to do that, this is the first story to allow us to navigate to popular links page and view the last version of popular links. The ability to view the popular links will be provided by subsequent tickets. In order to be able to view we needed to create the model so that we can insert the very first record.

### DIFF
--- a/test/integration/homepage_popular_links_test.rb
+++ b/test/integration/homepage_popular_links_test.rb
@@ -1,0 +1,35 @@
+require "integration_test_helper"
+
+class HomepagePopularLinksTest < JavascriptIntegrationTest
+  setup do
+    setup_users
+    @popular_links = FactoryBot.create(:popular_links)
+    visit_popular_links
+  end
+
+  should "show page title" do
+    assert_title "Popular on GOV.UK"
+  end
+
+  should "show 'Homepage' as a page title context" do
+    assert page.has_content?("Homepage")
+  end
+
+  should "have 6 links with title and url" do
+    assert page.has_css?(".govuk-summary-card__title", count: 6)
+    assert page.has_text?("Title", count: 6)
+    assert page.has_text?("URL", count: 6)
+  end
+
+  should "have popular links version and status" do
+    assert page.has_text?("Edition")
+    assert page.has_text?(@popular_links.version_number)
+    assert page.has_text?("Status")
+    assert page.has_text?("DRAFT")
+    assert page.has_css?(".govuk-tag--yellow")
+  end
+
+  def visit_popular_links
+    visit "/homepage/popular-links"
+  end
+end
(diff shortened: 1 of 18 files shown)


mtaylorgds:
Would be good to have a link to the Trello ticket in the PR description.

test/integration/homepage_popular_links_test.rb
```diff
@@ -0,0 +1,35 @@
...
+    setup_users
+    @popular_links = FactoryBot.create(:popular_links)
+    visit_popular_links
+  end
+
+  should "show page title" do
+    assert_title "Popular on GOV.UK"
+  end
+
+  should "show homepage as a page title context" do
```
mtaylorgds, line 14: Is "as" meant to be "has" here? Otherwise, this test name doesn't quite make sense to me.

test/models/edition_test.rb
```diff
@@ -234,6 +234,8 @@ def draft_second_edition_from(published_edition)
 
   # test cloning into different edition types
   Edition.subclasses.permutation(2).each do |source_class, destination_class|
+    next if source_class.instance_of?(PopularLinksEdition.class) || destination_class.instance_of?(PopularLinksEdition.class)
```
mtaylorgds, line 237: Should we be testing that a `PopularLinksEdition ` can't be cloned from/to?

mtaylorgds (commented):
I've added a few minor comments. I wasn't able to run it locally, so it's just based on my look through the code changes.

mtaylorgds (approved):
LGTM

davidtrussler:
Do we definitely need to add custom styles for the tags (`status_tag.scss`)? My understanding is that this is using the styles for tags as they exist now but they will be changed in a later version. If that's the case I think we should use the current styles and allow the update to happen as and when.

davidtrussler:
When re-using components from other repos (`_summary_card.html.erb`) we shouldn't just copy them into this repo as that loses the point of shared components. Having said that in this case it would involve a lot of work on both here and the repo it has come from so maybe it could be considered a further discrete piece of work. If we do do it that way I'd suggest we don't create it as a component in this repo but rather create it as a partial. The creation of components actually follows a slightly more complicated pattern.

davidtrussler:
I don't think the styles added in `_summary-card.scss` are actually applied to anything since the file is not referenced in `application.scss`. Also I think some of the elements they are intended to reference are not present. I'd also query why we need to over-ride the provided styles in this case.

syed-ali-tw:
> Do we definitely need to add custom styles for the tags (`status_tag.scss`)? My understanding is that this is using the styles for tags as they exist now but they will be changed in a later version. If that's the case I think we should use the current styles and allow the update to happen as and when.

There were changes suggested by Rebecca to not have all caps for status, letters to not be bold and spacing to be bit less.
At this point suggestion is to adjust the css as the current style do not render these changes.

davidtrussler (approved):
LGTM! 👍

PR #2218, opened 2024-06-24
Create an 'edit' popular links page

anatron:
[Trello card](https://trello.com/c/0TGc3HTr/300-create-an-edit-popular-links-page)

This PR will allow us to edit existing popular links in publisher.
This includes: 
- Adding the sidebar on the show page which contains the 'Create new edition' button.
- Adding the create new edition page, which includes the 'Edit popular links' button and 'Publish' button in the sidebar.
- Adding the edit page, which includes the options to edit and save.
- Adding the controllers and routes for create, edit, update and publish.
- Adding validation to the models for invalid URLs
- Integration, functional and model tests for create, edit, update and publish. 
- Adding publish_popular_links to the workflow and action model.


app/views/homepage/popular_links/_sidebar.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="sidebar__button sidebar__border">
+  <h2 class="sidebar__heading">Options</h2>
+  <% if @latest_popular_links.state == "published" %>
+    <%= primary_theme_navigation_button("Create new edition", :post, create_popular_links_path) %>
+  <% else %>
+    <%= primary_theme_navigation_button("Edit popular links", :get, edit_popular_links_path(@latest_popular_links)) %>
```
davidtrussler, line 6: Do these buttons need to be part of a form or can we use the `Link button` variation of the [Button component](https://components.publishing.service.gov.uk/component-guide/button) and supply the URL for the edit page?

app/views/homepage/popular_links/_sidebar.html.erb
```diff
@@ -0,0 +1,9 @@
...
+    <%= secondary_theme_navigation_button("Publish", :post, publish_popular_links_path(@latest_popular_links)) %>
```
davidtrussler, line 7: Also might be better if we can use the appropriate variation of the button component here inside a form.

app/views/homepage/popular_links/_sidebar.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="sidebar__button sidebar__border">
```
davidtrussler, line 1: I think I would assign this container a single class name (like "sidebar-components") to keep with the existing naming convention and use that to add custom CSS.

app/views/homepage/popular_links/_sidebar.html.erb
syed-ali-tw: we did look at this and tried this but from what I understand, it wont be possible to use button components as it would not allow me to do a POST request.

app/views/homepage/popular_links/_sidebar.html.erb
syed-ali-tw: same as previous, not sure if we can use Button component as we want to do a POST on one of the buttons, hence we tried to keep it consistent for both scenarios where we did a POST or GET.

test/functional/homepage_controller_test.rb
```diff
@@ -17,4 +17,103 @@ class HomepageControllerTest < ActionController::TestCase
...
+                                   "3" => { "title" => "title3", "url" => "https://www.url3.com" },
+                                   "4" => { "title" => "title4", "url" => "https://www.url4.com" },
+                                   "5" => { "title" => "title5", "url" => "https://www.url5.com" },
+                                   "6" => { "title" => "title6", "url" => "https://www.url6.com" } } }
+
+      assert_equal new_title, PopularLinksEdition.last.link_items[0][:title]
+      assert_equal new_url, PopularLinksEdition.last.link_items[0][:url]
+    end
+
+    should "redirect to show path on success" do
```
mtaylorgds, line 77: Should we be checking for the flash message as well?

test/functional/homepage_controller_test.rb
```diff
@@ -17,4 +17,103 @@ class HomepageControllerTest < ActionController::TestCase
...
+    should "render show template" do
+      post :create, params: { id: @popular_links.id }
+
+      assert_response :ok
+      assert_template "homepage/popular_links/show"
+    end
+
+    should "create a new draft popular links" do
+      assert_equal 1, PopularLinksEdition.count
+      assert_equal "published", PopularLinksEdition.last.state
```
mtaylorgds, line 34: Do we need these assertions at the start? Aren't they just testing that the setup code worked?

Similar comment for subsequent tests.

test/integration/homepage_popular_links_test.rb
```diff
@@ -3,30 +3,165 @@
...
+      assert page.has_text?("Edition")
+      assert page.has_text?(@popular_links.version_number)
+      assert page.has_text?("Status")
+      assert page.has_text?("DRAFT")
+      assert page.has_css?(".govuk-tag--yellow")
+      assert page.has_text?("Edit popular links")
+    end
+
+    should "create a new record with next version and 'draft' status" do
+      row = find_all(".govuk-summary-list__row")
```
mtaylorgds, line 63: Do we need these assertions before clicking the button?

test/integration/homepage_popular_links_test.rb
```diff
@@ -3,30 +3,165 @@
...
+      assert page.has_content?("Popular on GOV.UK")
+    end
+
+    should "have 6 links with title and url" do
+      assert page.has_css?(".govuk-input", count: 12)
+      assert page.has_text?("Title", count: 6)
+      assert page.has_text?("URL", count: 6)
+    end
+
+    should "update record when clicked on 'Save'" do
```
mtaylorgds, line 99: Test name reads a bit awkward—perhaps "update record when 'Save' is clicked" would read slightly nicer?

test/integration/homepage_popular_links_test.rb
```diff
@@ -3,30 +3,165 @@
...
+
+    should "trim spaces from start and end of urls" do
+      fill_in "popular_links[1][url]", with: " www.abc.com "
+      click_button("Save")
+
+      assert page.has_text?("www.abc.com")
+      assert_not page.has_text?(" www.abc.com ")
+    end
+
+    should "render create page when clicked on 'cancel'" do
```
mtaylorgds, line 124: Similarly, this test name reads a bit awkward—perhaps "render create page when 'Cancel' is clicked" would read slightly nicer? (Note also that the button text should start with a capital letter to match actual button, since we're doing that for all the other button texts.)

mtaylorgds, line 124: Should we test that any changes made aren't saved when the cancel button is clicked?

test/integration/homepage_popular_links_test.rb
```diff
@@ -3,30 +3,165 @@
...
+    end
+  end
+
+  context "#publish" do
+    setup do
+      click_button("Create new edition")
+    end
+
+    should "publish the latest draft popular links" do
+      row = find_all(".govuk-summary-list__row")
```
mtaylorgds, line 137: Again, not sure we need to assert that start state.

test/integration/homepage_popular_links_test.rb
```diff
@@ -3,30 +3,165 @@
...
+    end
   end
 
-  should "have popular links version and status" do
-    assert page.has_text?("Edition")
-    assert page.has_text?(@popular_links.version_number)
-    assert page.has_text?("Status")
-    assert page.has_text?("DRAFT")
-    assert page.has_css?(".govuk-tag--yellow")
+  context "#publish" do
```
mtaylorgds, line 149: Unless I'm missing something, this is a duplicate of the preceding context block and test?

test/models/popular_links_edition_test.rb
```diff
@@ -11,32 +11,46 @@ class PopularLinksEditionTest < ActiveSupport::TestCase
...
     assert_not popular_links.valid?
-    assert_equal popular_links.errors[:item][0], "A Title is required for Link 1"
+    assert popular_links.errors.messages[:url2].include?("URL is invalid for Link 2, all URLs should have at least one '.' and no spaces.")
+    assert popular_links.errors.messages[:url1].include?("URL is required for Link 1")
+  end
+
+  should "create new record from last 'published' record with status as 'draft' and increased 'version_number'" do
+    popular_links = FactoryBot.create(:popular_links, state: "published")
+
+    assert_equal "published", PopularLinksEdition.last.state
```
mtaylorgds, line 48: Not sure we need these assertions in this test.

mtaylorgds (changes requested):
Good work!

Could use a few minor changes before merging, the biggest ones are around using the components from the publishing components gem wherever possible.

app/views/homepage/popular_links/_sidebar.html.erb
davidtrussler: I think these are simple links really to the edit page so we should mark them up as that and wouldn't need to do a POST request via a form. Unless I have missed something.

app/views/homepage/popular_links/_sidebar.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="sidebar-components_sidebar__border sidebar-component__button">
```
davidtrussler, line 1: I think this overcomplicating the CSS classes here - I would just give this element a class name of "sidebar-components". Then in the CSS we just need a rule to set the width of `.govuk-button` elements within that to 100%.

davidtrussler:
@anatron @syed-ali-tw Is there a design available for the new sidebar? I wonder as well if this is intended to be sticky rather than staying fixed at the top when the user scrolls down.

app/views/homepage/popular_links/_sidebar.html.erb
davidtrussler: As Mark has said elsewhere that shouldn't be a problem and we really should be using Publishing Components at all times.

app/views/homepage/popular_links/_form.html.erb
```diff
@@ -0,0 +1,18 @@
+<h2 class="govuk-heading-m">Link <%= index +1 %></h2>
+<%= render "govuk_publishing_components/components/input", {
```
davidtrussler, line 2: We need to include the "With error items" option of the [Input component](https://components.publishing.service.gov.uk/component-guide/input#with_error_items) here so that the error is flagged on the element that has an error as well as the top of the page.

app/helpers/errors_helper.rb
```diff
@@ -1,12 +1,18 @@
 module ErrorsHelper
-  def errors_for(errors, attribute)
+  def errors_for(errors, attribute, full_message: true)
```
mtaylorgds, line 2: Maybe name this parameter to be clearer that it's expecting a boolean (e.g. `use_full_message`)? Or maybe even clearer, have two public methods that just have the `errors` and `attribute` parameters (one for using `message` and one for using `full_message`), and a private method that is this one with the boolean parameter (if you can have private methods in a module…?).

anatron:
> @anatron @syed-ali-tw Is there a design available for the new sidebar? I wonder as well if this is intended to be sticky rather than staying fixed at the top when the user scrolls down.

Yes, it is intended to be sticky. This will be part of another PR.

app/views/homepage/popular_links/_sidebar.html.erb
davidtrussler: I'm still not sure why these are not rendered as links: I can't see the advantage of using a button in a form for this. I think the precedent in Whitehall would be to use a link, seems to be the equivalent of "Edit draft" on [this page](https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/detailed-guides/1575767).

mtaylorgds: I agree that the edit "button" could be a link, since it's a get request. The "Create new edition" is actually causing a data change on the backend (creating a new edition), so should be a POST request (and so needs a form with submit button).

mtaylorgds: (for the create button that would match [what Whitehall does](https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/publications/1571182)).

syed-ali-tw: We have made the button for the 'GET' request to edit page to be a `Link button` now.

mtaylorgds (approved):
Looks good from my perspective. Great job!

app/views/homepage/popular_links/_form.html.erb
```diff
@@ -0,0 +1,24 @@
+<%= render "govuk_publishing_components/components/heading", {
```
davidtrussler, line 1: I think we should use a [Form fieldset](https://components.publishing.service.gov.uk/component-guide/fieldset) component to group each of the link sections together, which would be helpful for assistive technology, rather than having a form full of individual inputs. Each of the headings in that case would be a `<legend>` element.

davidtrussler:
> > @anatron @syed-ali-tw Is there a design available for the new sidebar? I wonder as well if this is intended to be sticky rather than staying fixed at the top when the user scrolls down.
> 
> Yes, it is intended to be sticky. This will be part of another PR.

OK, though it should be a fairly simple CSS addition if you think it's worth doing it as part of this work.

davidtrussler (approved):
LGTM! 👍

PR #2228, opened 2024-07-03
Publishing popular links - make sidebar sticky

ellohez:

[Trello Card](https://trello.com/c/szYiEKlQ/298-enable-publishing-of-popular-links)

### DIFF
--- a/app/assets/stylesheets/sidebar_components.scss
+++ b/app/assets/stylesheets/sidebar_components.scss
@@ -1,6 +1,0 @@
-.sidebar-components {
-  .govuk-button {
-    width: 100%;
-  }
-  border-top: 2px solid $govuk-brand-colour;
-}
--- a/app/views/homepage/popular_links/show.html.erb
+++ b/app/views/homepage/popular_links/show.html.erb
@@ -7,6 +7,6 @@
     <%= render "homepage/popular_links/link", item:, index: %>
   <% end %>
 </div>
-<div class = "govuk-grid-column-one-third" >
+<div class = "govuk-grid-column-one-third popular-links__sidebar" >
   <%= render "homepage/popular_links/sidebar" %>
 </div>
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -4,4 +4,4 @@ $govuk-page-width: 1140px;
 @import "govuk_publishing_components/all_components";
 @import "downtimes";
 @import "summary-card";
-@import "sidebar_components";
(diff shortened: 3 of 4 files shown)


app/assets/stylesheets/sidebar_components.scss
```diff
@@ -4,3 +4,10 @@
   }
   border-top: 2px solid $govuk-brand-colour;
 }
+
+.sticky {
```
mtaylorgds, line 8: Rather than name this after what it does, maybe it would be better to name it as what it is (e.g. a sidebar). Could it even be directly applied to the `sidebar-components` class (or does that not work because that `div` is at a different level in the hierarchy)?

@davidtrussler might have a better view on this matter.

davidtrussler: Yes I think we should avoid a generic class name like this and also name it in the way Mark suggests above. I think that's not completely straightforward though since we're applying the style to the container of the thing we want to be sticky. We may need to rename elements a bit.

app/views/homepage/popular_links/show.html.erb
```diff
@@ -7,6 +7,6 @@
     <%= render "homepage/popular_links/link", item:, index: %>
   <% end %>
 </div>
-<div class = "govuk-grid-column-one-third" >
+<div class = "govuk-grid-column-one-third sticky" >
```
davidtrussler, line 10: Hmm, so I think this has sparked a naming convention debate (in my head at least). Probably the SCSS file should be called `popular-links`. Then my suggestion would be to use the name `popular-links__sidebar-components` (or something like it) for this element and maybe rename the child div of this to `sidebar-components__contents` (though strictly speaking it probably doesn't need a name at all). Then duplicate those names in the appropriate places in the Sass.

davidtrussler: Think I'd go for something like this: 
```
<div class = "govuk-grid-column-one-third popular-links__sidebar">
  <div class="sidebar__components">
```

davidtrussler:
@ellohez @anatron Sorry for the confusing messages about naming!

There's also a couple of other things you might want to check with Rebecca: 
- Would we want a bit of space at the top of the viewport rather than the sticky element being flush to the top? (I don't know if there is a convention for this elsewhere)
- What would be the expected behaviour on smaller viewports? (when the sidebar renders below the other content)

davidtrussler (approved):
If Rebecca is happy with this I'm happy to approve 👍

PR #2235, opened 2024-07-06
Publish popular links

syed-ali-tw:
This PR does the following:

- Calls publishing api to store the updated popular links content to draft content store when an edit and save action is performed on popular links.
- Calls publishing api to store the published popular links content to content store when a publish action is performed on popular links.

[trello](https://trello.com/c/szYiEKlQ/298-enable-publishing-of-popular-links)

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -16,8 +16,8 @@ class HomepageController < ApplicationController
   end
 
   def update
-    update_link_items
-    UpdateWorker.perform_async(@latest_popular_links.id.to_s, update_action_is_publish?)
+    @latest_popular_links.link_items = remove_leading_and_trailing_url_spaces(params[:popular_links].values)
+    @latest_popular_links.save_draft
 
     flash[:success] = "Popular links draft saved.".html_safe
     redirect_to show_popular_links_path
@@ -26,7 +26,8 @@ class HomepageController < ApplicationController
   end
 
   def publish
-    publish_latest_popular_links
+    @latest_popular_links.publish
+    flash[:success] = "Popular links successfully published.".html_safe
     render "homepage/popular_links/show"
   end
 
@@ -36,11 +37,6 @@ private
     @latest_popular_links = PopularLinksEdition.last
   end
 
-  def update_link_items
-    @latest_popular_links.link_items = remove_leading_and_trailing_url_spaces(params[:popular_links].values)
-    @latest_popular_links.save!
-  end
-
   def remove_leading_and_trailing_url_spaces(links)
     link_items = []
     links.each do |link|
@@ -48,17 +44,5 @@ private
       link_items << link
     end
     link_items
-  end
-
-  def publish_latest_popular_links
-    @latest_popular_links.publish_popular_links
-  end
-
-  def update_action_is_publish?
-    attempted_activity == :publish
-  end
-
-  def attempted_activity
-    Edition::ACTIONS.invert[params[:commit]]
   end
 end
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -17,6 +17,8 @@ class HomepageController < ApplicationController
 
   def update
     update_link_items
+    UpdateWorker.perform_async(@latest_popular_links.id.to_s, update_action_is_publish?)
+
     flash[:success] = "Popular links draft saved.".html_safe
     redirect_to show_popular_links_path
   rescue StandardError
@@ -50,5 +52,13 @@ private
 
   def publish_latest_popular_links
     @latest_popular_links.publish_popular_links
+  end
+
+  def update_action_is_publish?
+    attempted_activity == :publish
+  end
+
+  def attempted_activity
+    Edition::ACTIONS.invert[params[:commit]]
   end
 end
--- a/app/presenters/formats/popular_links_presenter.rb
+++ b/app/presenters/formats/popular_links_presenter.rb
@@ -1,0 +1,46 @@
+module Formats
+  class PopularLinksPresenter
+    # Unlike presenters for other models which inherit from editions model,
+    # PopularLinksPresenter does not inherit from EditionFormatPresenter
+    # This is because Edition presenter has dependency on Artefact and
+    # Popular links are not coupled with Artefact unlike any other models.
+    # Hence the difference
+    def initialize(popular_links_edition)
+      @popular_links_edition = popular_links_edition
+    end
+
+    def render_for_publishing_api(*)
+      required_fields.merge optional_fields
+    end
+
+  private
+
+    attr_reader :popular_links_edition
+
+    def required_fields
+      {
+        title: popular_links_edition.title,
+        schema_name: "link_collection",
+        document_type: "link_collection",
+        publishing_app: "publisher",
+        rendering_app: "frontend",
+        details:,
+      }
+    end
+
+    def details
+      {
+        link_items: popular_links_edition.link_items,
+      }
+    end
+
+    def optional_fields
+      access_limited = { auth_bypass_ids: [popular_links_edition.auth_bypass_id] }
+      { access_limited:, public_updated_at: public_updated_at.rfc3339(3) }
+    end
+
+    def public_updated_at
+      popular_links_edition.updated_at
+    end
+  end
+end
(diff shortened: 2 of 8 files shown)


app/presenters/formats/popular_links_presenter.rb
```diff
@@ -0,0 +1,47 @@
...
+    # Unlike presenter for other models which inherit from editions model,
+    # PopularLinksPresenter do not inherit from EditionFormatPresenter
+    # This is because Edition presenter has dependency on Artifact and
+    # Popular links are not coupled with Artefact unlike any other models.
+    # Hence the difference
+    def initialize(popular_links_edition)
+      @popular_links_edition = popular_links_edition
+    end
+
+    def render_for_publishing_api(*)
```
mtaylorgds, line 12: Interesting, I didn't know about this `(*)` syntax!

test/functional/homepage_controller_test.rb
```diff
@@ -58,6 +58,8 @@ class HomepageControllerTest < ActionController::TestCase
     end
 
     should "update latest PopularLinksEdition with changed title and url" do
+      UpdateService.stubs(:call)
```
mtaylorgds, line 61: Do we need this stubbing? Commenting it out doesn't seem to cause any ill-effect.

test/functional/homepage_controller_test.rb
```diff
@@ -76,8 +78,23 @@ class HomepageControllerTest < ActionController::TestCase
...
+                                 { "1" => { "title" => "title", "url" => "url.com" },
+                                   "2" => { "title" => "title2", "url" => "https://www.url2.com" },
+                                   "3" => { "title" => "title3", "url" => "https://www.url3.com" },
+                                   "4" => { "title" => "title4", "url" => "https://www.url4.com" },
+                                   "5" => { "title" => "title5", "url" => "https://www.url5.com" },
+                                   "6" => { "title" => "title6", "url" => "https://www.url6.com" } } }
+    end
+
     should "redirect to show path on success" do
+      UpdateService.stubs(:call)
```
mtaylorgds, line 95: Do we need this stubbing? Commenting it out doesn't seem to cause any ill-effect.

test/integration/homepage_popular_links_test.rb
```diff
@@ -78,6 +78,7 @@ class HomepagePopularLinksTest < JavascriptIntegrationTest
 
   context "#edit" do
     setup do
+      UpdateService.stubs(:call)
```
mtaylorgds, line 81: Do we need this stubbing? Commenting it out doesn't seem to cause any ill-effect.

test/integration/homepage_popular_links_test.rb
```diff
@@ -135,6 +136,20 @@ class HomepagePopularLinksTest < JavascriptIntegrationTest
     end
   end
 
+  context "#publish" do
+    setup do
+      Services.publishing_api.stubs(:publish)
```
mtaylorgds, line 141: Do we need this stubbing? Commenting it out doesn't seem to cause any ill-effect.

app/presenters/formats/popular_links_presenter.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def details
+      {
+        link_items: popular_links_edition.link_items,
+      }
+    end
+
+    def optional_fields
+      access_limited = { auth_bypass_ids: [popular_links_edition.auth_bypass_id] }
+      { access_limited:, public_updated_at: public_updated_at.rfc3339(3) }
```
mtaylorgds, line 40: `public_updated_at` is also in hash returned by `required_fields`—does it make sense to be in both? Is it required, or optional?

test/unit/presenters/formats/popular_links_format_presenter_test.rb
```diff
@@ -0,0 +1,33 @@
...
+  def subject
+    Formats::PopularLinksPresenter.new(FactoryBot.create(:popular_links))
+  end
+
+  context "#render_for_publishing_api" do
+    setup do
+      @result = subject.render_for_publishing_api
+    end
+
+    should "have expected required fields" do
```
mtaylorgds, line 13: Should we consider making the tests here a bit more consistent with other format presenter tests? They all seem to have [individual tests for fields](https://github.com/alphagov/publisher/blob/cd76211dda73c9bb2dbc96d3c5bb126413cd6f9d/test/unit/presenters/formats/edition_format_presenter_test.rb#L144), with names like "[:change_note]".

test/unit/presenters/formats/popular_links_format_presenter_test.rb
```diff
@@ -0,0 +1,33 @@
...
+                                   { url: "https://www.url2.com", title: "title2" },
+                                   { url: "https://www.url3.com", title: "title3" },
+                                   { url: "https://www.url4.com", title: "title4" },
+                                   { url: "https://www.url5.com", title: "title5" },
+                                   { url: "https://www.url6.com", title: "title6" }] }),
+                   @result[:details]
+    end
+
+    should "have expected optional fields" do
+      assert_not_empty @result[:access_limited]
```
mtaylorgds, line 29: Maybe we can test the actual `access_limited` value, [like how the edition format presenter test does](https://github.com/alphagov/publisher/blob/cd76211dda73c9bb2dbc96d3c5bb126413cd6f9d/test/unit/presenters/formats/edition_format_presenter_test.rb#L159).

app/controllers/homepage_controller.rb
```diff
@@ -16,15 +16,16 @@ def edit
...
+    @latest_popular_links.update_draft(update_link_items)
     flash[:success] = "Popular links draft saved.".html_safe
     redirect_to show_popular_links_path
   rescue StandardError
     render "homepage/popular_links/edit"
   end
 
   def publish
-    publish_latest_popular_links
+    @latest_popular_links.publish
```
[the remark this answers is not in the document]
syed-ali-tw: Yes, though we were thinking to make a small separate PR for this just for error message, though seems like we can have a separate commit and not requires a separate PR

app/controllers/homepage_controller.rb
```diff
@@ -16,15 +16,16 @@ def edit
   end
 
   def update
-    update_link_items
+    @latest_popular_links.update_draft(update_link_items)
```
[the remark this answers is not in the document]
syed-ali-tw: Yes, though we were thinking to make a small separate PR for this just for error message, though seems like we can have a separate commit and not requires a separate PR

app/presenters/formats/popular_links_presenter.rb
syed-ali-tw: good catch, In my head, I had moved it to optional fields.

app/controllers/homepage_controller.rb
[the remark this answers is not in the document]
mtaylorgds: Given that none of this is in real use yet, I think it's fine to have a separate PR if you like for the error handling.

test/unit/presenters/formats/popular_links_format_presenter_test.rb
syed-ali-tw: I did not notice that, I was bit iffy on it. I felt it was a bit of overkill to have separate tests for each field.I still feel we dont need to do that

test/functional/homepage_controller_test.rb
syed-ali-tw: this is interesting, this is working without stubbing because, we are stubbing it in test helper [here](https://github.com/alphagov/publisher/blob/cd76211dda73c9bb2dbc96d3c5bb126413cd6f9d/test/test_helper.rb#L52)
not sure, I like this

mtaylorgds: Interesting. I think doing that is OK, so long as we also have test coverage that we're sending the right data (which I think we do).

test/unit/presenters/formats/popular_links_format_presenter_test.rb
```diff
@@ -0,0 +1,33 @@
+require "test_helper"
+
+class PopularLinksFormatPresenterTest < ActiveSupport::TestCase
```
mtaylorgds, line 3: I think we might be missing a "be valid against schema" test (looking at the other format presenter tests).

mtaylorgds (commented):
One more comment about testing the generated payload against the schema.

mtaylorgds (approved):
LGTM

PR #2248, opened 2024-07-16
Restructure integration tests

mtaylorgds:
Prepare for integration test helpers that are useful for pages transitioned to the GOV.UK Design System.

The classes in the integration test helper file contain lots of methods tightly-coupled to the old, bootstrap UI, structure. Rename those classes to have a "legacy" prefix, and create new (for now, minimal) ones for use by transitioned pages (currently only the downtime and popular link editing pages have integration tests and are using the GOV.UK Design System).

By keeping the old and new helpers completely separate, it will make cleanup at the end (once all pages have been transitioned, and are using the new helper classes) a lot easier, as we can simply delete the legacy integration test helper file.

This PR also addresses a weird implementation quirk, where the `ActionDispatch::IntegrationTest` class was being redefined, rather than extended from. I can't see any reason why that would have been done, and so have corrected it to be implemented more normally.

[Trello card](https://trello.com/c/wucvIcYg)

### DIFF
--- a/test/integration_test_helper.rb
+++ b/test/integration_test_helper.rb
@@ -3,7 +3,7 @@ require "capybara/rails"
 require "capybara-select-2"
 require "support/govuk_test"
 
-class PublisherIntegrationTest < ActionDispatch::IntegrationTest
+class IntegrationTest < ActionDispatch::IntegrationTest
   include Capybara::DSL
   include CapybaraSelect2
   include CapybaraSelect2::Helpers
@@ -29,7 +29,7 @@ class PublisherIntegrationTest < ActionDispatch::IntegrationTest
   end
 end
 
-class JavascriptIntegrationTest < PublisherIntegrationTest
+class JavascriptIntegrationTest < IntegrationTest
   setup do
     Capybara.current_driver = Capybara.javascript_driver
   end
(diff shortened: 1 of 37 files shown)


syed-ali-tw:
(currently only the downtime pages have been transitioned and have integration tests).
not transitioned but homepage popular links use design system

test/integration_test_helper.rb
```diff
@@ -3,7 +3,7 @@
 require "capybara-select-2"
 require "support/govuk_test"
 
-class ActionDispatch::IntegrationTest
+class PublisherIntegrationTest < ActionDispatch::IntegrationTest
```
syed-ali-tw, line 6: I see why we cant just name the class as 'IntegrationTest' but I am not sure about 'PublisherIntegrationTest'. its more like a helper to me but inheriting from helper would also not make much sense. may be a bit more generic like 'AppIntegrationTest' ?

mtaylorgds:
> (currently only the downtime pages have been transitioned and have integration tests). not transitioned but homepage popular links use design system

I've reworded the sentence,

test/integration_test_helper.rb
mtaylorgds: We probably could rename this to be `IntegrationTest`, if desired (and the legacy one to `LegacyIntegrationTest`?

I don't really understand how `AppIntegrationTest` is any different to `PublisherIntegrationTest` (in meaning)?

syed-ali-tw (approved):

syed-ali-tw:
LGTM

PR #2250, opened 2024-07-17
Switch from sassc-rails to dartsass-rails

richardTowers:
This is an initial stab at tackling this piece of tech debt - https://trello.com/c/CCLlRFxl/90-migrate-publisher-to-dart-sass
[Mainstream team trello card](https://trello.com/c/rtkv3YMZ/348-migrate-mainstream-publisher-to-dart-sass-publisher)

Mostly, this follows the excellent documentation here: https://docs.publishing.service.gov.uk/manual/migrate-to-dart-sass-from-libsass.html

I had to add the legacy-application CSS / JS to the manifest file to stop the app from complaining when running it locally. I've now got it to the point where it runs, but I'm a bit concerned the CSS isn't perfect because the skip link appears at the top of the page:

image

Hopefully this is a good start, but I think it might need a bit of care and attention before it's ready to merge.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ coverage
 test/reports
 .generators
 .rakeTasks
+/app/assets/builds/*
 
 *.daemonette-dump
 data/local_interactions.csv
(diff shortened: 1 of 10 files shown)


richardTowers:
Hmm, I think this might be complicated a bit by the fact that `govuk_admin_template` introduces a dependency on `bootstrap-sass`, which itself depends on `sassc` - so even with out a direct dependency on it, we end up with sassc still in the mix 🤔

mtaylorgds:
Closing as it has been superseded by #2278

PR #2259, opened 2024-07-24
Log a warning on DocumentNotFound

mtaylorgds:
Log a warning when a `Mongoid::Errors::DocumentNotFound` error is raised in a controller. These errors were being silently swallowed by the application.

The error occurs when an attempt is made to load a document that doesn't exist, which can happen when the overnight DB replication from prod->integration fails. It will also happen if the edition ID is manipulated in the URL (e.g. in URLs of the form
 `/editions/[edition-id]`), so it's not _necessarily_ an error when it happens.

### Example log

image

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -19,7 +19,8 @@ class ApplicationController < ActionController::Base
     prepend_view_path template_folder_for(publication)
   end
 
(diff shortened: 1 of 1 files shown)


baisa (approved):
Super useful, thank you @mtaylorgds 👌

PR #2262, opened 2024-07-25
Handle publishing api errors and save content before publishing

syed-ali-tw:
This commit tries to bring following changes
1. Handle publishing api errors and give user more understandable messages where possible for publishing api errors.
2. Handle database errors.

[Trello](https://trello.com/c/sw3DmXgK/363-update-error-messaging-copy)

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -21,20 +21,54 @@ class HomepageController < ApplicationController
 
     flash[:success] = "Popular links draft saved.".html_safe
     redirect_to show_popular_links_path
-  rescue StandardError
+  rescue GdsApi::HTTPErrorResponse
+    flash[:danger] = publishing_api_save_error_message.html_safe
+    render "homepage/popular_links/edit"
+  rescue Mongoid::Errors::Validations
+    render "homepage/popular_links/edit"
+  rescue StandardError => e
+    Rails.logger.error "Error #{e.class} #{e.message}"
+    flash[:danger] = "Due to an application error, the edition couldn't be saved."
     render "homepage/popular_links/edit"
   end
 
   def publish
-    @latest_popular_links.publish
+    @latest_popular_links.publish_latest
+
     flash[:success] = "Popular links successfully published.".html_safe
-    render "homepage/popular_links/show"
+  rescue GdsApi::HTTPErrorResponse => e
+    flash[:danger] = rescue_already_published_error(e)
+  rescue StandardError => e
+    Rails.logger.error "Error #{e.class} #{e.message}"
+    flash[:danger] = "Due to an application error, the edition couldn't be published."
+  ensure
+    redirect_to show_popular_links_path
   end
 
 private
 
+  def rescue_already_published_error(error)
+    already_published_error?(JSON.parse(error.http_body)) ? "Popular links publish was unsuccessful, cannot publish an already published content item.".html_safe : publishing_api_publish_error_message
+  end
+
+  def already_published_error?(error_body)
+    error_body["error"] && error_body["error"]["message"] && error_body["error"]["message"].include?("already published content item")
+  end
+
   def fetch_latest_popular_link
     @latest_popular_links = PopularLinksEdition.last
+  end
+
+  def publishing_api_publish_error_message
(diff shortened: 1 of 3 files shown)


app/controllers/homepage_controller.rb
```diff
@@ -21,22 +21,56 @@ def update
...
+    redirect_to show_popular_links_path
   end
 
 private
 
+  def rescue_already_published_error(error)
+    already_published_error(JSON.parse(error.http_body)) ? "Popular links publish was unsuccessful, cannot publish an already published content item.".html_safe : publishing_api_publish_error_message
+  end
+
+  def already_published_error(error_body)
```
mtaylorgds, line 54: Given this method returns a boolean, should we append a `?` to the method name?

app/controllers/homepage_controller.rb
```diff
@@ -21,22 +21,56 @@ def update
...
 
+  def publishing_api_publish_error_message
+    "Popular links publish was unsuccessful due to a service problem. #{try_again}".html_safe
+  end
+
+  def publishing_api_save_error_message
+    "Popular links save was unsuccessful due to a service problem. #{try_again}".html_safe
+  end
+
+  def try_again
```
mtaylorgds, line 70: Maybe we should rename this method (or extract the text to a constant). The current name would suggest to me that it is going to try doing something again, not that it returns a message.

app/models/popular_links_edition.rb
```diff
@@ -34,7 +34,8 @@ def create_draft_popular_links_from_last_record
     popular_links
   end
 
-  def publish
+  def publish_latest
+    save_draft
```
mtaylorgds, line 38: This change doesn't seem related to the PR text, and there's no explanation in the commit about why we're doing this—would be good to have the reasoning explained in the commit text (and it should probably be a separate commit).

test/functional/homepage_controller_test.rb
```diff
@@ -104,13 +104,84 @@ class HomepageControllerTest < ActionController::TestCase
...
+                                     "6" => { "title" => "title6", "url" => "https://www.url6.com" } } }
+
+        assert_template "homepage/popular_links/edit"
+      end
+    end
+
+    context "publishing api errors" do
+      should "render edit template on publishing api error" do
+        new_title = "title has changed"
+        stub_publishing_api_put_content_error
```
mtaylorgds, line 155: Since we've got a setup method in the previous context block (for database errors), we should probably be consistent and do the same here (for stubbing the publishing api).

test/integration/homepage_popular_links_test.rb
```diff
@@ -133,6 +133,18 @@ class HomepagePopularLinksTest < JavascriptIntegrationTest
 
       assert_not page.has_text?("www.abc.com")
     end
+
+    should "display 'service problem' error message when publishing api gives an error" do
```
mtaylorgds, line 137: Do we need the extra tests in here? Don't they just duplicate the ones in `test/functional/homepage_controller_test.rb`? I think the integration tests are slower, so we should try and not have too many of them.

If we do want to keep them, we should probably extract the stubbing to methods like is done for the 
tests in `test/functional/homepage_controller_test.rb`.

mtaylorgds (commented):
Added some comments.

Also, can we try and be a bit more specific with our branch names in future?

test/integration/homepage_popular_links_test.rb
syed-ali-tw: Yeah, we can get rid of these now.

test/functional/homepage_controller_test.rb
syed-ali-tw: I did think about it but there are different stubbing for publishing api inside this context block and hence chose not to use setup block

syed-ali-tw:
> Added some comments.
> 
> Also, can we try and be a bit more specific with our branch names in future?

Yes, this branch originally just intended to fix a bug that occurred because we had no rescue for publishing api in publish block but then we realised that error message was part of it and it was too late or not worth it to change branch name.

mtaylorgds (approved):
LGTM!

PR #2284, opened 2024-08-15
321 upgrade publications page table

davidtrussler:
[Trello](https://trello.com/c/lSVHry7Z/321-upgrade-publications-page-table)

These changes add the new table, using the Design System, to the Publications page. This is only rendered with the Design System flag. There is still work to do on this featrure that will be covered in later PRs:
- complete the tests for the new table
- update the layout to deal with some problems that became apparent once it is in place

|Current|Design System|
|-|-|
|Screenshot 2024-08-15 at 13 13 22|Screenshot 2024-08-15 at 13 13 54|


mtaylorgds (commented):
Added a few comments.

test/unit/helpers/admin/publications_table_helper_test.rb
```diff
@@ -0,0 +1,185 @@
...
+        assigned_to_id: "66911dbf2c88ee0001d8af62",
+        reviewer: nil,
+      )
+
+      assert_includes reviewer(edition, current_user), '<button class="gem-c-button govuk-button" type="submit">Claim 2i</button>'
+    end
+  end
+
+  context "#format" do
+    should "return the correct value for the format" do
```
mtaylorgds, line 159: I wonder whether we should be testing all the content types here…?

test/unit/helpers/admin/publications_table_helper_test.rb
```diff
@@ -0,0 +1,185 @@
...
+      )
+
+      edition_number_text = edition_number(archived_edition)
+
+      assert_match "2 - ", edition_number_text
+      assert_match "#3 in in review", edition_number_text
+    end
+  end
+
+  context "#important_note" do
```
mtaylorgds, line 63: I think we're missing a case for when the edition doesn't have an important note.

test/unit/helpers/admin/publications_table_helper_test.rb
```diff
@@ -0,0 +1,185 @@
...
+    should "return the important note text for an edition" do
+      user = FactoryBot.create(:user, name: "Keir Starmer")
+      edition = FactoryBot.create(:edition)
+      user.record_note(edition, "This is an important note", Action::IMPORTANT_NOTE)
+
+      assert_equal "This is an important note", important_note(edition)
+    end
+  end
+
+  context "#awaiting_review" do
```
mtaylorgds, line 73: I think we're missing a case for when the edition is not "in review".

test/unit/helpers/admin/publications_table_helper_test.rb
```diff
@@ -0,0 +1,185 @@
...
+    should "return the correct text for the 'Awaiting review' field for an edition in review" do
```
mtaylorgds, line 74: We can probably simplify this test name slightly. For example:
```suggestion
    should "return the correct text for an edition in review" do
```

test/unit/helpers/admin/publications_table_helper_test.rb
```diff
@@ -0,0 +1,185 @@
...
+    should "return the correct value for the format" do
+      edition = FactoryBot.create(
+        :transaction_edition,
+      )
+
+      assert_equal "Transaction", format(edition)
+    end
+  end
+
+  # TODO: How to stub value for "last_fact_checked_at"?
```
mtaylorgds, line 168: We should sort out this test before merging.

test/unit/helpers/admin/publications_table_helper_test.rb
```diff
@@ -0,0 +1,185 @@
...
+          state: "in_review",
+          review_requested_at: "2024-07-12",
+        )
+
+        assert_equal "21 days", awaiting_review(edition)
+      end
+    end
+  end
+
+  context "#scheduled" do
```
mtaylorgds, line 89: I think we're missing a case for when the edition is not scheduled for publication.

test/unit/helpers/admin/publications_table_helper_test.rb
```diff
@@ -0,0 +1,185 @@
...
+    should "return the correct text for the 'Scheduled' field for an edition in 'Scheduled for publishing'" do
```
mtaylorgds, line 90: I think we can slightly simplify the test name. For example:
```suggestion
    should "return the correct text for an edition that is scheduled for publishing" do
```

test/unit/helpers/admin/publications_table_helper_test.rb
```diff
@@ -0,0 +1,185 @@
...
+          publish_at: "2024-08-21 10:04:50.119 UTC",
+        )
+
+        assert_equal "11:04am, 21 Aug 2024", scheduled(edition)
+      end
+    end
+  end
+
+  context "#published_by" do
+    should "return the correct text for the 'Published by' field for an edition in 'Published'" do
```
mtaylorgds, line 106: I think we can slightly simplify the test name. For example:
```suggestion
    should "return the correct text for a published edition" do
```

test/unit/helpers/admin/publications_table_helper_test.rb
```diff
@@ -0,0 +1,185 @@
...
+          state: "scheduled_for_publishing",
+          publish_at: "2024-08-21 10:04:50.119 UTC",
+        )
+
+        assert_equal "11:04am, 21 Aug 2024", scheduled(edition)
+      end
+    end
+  end
+
+  context "#published_by" do
```
mtaylorgds, line 105: I think we're missing a case for when the edition is not scheduled for publication.

app/helpers/publications_table_helper.rb
```diff
@@ -0,0 +1,51 @@
+module PublicationsTableHelper
+  # test written
```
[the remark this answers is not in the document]
davidtrussler: Yes, those were just comments to myself - didn't mean to leave them in. Now removed.

spec/javascripts/spec/publisher_table.spec.js
```diff
@@ -0,0 +1,157 @@
...
+    var expandContractLink, detailsSections, event
+
+    beforeEach(function () {
+      expandContractLink = module.querySelector('p.publisher-table__heading a')
+      detailsSections = module.querySelectorAll('details')
+      event = new Event('click')
+    })
+
+    it('should be labelled "Expand all" when all of the "More details" sections are closed', function () {
+      detailsSections[0].open = false
```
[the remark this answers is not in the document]
davidtrussler: Yes, I agree with this in principle, but in this case the behaviour of the "More details" sections is contained in the Publishing Components gem and is not part of this module. In a way it highlights (again) how the whole concept of this table is a bit of a mash-up and not entirely satisfactory (at least not to me).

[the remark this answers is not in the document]
mtaylorgds: Would it not still be possible to click on the "more details" link for each section though (instead of setting `detailsSections[x].open = false`)?

app/assets/javascripts/modules/publisher-table.js
```diff
@@ -0,0 +1,48 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function PublisherTable ($module) {
```
mtaylorgds, line 5: I wonder whether we should rename this as "PublisherTable" seems a bit generic…maybe "PublicationsTable"?

spec/javascripts/spec/publisher_table.spec.js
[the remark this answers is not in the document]
davidtrussler: I've been looking at this and haven't been able to get that to work in the test. From our discussion earlier I think we can leave this at its for the minute and maybe pick it up as some technical debt later. There may be some other improvements to be made now that I've looked into it a bit more anyway.

spec/javascripts/spec/publications_table.spec.js
```diff
@@ -0,0 +1,171 @@
...
+    })
+  })
+
+  describe('The Expand/Contract all link', function () {
+    var expandContractLink, detailsSections, clickEvent
+
+    beforeEach(function () {
+      expandContractLink = module.querySelector('p.publications-table__heading a')
+      detailsSections = module.querySelectorAll('details')
+      // detailsSections = document.getElementsByClassName('details')
```
mtaylorgds, line 96: I think we can remove this commented-out line.

test/unit/helpers/admin/publications_table_helper_test.rb
```diff
@@ -0,0 +1,259 @@
...
+        )
+
+        assert_nil scheduled(edition_published)
+        assert_equal "11:04am, 21 Aug 2024", scheduled(edition_scheduled_for_publishing)
+      end
+    end
+  end
+
+  context "#published_by" do
+    should "return the correct text for a ;published edition" do
```
mtaylorgds, line 120: There seems to be a rogue semicolon in the test name

mtaylorgds (changes requested):
I think we're going to need to implement pagination before merging this PR—I tried it locally, with a copy of integration data, and it took over 9m for the app to return a response (and the browser gave up waiting after about 3m).

mtaylorgds (approved):
LGTM

PR #2291, opened 2024-08-19
Add homepage editor permission for all controller actions

syed-ali-tw:
In order to restrict access to homepage popular links, we have added homepage editor permission. 

**Why inside Mainstream?**
- Homepage popular links is going to be managed inside Mainstream and hence its related permission.

**Considerations and side-effects:**
Homepage appears as a separate app on sign-on though technically its just a separate controller accessed via Mainstream, this has a side-effect to understand that homepage permissions are to be access via Mainstream. This is something we understand as a team and have gone ahead with.
We are trying to mitigate the degree of effects through our messaging on signon and via access error message. We shall be iterating on this based on any feedback.

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -1,6 +1,9 @@
 class HomepageController < ApplicationController
   layout "design_system"
   before_action :fetch_latest_popular_link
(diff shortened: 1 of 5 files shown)


baisa (approved):
LGTM!

PR #2299, opened 2024-08-23
Add pagination to publication page

mtaylorgds:
Add pagination controls underneath the publications table on the root path of Mainstream Publisher.

image

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -20,6 +20,7 @@ $govuk-page-width: 1140px;
 @import 'govuk_publishing_components/components/layout-header';
 @import 'govuk_publishing_components/components/lead-paragraph';
 @import 'govuk_publishing_components/components/notice';
+@import 'govuk_publishing_components/components/previous-and-next-navigation';
 @import 'govuk_publishing_components/components/search';
 @import 'govuk_publishing_components/components/skip-link';
 @import 'govuk_publishing_components/components/success-alert';
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -27,12 +27,13 @@ class RootController < ApplicationController
       assigned_to_filter: assignee_filter,
       format_filter:,
       title_filter:,
+      page: filter_params_hash[:page],
     )
   end
 
 private
 
   def filter_params
-    params.permit(:assignee_filter, :format_filter, :title_filter, states_filter: [])
+    params.permit(:page, :assignee_filter, :format_filter, :title_filter, states_filter: [])
   end
 end
--- a/app/helpers/pagination_helper.rb
+++ b/app/helpers/pagination_helper.rb
@@ -1,0 +1,158 @@
+module PaginationHelper
+  class << self
+    LARGE_NUMBER_OF_PAGES = 6
+
+    def pagination_hash(current_page:, total_pages:, path:)
+      return if total_pages == 1
+
+      self.current_page = current_page
+      self.total_pages = total_pages
+      self.path = normalise_path(path, current_page)
+
+      previous_page = current_page - 1 if current_page >= 2
+      next_page = current_page + 1 if current_page < total_pages
+
+      previous_page_href = build_path_for(current_page - 1)
+      self.previous_href = previous_page.present? ? previous_page_href : nil
+
+      next_page_href = build_path_for(current_page + 1) if next_page.present?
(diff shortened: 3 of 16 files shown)


app/assets/stylesheets/publications-table.scss
```diff
@@ -80,3 +80,9 @@
     }
   }
 }
+
+.govuk-pagination {
+  .govuk-pagination__item.govuk-pagination__item--current {
+    color: #ffffff;
+  }
+}
```
davidtrussler, line 88: We should only really need the `.govuk-pagination__item--current` class name here: given the naming system it doesn't need to be nested in or extending any other class. 

Also we should use the `govuk-colour("white")` variable to set the colour.

davidtrussler:
I've just made one small comment in the CSS. Otherwise it all looks good. 

I'm still a bit concerned about implementing a pagination that is slightly different to Whitehall in behaviour and appearance and feel we really ought to be extending that component for use here. But maybe that's something for the future if necessary.

mtaylorgds:
> I've just made one small comment in the CSS. Otherwise it all looks good.

I've pushed the change you suggested.

> I'm still a bit concerned about implementing a pagination that is slightly different to Whitehall in behaviour and appearance and feel we really ought to be extending that component for use here. But maybe that's something for the future if necessary.

It's not a component per se in Whitehall, it still uses the same external library (Kaminari) for pagination, it just uses a helper to generate a "pagination_hash". I don't really understand the Whitehall code, which is partially why I'm reticent to reuse it here—I don't understand what value it provides over what I've done here, which is simple styling changes.

app/assets/stylesheets/publications-table.scss
davidtrussler: Think I'd literally just use 
```
.govuk-pagination__item--current {
  color: govuk-colour("white");
}
```
Should be specific enough since `--current` is extending `govuk-pagination__item`

davidtrussler:
> > I've just made one small comment in the CSS. Otherwise it all looks good.
> 
> I've pushed the change you suggested.
> 
> > I'm still a bit concerned about implementing a pagination that is slightly different to Whitehall in behaviour and appearance and feel we really ought to be extending that component for use here. But maybe that's something for the future if necessary.
> 
> It's not a component per se in Whitehall, it still uses the same external library (Kaminari) for pagination, it just uses a helper to generate a "pagination_hash". I don't really understand the Whitehall code, which is partially why I'm reticent to reuse it here—I don't understand what value it provides over what I've done here, which is simple styling changes.

Yes, I was just looking at it as well. Given there isn't a design for pagination I agree with your point here. 👍

app/assets/stylesheets/publications-table.scss
mtaylorgds: Good point, I've made that change now.

mtaylorgds:
@davidtrussler Have updated some integration tests to turn the publications page toggle off if they navigate to that page, as some of the tests were failing depending on what order the tests were being run in.

davidtrussler:
I wonder if we should check this from a Design POV before merging. I realise it's not styled exactly as Whitehall but it feels a little less elegant than that pagination on smaller viewports. It's still perfectly functional though so could be a future iteration if required.  
|Publisher|Whitehall|
|-|-|
|Screenshot 2024-08-28 at 12 58 30|Screenshot 2024-08-28 at 12 58 59|

app/views/kaminari/_first_page.html.erb
```diff
@@ -0,0 +1,11 @@
+<%# Link to the "First" page
+  - available local variables
+    url:           url to the first page
+    current_page:  a page object for the currently displayed page
+    total_pages:   total number of pages
+    per_page:      number of items to fetch per page
+    remote:        data-remote
+-%>
+<span class="govuk-pagination__prev">
```
davidtrussler, line 9: I think using `<span>` elements for each of the items here is slightly problematic. If possible these should be `<li>` elements. 

Also at smaller sizes the `govuk-pagination__prev` and `govuk-pagination__next` classes add some slightly awkward padding when used for the first and last items as well.

mtaylorgds: I've copied the implementation from Whitehall, so this now works identically to that and uses `li` elements.

davidtrussler (approved):
Looks great! 👍 
Also noted the ticket for creating the new component. 🙂

PR #2301, opened 2024-08-27
205 filter publications UI

davidtrussler:
[Trello](https://trello.com/c/wucvIcYg/205-upgrade-publications-page-search-filter)

This work primarily adds the UI to the filter section of the publications page into the left hand column (see screenshot below).

In addition to that there is some logic to pre-populate fields where required which could probably do with a bit more discussion form a Rails perspective.

Screenshot 2024-08-27 at 12 21 59


app/views/root/_filter.html.erb
```diff
@@ -0,0 +1,81 @@
...
+
+  @presenter.available_users.map do | user |
+    if user.id.to_s == @presenter.assigned_to
+      assignees << {text: user.name, value: user.id, selected: "true"}
+    else
+      assignees << {text: user.name, value: user.id}
+    end
+  end
+
+  format_filter_selection_options.map do | format |
```
mtaylorgds, line 14: The `format_filter_selection_options` method is only called here, so we could move this functionality to that method to avoid doing two loops (obviously we would need to also add a `content_type` parameter, and pass that in here). We should probably also rename the method to `content_type_filter_selection_options`.

We could even update the `FilteredEditionsPresenter` to just return the list that we want in the first place.

If we move all these loops to the `FilteredEditionsPresenter`, then we also avoid the need to add some of those extra methods that have been added in this PR.

app/views/root/_filter.html.erb
```diff
@@ -0,0 +1,81 @@
+<%
+  assignees = [{text: "All assignees", value: ""}]
+  content_types = []
+  states = []
+
+  @presenter.available_users.map do | user |
```
mtaylorgds, line 6: We should probably be consistent with what we do for the content type filter, and either use a helper method, or just move this code to the `FilteredEditionsPresenter`, since the `available_users` method is only called in this one place and I don't think the presenter being coupled to the view like that would be a problem (it would also make it easy to unit-test).

app/views/root/_filter.html.erb
```diff
@@ -0,0 +1,81 @@
...
+
+  format_filter_selection_options.map do | format |
+    if format[1] == @presenter.content_type
+      content_types << {text: format[0], value: format[1], selected: "true"}
+    else
+      content_types << {text: format[0], value: format[1]}
+    end
+  end
+
+  state_names.collect do |scope, status_label|
```
mtaylorgds, line 22: I'd be tempted to move this logic to the `FilteredEditionsPresenter`.

mtaylorgds: We should probably also be consistent in whether we use `map` or `collect` (they appear to be aliases for the same method as far as I can tell).

app/views/root/_filter.html.erb
```diff
@@ -0,0 +1,81 @@
...
+
+    <%= render "govuk_publishing_components/components/select", {
+      id: "assignee_filter",
+      full_width: true,
+      label: "Assigned to",
+      options: assignees,
+    } %>
+
+    <%= render "govuk_publishing_components/components/select", {
+      id: "format_filter",
```
mtaylorgds, line 56: Maybe now would be a good time to rename the various usages of the term "format" with "content type" in the new (non-legacy) code.

mtaylorgds (commented):
It doesn't necessarily have to be part of this PR, but we should at least do it as part of the Trello ticket to create a non-legaqy version of the integration tests in `test/integration/legacy_root_overview_test.rb`

app/presenters/filtered_editions_presenter.rb
```diff
@@ -3,18 +3,64 @@
...
   end
 
+  def title
+    @title_filter
+  end
+
+  def content_types
+    types = []
+
+    content_type_filter_selection_options.map do |format|
```
mtaylorgds, line 21: We should probably rename `format` to `content_type` for consistency with the other changes.

app/presenters/filtered_editions_presenter.rb
```diff
@@ -26,10 +72,33 @@ def editions
...
+  def content_type_filter_selection_options
+    [%w[All all]] +
+      Artefact::FORMATS_BY_DEFAULT_OWNING_APP["publisher"].map do |format_name|
+        displayed_format_name = format_name.humanize
+        displayed_format_name += " (Retired)" if Artefact::RETIRED_FORMATS.include?(format_name)
+        [displayed_format_name, format_name]
+      end
+  end
+
   def editions_by_format
```
mtaylorgds, line 98: We should probably rename this to `editions_by_content_type` for consistency.

mtaylorgds (commented):
A couple of "format" references left that we should probably rename to "content_type"; other than that, LGTM—but someone else will have to approve, since I pushed one of the commits on this PR.

syed-ali-tw (approved):

PR #2311, opened 2024-09-05
Improve width of publication table

ellohez:
Fix issue of more details -> summary being unreadable Summary details now span width of table when displayed
[Increase width of 'more details' section of the Publications table](https://trello.com/c/6jf5jino)

Implementation of David's fix for the width of the expanded details summary section. 
This is needed for user testing of the new default publications page. The team has agreed that more work maybe necessary on this in the future but that this fix should be released at this time, so as not to hold up user testing.


davidtrussler:
Hi @ellohez - there's just a couple of linting issues (blank lines either should or shouldn't be there) being complained about (see [here](https://github.com/alphagov/publisher/actions/runs/10722307262/job/29732928424?pr=2311)), otherwise all looks good to me.

PR #2315, opened 2024-09-06
205 filter publications UI root overview tests

davidtrussler:
[Trello](https://trello.com/c/wucvIcYg/205-upgrade-publications-page-search-filter)

The changes in this PR add a number of tests from the legacy Publications page and refactors them for the new Design System UI. The tests are based on the legacy tests in `/test/integration/legacy_root_overview_test.rb`.

Some legacy tests have not been ported to the Design System as they raise a number of issues about changed or missing functionality that will be dealt with separately. Details of these are in the Trello card.

### DIFF
--- a/test/integration/root_overview_test.rb
+++ b/test/integration/root_overview_test.rb
@@ -25,4 +25,282 @@ class RootOverviewTest < IntegrationTest
     assert_content("21 document(s)")
     assert_no_content("Guides and Gals")
   end
+
+  should "filter by assigned user" do
+    alice = FactoryBot.create(:user, :govuk_editor, name: "Alice", uid: "alice")
+    bob     = FactoryBot.create(:user, :govuk_editor, name: "Bob", uid: "bob")
+    charlie = FactoryBot.create(:user, :govuk_editor, name: "Charlie", uid: "charlie")
+
+    guide_edition_x = FactoryBot.create(:guide_edition, title: "XXX", slug: "xxx")
+    guide_edition_y = FactoryBot.create(:guide_edition, title: "YYY", slug: "yyy")
+    FactoryBot.create(:guide_edition, title: "ZZZ", slug: "zzz")
+
+    bob.assign(guide_edition_x, alice)
+    bob.assign(guide_edition_y, charlie)
+
+    visit "/"
+
+    filter_by_user("All")
+
+    assert page.has_content?("XXX")
+    assert page.has_content?("YYY")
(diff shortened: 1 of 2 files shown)


test/integration/root_overview_test.rb
```diff
@@ -25,4 +25,282 @@ class RootOverviewTest < IntegrationTest
...
     assert_no_content("Guides and Gals")
   end
+
+  should "filter by assigned user" do
+    alice = FactoryBot.create(:user, :govuk_editor, name: "Alice", uid: "alice")
+    bob     = FactoryBot.create(:user, :govuk_editor, name: "Bob", uid: "bob")
+    charlie = FactoryBot.create(:user, :govuk_editor, name: "Charlie", uid: "charlie")
+
+    x = FactoryBot.create(:guide_edition, title: "XXX", slug: "xxx")
+    y = FactoryBot.create(:guide_edition, title: "YYY", slug: "yyy")
```
syed-ali-tw, line 35: may be we can name it bit better, even something like guide_edition_1, 2

test/integration/root_overview_test.rb
```diff
@@ -25,4 +25,282 @@ class RootOverviewTest < IntegrationTest
...
+
+    within(".publications-table tr:first-child details .govuk-summary-list__row:nth-child(4) .govuk-summary-list__value") do
+      find_button("Claim 2i").click
+    end
+
+    assert edition_url(edition), current_url
+    assert page.has_content?("Another McPerson has already claimed this 2i")
+    assert page.has_select?("Reviewer", selected: another_user.name)
+    assert page.has_select?("Assigned to", selected: assignee.name)
+
```
syed-ali-tw, line 225: once we have checked that one cant claim 2i when someone else has, I dont know if there is a need to test the remaining of the test? or may be if we want to do that we can express the intention in the test.

test/integration/root_overview_test.rb
davidtrussler: Makes sense, I've renamed them keeping the 'x' and 'y' - 'guide_edition_x', 'guide_edition_y'.

test/integration/root_overview_test.rb
davidtrussler: Yes, good point. I've kept the test as is but updated the description: I think it is a useful thing to test for still.

syed-ali-tw:
LGTM

syed-ali-tw (approved):

PR #2316, opened 2024-09-06
Default the assignee filter on the publications page

mtaylorgds:
Default the assignee filter on the publications page

When visiting the root page, with no filters set, default the assignee
 filter so that only editions assigned to the current user are shown.

Order of precedence is:
1. assignee parameter
2. assignee in session
3. current user

[Trello card](https://trello.com/c/TVkDk1Oe)

### DIFF
--- a/test/unit/presenters/filtered_editions_presenter_test.rb
+++ b/test/unit/presenters/filtered_editions_presenter_test.rb
@@ -176,24 +176,49 @@ class FilteredEditionsPresenterTest < ActiveSupport::TestCase
     end
   end
 
-  context "#available_users" do
+  context "#assignees" do
+    should "return an 'all assignees' item" do
+      assignees = FilteredEditionsPresenter.new.assignees
+
+      assert_equal([{ text: "All assignees", value: "" }], assignees)
+    end
+
+    should "return users who are not the assignee as unselected" do
+      anna = FactoryBot.create(:user, name: "Anna")
+      assignees = FilteredEditionsPresenter.new.assignees
+
+      assert_includes(assignees, { text: anna.name, value: anna.id })
+    end
+
+    should "return the assignee as selected" do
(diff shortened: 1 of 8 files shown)


test/unit/presenters/filtered_editions_presenter_test.rb
```diff
@@ -176,24 +176,48 @@ class FilteredEditionsPresenterTest < ActiveSupport::TestCase
...
+
+      assert_equal([{ text: "All assignees", value: "" }], assignees)
+    end
+
+    should "return users who are not the assignee as unselected" do
+      anna = FactoryBot.create(:user, name: "Anna")
+      assignees = FilteredEditionsPresenter.new.assignees
+
+      assert_includes(assignees, { text: anna.name, value: anna.id })
+    end
```
baisa, line 191: should we add selected: "false" here?

test/unit/presenters/filtered_editions_presenter_test.rb
```diff
@@ -176,24 +176,48 @@ class FilteredEditionsPresenterTest < ActiveSupport::TestCase
...
       FactoryBot.create(:user, name: "disabled user", disabled: true)
 
-      users = FilteredEditionsPresenter.new.available_users.to_a
+      users = FilteredEditionsPresenter.new.assignees
 
-      assert_equal(users, [enabled_user])
+      assert_equal(2, users.count)
+      assert_equal("All assignees", users[0][:text])
+      assert_equal(enabled_user.name, users[1][:text])
     end
```
baisa, line 221: we should test more specifically that disabled user is not included, for example with `assert_not_includes users.map { |user| user[:text] }, disabled_user.name`

syed-ali-tw: makes sense

baisa (approved):

PR #2338, opened 2024-09-23
425 metadata tab edit page

syed-ali-tw:
[Trello](https://trello.com/c/N4sWF2NI/425-metadata-tab-edit-page)

Move Metadata tab to new design system

Screenshot_metadata_tab

### DIFF
--- a/app/helpers/tabbed_nav_helper.rb
+++ b/app/helpers/tabbed_nav_helper.rb
@@ -1,9 +1,8 @@
 module TabbedNavHelper
   def edition_nav_items(edition)
     nav_items = []
-    items = %w[edit tagging metadata history admin related_external_links unpublish]
 
-    items.each do |item|
+    all_tab_names.each do |item|
       nav_items << standard_nav_items(item, edition)
     end
 
@@ -28,5 +27,16 @@ module TabbedNavHelper
         current:,
       },
     ]
+  end
+
+  def current_tab_name
+    current_tab = (request.path.split("/") & all_tab_names).first
+    current_tab == "metadata" ? "metadata" : "temp_nav_text"
+  end
+
+private
+
+  def all_tab_names
+    %w[edit tagging metadata history admin related_external_links unpublish]
   end
 end
--- a/test/integration/edition_edit_test.rb
+++ b/test/integration/edition_edit_test.rb
@@ -6,58 +6,92 @@ class EditionEditTest < IntegrationTest
     test_strategy = Flipflop::FeatureSet.current.test!
     test_strategy.switch!(:design_system_edit, true)
     stub_linkables
-    edition = FactoryBot.create(:guide_edition, title: "Edit page title", state: "draft")
-    visit edition_path(edition)
   end
 
-  should "show document summary and title" do
-    assert page.has_title?("Edit page title")
-
-    row = find_all(".govuk-summary-list__row")
-    assert row[0].has_content?("Assigned to")
-    assert row[1].has_text?("Content type")
-    assert row[1].has_text?("Guide")
-    assert row[2].has_text?("Edition")
-    assert row[2].has_text?("1")
-    assert row[2].has_text?("Draft")
-  end
-
-  should "show all the tabs for the edit" do
-    assert page.has_text?("Edit")
-    assert page.has_text?("Tagging")
-    assert page.has_text?("Metadata")
-    assert page.has_text?("History and notes")
-    assert page.has_text?("Admin")
-    assert page.has_text?("Related external links")
-    assert page.has_text?("Unpublish")
-  end
-
-  context "#metadata" do
+  context "when edition is draft" do
     setup do
-      click_link("Metadata")
+      edition = FactoryBot.create(:guide_edition, title: "Edit page title", state: "draft")
+      visit edition_path(edition)
     end
 
-    should "'Metadata' header and an update button" do
-      within :css, ".gem-c-heading" do
-        assert page.has_text?("Metadata")
+    should "show document summary and title" do
+      assert page.has_title?("Edit page title")
+
+      row = find_all(".govuk-summary-list__row")
+      assert row[0].has_content?("Assigned to")
+      assert row[1].has_text?("Content type")
+      assert row[1].has_text?("Guide")
+      assert row[2].has_text?("Edition")
+      assert row[2].has_text?("1")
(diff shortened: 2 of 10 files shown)


test/integration/edition_edit_test.rb
```diff
@@ -35,4 +31,59 @@ class EditionEditTest < IntegrationTest
...
     assert page.has_text?("Unpublish")
   end
+
+  context "#metadata" do
+    context "#draft" do
+      setup do
+        click_link("Metadata")
+      end
+
+      should "'Metadata' header and an update button" do
```
mtaylorgds, line 41: This does not read as a sentence. Perhaps use something like "see the 'Metadata' header and an update button" instead.

test/integration/edition_edit_test.rb
```diff
@@ -35,4 +31,59 @@ class EditionEditTest < IntegrationTest
...
+        choose("Welsh")
+        click_button("Update")
+
+        assert find(".gem-c-radio input[value='cy']").checked?
+        assert page.has_text?("Metadata has successfully updated")
+        assert page.has_field?("artefact[slug]", with: "changed-slug")
+      end
+    end
+
+    context "#published" do
```
mtaylorgds, line 65: Similar to `draft`, `published` is not the name of a method being tested, so we shouldn't have the # prefix in the context name. Might be worth elaborating a bit in the context name as well, for example, "viewing a published edition".

test/integration/edition_edit_test.rb
```diff
@@ -35,4 +31,59 @@ class EditionEditTest < IntegrationTest
...
+          ).id,
+          state: "published",
+          slug: "can-i-get-a-driving-licence",
+        )
+
+        visit edition_path(edition)
+        click_link("Metadata")
+      end
+
+      should "show current value for slug and language as texts" do
```
mtaylorgds, line 81: Not sure "as texts" is good English. We can probably just omit that bit anyway as it should be clear from reading the assertions that it is testing the displayed UI text.

mtaylorgds (commented):
This could be done in a subsequent PR (as part of the same story), but the validation needs work—there should be inline validation messages, and the message at the top should be more readable, and should be in an error summary, with links to the appropriate field.

I thought we were transitioning by content type—so you would only see the design system tabs for help pages and answers, but would still see the old bootstrap UI for other content types—has this changed?

Is there any test coverage for the validation messages?

app/models/artefact.rb
```diff
@@ -270,6 +270,17 @@ def exact_route?
     prefixes.empty?
   end
 
+  def locale_to_language(locale)
```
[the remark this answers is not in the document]
syed-ali-tw: do you mean a helper module?
Although its used for UI at the moment, there might be other uses at well, to me its a translation for a field in the Artefact model.

app/helpers/tabbed_nav_helper.rb
```diff
@@ -29,4 +28,15 @@ def edit_nav_item(label, href, current)
       },
     ]
   end
+
+  def current_tab
+    tab = request.path.split("/") & items
```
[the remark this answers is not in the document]
syed-ali-tw: will rename items as suggested, makes more sense.
I know what you mean though, using & is not so uncommon. Although the name of the tab should always be the last element, I could not find in the documentation that string split would guarantee the order. thus using & felt bit more safer albiet more clever

app/helpers/tabbed_nav_helper.rb
```diff
@@ -29,4 +28,15 @@ def edit_nav_item(label, href, current)
       },
     ]
   end
+
+  def current_tab
```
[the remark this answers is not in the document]
mtaylorgds: This comment was about the method name, which hasn't changed.

app/helpers/tabbed_nav_helper.rb
```diff
@@ -29,4 +28,15 @@ def edit_nav_item(label, href, current)
...
+    current_tab_name = request.path.split("/") & items
```
mtaylorgds, line 33: Maybe this one can be `tab_name` (though really it's an array, so should be something like `potential_tab_names`?

syed-ali-tw: have just kept it as current tab and returning the first element of the array itself.

test/integration/edition_edit_test.rb
```diff
@@ -8,31 +8,90 @@ class EditionEditTest < IntegrationTest
...
+      assert page.has_text?("Edit")
+      assert page.has_text?("Tagging")
+      assert page.has_text?("Metadata")
+      assert page.has_text?("History and notes")
+      assert page.has_text?("Admin")
+      assert page.has_text?("Related external links")
+      assert page.has_text?("Unpublish")
+    end
+
+    context "#metadata" do
```
mtaylorgds, line 39: I missed this one last time round, but `metadata` is not the name of the method in this context, so would be better as something like "viewing the metadata tab".

test/integration/edition_edit_test.rb
```diff
@@ -8,31 +8,90 @@ class EditionEditTest < IntegrationTest
...
+        click_button("Update")
+
+        assert find(".gem-c-radio input[value='cy']").checked?
+        assert page.has_text?("Metadata has successfully updated")
+        assert page.has_field?("artefact[slug]", with: "changed-slug")
+      end
+    end
+
+    context "when edition is draft" do
+      context "#metadata" do
```
mtaylorgds, line 69: Similarly `metadata` isn't a method name in this context (see earlier comment).

test/integration/edition_edit_test.rb
```diff
@@ -8,31 +8,90 @@ class EditionEditTest < IntegrationTest
...
+        choose("Welsh")
+        click_button("Update")
+
+        assert find(".gem-c-radio input[value='cy']").checked?
+        assert page.has_text?("Metadata has successfully updated")
+        assert page.has_field?("artefact[slug]", with: "changed-slug")
+      end
+    end
+
+    context "when edition is draft" do
```
mtaylorgds, line 68: I think this is supposed to be "when edition is published"?

And it should not be within the preceding "when edition is draft" context, but at the same level.

mtaylorgds (commented):
A few comments related to the test file structure.

mtaylorgds (approved):

PR #2350, opened 2024-09-30
454 customise select2 styles

davidtrussler:
[Trello](https://trello.com/c/j3xsgpqe)

These changes add a number of over-ride styles to existing bootstrap styles that became necessary after the upgrade to Dart Sass. 

The need for these became evident in the request receive to fix a problem that became evident after that upgrade (detailed in the Trello card), concerning the user's ability to remove a selection from a dropdown selector. Upon closer inspection it became apparent that a number of other over-ride styles were required to restore some instances of broken UI as follows (see screenshots below): 
- The crown is missing in the header
- The dropdown is missing the remove and arrow icons
- The Select2 search bar is missing the search icon and some shading around the box
- The Select2 search bar in active state is missing the spinner icon

There are some styles added specifically for the display of right-to-left languages.

In addition some over-rides have been added for the Select2 element that are not used in this app but are present for completeness and potential future use. 

|Over-ride added|Current|Update|
|-|-|-|
|Crown in navbar|Screenshot 2024-09-30 at 17 34 22|Screenshot 2024-09-30 at 17 34 07|
|Select2 dropdown|Screenshot 2024-09-30 at 17 37 10|Screenshot 2024-09-30 at 17 36 55|
|Select2 search bar|Screenshot 2024-09-30 at 17 41 24|Screenshot 2024-09-30 at 17 41 43|
|Select2 search bar (active state)|Screenshot 2024-09-30 at 17 44 23|Screenshot 2024-09-30 at 17 43 52|

### DIFF
--- a/app/assets/stylesheets/bootstrap_and_overrides.scss
+++ b/app/assets/stylesheets/bootstrap_and_overrides.scss
@@ -1,4 +1,5 @@
-// stylelint-disable selector-no-qualifying-type
+// stylelint-disable scss/comment-no-loud, selector-no-qualifying-type
+
 .alert .modal {
   color: $text-color;
 }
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/bootstrap_and_overrides.scss
```diff
@@ -108,3 +108,84 @@ legend p {
...
+      background: url("select2.png") no-repeat 0 1px
+    }
+  }
+}
+
+.navbar-brand {
+  background: url("govuk_admin_template/header-crown.png") 0 .67em no-repeat;
+}
+
+// stylelint-disable function-linear-gradient-no-nonstandard-direction
```
mtaylorgds, line 128: Should we be re-enabling the checks after (or using `stylelint-disable-line` or `stylelint-disable-next-line` on each individual line)?

(also applies later in this file)

mtaylorgds, line 128: Seems like, from [this warning in the stylelint docs](https://stylelint.io/user-guide/ignore-code/#:~:text=WARNING,or%20value%20lists.), that maybe it would be best to use `/* */` comments?

(also applies later in this file)

davidtrussler: > Seems like, from [this warning in the stylelint docs](https://stylelint.io/user-guide/ignore-code/#:~:text=WARNING,or%20value%20lists.), that maybe it would be best to use `/* */` comments?
> 
> (also applies later in this file)

Yes I agree except `/* */` fails the linter itself with the message: "Expected // for comments instead of /*  scss/comment-no-loud", which seems to contradict its own documentation. To get round this I've added a comment to the one already in place at the top to disable that one.

mtaylorgds (approved):

PR #2352, opened 2024-10-01
439 implement new summary card component

davidtrussler:
[Trello](https://trello.com/c/LBA1E9tm/439-implement-new-summary-card-component-publisher)

The changes here:
- remove the existing component
- remove the existing stylesheet
- remove the reference to the existing stylesheet from `application.scss` 
- add the component from the Publishing Components gem
- add the Summary card styles from the Publishing Components gem

Additionally 
- The page summary section at the top and the links below have been separated into `<section>` elements. This is a semantic improvement though here it also serves to avoid the double border that is rendering under the top section in the current implementation, caused by the way the component is designed to display when there are multiple instances of it in a section.
- The way the colours used to represent different states (Draft, Published etc.) are defined has been updated to use the `app/assets/stylesheets/tags.scss` file, which was previously created to map specific colours onto states. This ensures there is a "single source of truth" for the correspondence between state and colour. 

|Before|After|
|-|-|
|Screenshot 2024-10-02 at 14 58 29|Screenshot 2024-10-02 at 14 47 21|


syed-ali-tw (approved):

syed-ali-tw:
LGTM.

PR #2354, opened 2024-10-04
Restrict access to publications based on the user's organisation

mtaylorgds:
## What
Prevent non-GDS users from accessing editions that are "owned" by other organisations. GDS users will still see all editions.

Introduces the concept of editions having an array of owning organisations, as represented by an array of those organisations' slugs.

The organisation slug is used as it is the only information we have to tie a user to an organisation (via the user's `organisation_slug`).

This is part of our work to open up Mainstream Publisher to users outside of GDS.

This PR does not include assigning any content items to organisations—that will follow as a separate piece of work. This means that, for now, turning on the feature toggle will result in non-GDS users simply seeing no content in Mainstream Publisher.

## Feature toggle
The changes are behind a new feature toggle `restrict_access_by_org`, which defaults to off.

## Trello
[Trello card](https://trello.com/c/rAqyyCyt/345-restrict-access-to-publications-based-on-the-users-organisation-publisher)

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -35,9 +35,19 @@ class Edition
 
   field :auth_bypass_id,       type: String, default: -> { SecureRandom.uuid }
 
+  field :owning_org_slugs, type: Array, default: []
+
   belongs_to :assigned_to, class_name: "User", optional: true
 
   embeds_many :link_check_reports
+
+  scope :accessible_to,
+        lambda { |user|
+          return all unless Flipflop.enabled?(:restrict_access_by_org)
+          return all if user.gds_editor?
+
+          where(owning_org_slugs: user.organisation_slug)
+        }
 
   # state_machine comes from Workflow
   state_machine.states.map(&:name).each do |state|
@@ -501,6 +511,13 @@ class Edition
     end
 
     paths
+  end
+
+  def is_accessible_to?(user)
+    return true unless Flipflop.enabled?(:restrict_access_by_org)
+    return true if user.gds_editor?
+
+    owning_org_slugs.include?(user.organisation_slug)
   end
 
 private
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -35,7 +35,7 @@ class Edition
 
   field :auth_bypass_id,       type: String, default: -> { SecureRandom.uuid }
 
-  field :owning_org_slugs, type: Array, default: []
+  field :owning_org_content_ids, type: Array, default: []
 
   belongs_to :assigned_to, class_name: "User", optional: true
 
@@ -46,7 +46,7 @@ class Edition
           return all unless Flipflop.enabled?(:restrict_access_by_org)
           return all if user.gds_editor?
 
-          where(owning_org_slugs: user.organisation_slug)
+          where(owning_org_content_ids: user.organisation_content_id)
         }
 
   # state_machine comes from Workflow
@@ -517,7 +517,7 @@ class Edition
     return true unless Flipflop.enabled?(:restrict_access_by_org)
     return true if user.gds_editor?
 
-    owning_org_slugs.include?(user.organisation_slug)
+    owning_org_content_ids.include?(user.organisation_content_id)
   end
 
 private
--- a/test/functional/editions_controller_test.rb
+++ b/test/functional/editions_controller_test.rb
@@ -5,6 +5,9 @@ class EditionsControllerTest < ActionController::TestCase
     login_as_stub_user
     stub_linkables
     stub_holidays_used_by_fact_check
+
+    test_strategy = Flipflop::FeatureSet.current.test!
+    test_strategy.switch!(:restrict_access_by_org, false)
   end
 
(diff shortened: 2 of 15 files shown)


app/models/edition.rb
```diff
@@ -503,6 +513,13 @@ def paths
     paths
   end
 
+  def accessible_to?(user)
```
syed-ali-tw, line 516: can we rename this to `is_accessible_to?` only because scope is similar named.

test/functional/editions_controller_test.rb
```diff
@@ -64,4 +64,55 @@ class EditionsControllerTest < ActionController::TestCase
       assert_equal EditionsController.new.method(:metadata).super_method.name, :show
     end
   end
+
+  context "when 'restrict_access_by_org' feature toggle is disabled" do
```
syed-ali-tw, line 68: instead of disabled, can we use false. as 'false' best describes the value of the switch

test/functional/editions_controller_test.rb
```diff
@@ -64,4 +64,55 @@ class EditionsControllerTest < ActionController::TestCase
...
+
+          get action, params: { id: @edition.id }
+
+          assert_response :ok
+        end
+      end
+    end
+  end
+
+  context "when 'restrict_access_by_org' feature toggle is enabled" do
```
syed-ali-tw, line 91: instead of enabled, can we use returns true. as 'true' best describes the value of the switch

test/functional/editions_controller_test.rb
```diff
@@ -64,4 +64,55 @@ class EditionsControllerTest < ActionController::TestCase
       assert_equal EditionsController.new.method(:metadata).super_method.name, :show
     end
   end
+
+  context "when 'restrict_access_by_org' feature toggle is disabled" do
+    setup do
+      test_strategy = Flipflop::FeatureSet.current.test!
```
syed-ali-tw, line 70: We can move the test_strategy to the top most setup block of the class, avoids repeating.

test/functional/legacy_editions_controller_test.rb
```diff
@@ -1304,4 +1304,55 @@ class LegacyEditionsControllerTest < ActionController::TestCase
       end
     end
   end
+
+  context "when 'restrict_access_by_org' feature toggle is disabled" do
+    setup do
+      test_strategy = Flipflop::FeatureSet.current.test!
```
syed-ali-tw, line 1310: same as above, We can move the test_strategy to the top most setup block of the class, avoids repeating.

test/functional/editions_controller_test.rb
mtaylorgds: I think enabled/disabled is common parlance for feature toggle status, and reads better than true/false.

test/functional/editions_controller_test.rb
mtaylorgds: I think enabled/disabled is common parlance for feature toggle status, and reads better than true/false.

test/functional/editions_controller_test.rb
mtaylorgds: This context block is explicitly for when the feature toggle is disabled, so I think it makes sense for the test_strategy setup to be here. Technically it shouldn't be needed since the toggle will be off by default, but it felt a bit weird to have the context say the feature toggle is off and then not actually do anything to turn it off.

test/functional/editions_controller_test.rb
```diff
@@ -64,4 +67,50 @@ class EditionsControllerTest < ActionController::TestCase
...
+    end
+  end
+
+  context "when 'restrict_access_by_org' feature toggle is enabled" do
+    setup do
+      test_strategy = Flipflop::FeatureSet.current.test!
+      test_strategy.switch!(:restrict_access_by_org, true)
+    end
+
+    teardown do
```
syed-ali-tw, line 95: more of a question because I am not sure, do we need this teardown block? will the setup in test strategy be enough to make sure switch is set to 'false' before each test?

test/functional/legacy_editions_controller_test.rb
```diff
@@ -1304,4 +1307,50 @@ class LegacyEditionsControllerTest < ActionController::TestCase
...
+    end
+  end
+
+  context "when 'restrict_access_by_org' feature toggle is enabled" do
+    setup do
+      test_strategy = Flipflop::FeatureSet.current.test!
+      test_strategy.switch!(:restrict_access_by_org, true)
+    end
+
+    teardown do
```
syed-ali-tw, line 1335: same as [above comment](https://github.com/alphagov/publisher/pull/2354/files#r1791547720)

test/functional/editions_controller_test.rb
mtaylorgds: It could affect other tests if we left it turned on (in other test files), unless we turn it off in every other test file. I think it's best turning it off here to make sure.

syed-ali-tw (approved):

baisa (approved):

PR #2377, opened 2024-10-16
Amend Design System layout to make correct use of the grid system

davidtrussler:
[Trello](https://trello.com/c/Q7Gx5wiI/486-amend-design-system-layout-to-make-correct-use-of-the-grid-system)

These changes updates the Design System template to apply the `govuk-grid-row` wrapper on each view rather than globally. This is necessary to ensure that the Design System Grid is fully utilised throughout the app, by allowing the `govuk-grid-row` wrapper to be able to be used multiple times in each view where required.

In summary:
- Delete `govuk-grid-row` wrapper from Design System template
- Add `govuk-grid-row` wrapper to
  - Publications view
  - Downtimes views
  - Reports view
  - Edit views

### DIFF
--- a/app/assets/stylesheets/editions.scss
+++ b/app/assets/stylesheets/editions.scss
@@ -1,4 +1,4 @@
-.editions__edit {
+.editions__edit__summary {
   .govuk-tag {
     display: inline;
   }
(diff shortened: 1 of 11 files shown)


ellohez (approved):
LGTM!

PR #2380, opened 2024-10-18
Send `last_edited_by_editor_id` to Publishing API

pezholio:
This is a new field that will allow us to identify the last person who has edited an Edition within the Publishing API.

### DIFF
--- a/app/presenters/formats/edition_format_presenter.rb
+++ b/app/presenters/formats/edition_format_presenter.rb
@@ -28,6 +28,7 @@ module Formats
         schema_name:,
         document_type:,
         public_updated_at: public_updated_at.rfc3339(3),
+        last_edited_by_editor_id:,
         publishing_app: "publisher",
         rendering_app:,
         routes:,
@@ -36,7 +37,7 @@ module Formats
         change_note: edition.latest_change_note,
         details:,
         locale: artefact.language,
-      }
+      }.compact
     end
 
     def schema_name
@@ -96,6 +97,13 @@ module Formats
 
     def public_updated_at
       edition.public_updated_at || edition.updated_at
+    end
+
+    # We can't reliably get the exact user who last edited the edition,
+    # so we rely on who created the edition, which is a fair enough
+    # approximation
+    def last_edited_by_editor_id
+      edition.created_by&.uid
     end
   end
 end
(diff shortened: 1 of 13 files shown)


test/support/presenter_test_helpers.rb
```diff
@@ -0,0 +1,10 @@
+module PresenterTestHelpers
+  def it_includes_last_edited_by_editor_id
+    should "[:last_edited_by_editor_id]" do
+      editor = FactoryBot.create(:user)
+      edition.actions.create!(request_type: Action::CREATE, requester: editor)
+
+      assert_equal editor.uid, result[:last_edited_by_editor_id]
+    end
+  end
+end
```
baisa, line 10: We should include test that ensures us we are actually sending last editor id, not id of the user who created the edition. Although it seems like we do not have access to this information atm.

pezholio: Cool, I've updated the tests to cover both eventualities. Looking at the behaviour of the code, the only time a `CREATE` action is created is when creating a brand new edition (i.e one without any children), if a new edition is created from an existing edition, the creates a `NEW_VERSION` action.

test/support/presenter_test_helpers.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    test "[:last_edited_by_editor_id] should return the creator of a new version" do
+      editor = FactoryBot.create(:user)
+      edition.actions.create!(request_type: Action::NEW_VERSION, requester: editor)
+
+      assert_equal editor.uid, result[:last_edited_by_editor_id]
+    end
+
+    test "[:last_edited_by_editor_id] should return nil when there are no new version or create actions" do
+      assert_equal nil, result[:last_edited_by_editor_id]
```
baisa, line 18: Consider using `assert_nil` as `asset_equal nil` will soon be deprecated.

app/presenters/formats/edition_format_presenter.rb
```diff
@@ -97,5 +98,9 @@ def major_change?
     def public_updated_at
       edition.public_updated_at || edition.updated_at
     end
+
+    def last_edited_by_editor_id
+      edition.created_by&.uid
+    end
```
baisa, line 104: Maybe worth adding a comment here to explain that we do not really have access to information who edited the edition last, so we rely on created_by.

baisa (commented):
Once we add explanatory comment (as below) it LGTM

baisa (approved):

PR #2385, opened 2024-10-21
Crating a rake task to publish any missed scheduled editions

baisa:
[Trello card](https://trello.com/c/FDV98XMH/489-ensure-scheduled-editions-are-published-mainstream)

Daily cronjob is set up to run this rake task in govuk-helm-charts [PR](https://github.com/alphagov/govuk-helm-charts/pull/2716)

### DIFF
--- a/lib/tasks/publish_missed_scheduled_editions.rake
+++ b/lib/tasks/publish_missed_scheduled_editions.rake
@@ -1,0 +1,11 @@
+namespace :editions do
+  desc "Publish editions missed in scheduled publishing"
+  task publish_missed_scheduled_editions: :environment do
+    scheduled_editions = Edition.scheduled_for_publishing.where(:publish_at.lt => Time.zone.now)
+
+    scheduled_editions.each do |edition|
+      ScheduledPublisher.new.perform(edition.id.to_s)
+      Rails.logger.warn "Published missed scheduled edition with ID: #{edition.id} "
+    end
+  end
+end
--- a/test/unit/tasks/publish_missed_scheduled_editions_test.rb
+++ b/test/unit/tasks/publish_missed_scheduled_editions_test.rb
@@ -1,0 +1,27 @@
+require "test_helper"
+require "rake"
+
+class PublishMissedScheduledEditionsTaskTest < ActiveSupport::TestCase
+  def setup
+    Rake::Task["editions:publish_missed_scheduled_editions"].reenable
+  end
+
+  test "publishes editions that have missed their scheduled publish time" do
+    missed_edition = FactoryBot.create(:edition, state: "scheduled_for_publishing", publish_at: 1.hour.from_now)
+    missed_edition.update_attribute(:publish_at, 1.hour.ago) # rubocop:disable Rails/SkipsModelValidations
+
+    Rake::Task["editions:publish_missed_scheduled_editions"].invoke
+    missed_edition.reload
+
+    assert_equal "published", missed_edition.state
+  end
+
+  test "does not publish editions with publishing scheduled in the future" do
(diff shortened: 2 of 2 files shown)


lib/tasks/publish_missed_scheduled_editions.rake
```diff
@@ -0,0 +1,11 @@
+namespace :editions do
+  desc "Publish editions missed in scheduled publishing"
+  task publish_missed_scheduled_editions: :environment do
+    scheduled_editions = Edition.scheduled_for_publishing.where(:publish_at.lt => Time.zone.now)
+
+    scheduled_editions.each do |edition|
+      ScheduledPublisher.new.perform(edition.id.to_s)
+      Rails.logger.info "Published edition with ID: #{edition.id}"
```
mtaylorgds, line 8: Maybe we could be more explicit in the message about it being an edition that got missed from publication (to make searching for these issues in the logs easier).

We could also consider logging at `warn` rather than `info` level.

lib/tasks/publish_missed_scheduled_editions.rake
```diff
@@ -0,0 +1,11 @@
+namespace :editions do
+  desc "Publish editions missed in scheduled publishing"
+  task publish_missed_scheduled_editions: :environment do
+    scheduled_editions = Edition.scheduled_for_publishing.where(:publish_at.lt => Time.zone.now)
```
mtaylorgds, line 4: I wonder whether it's worth doing something more [similar to Whitehall](https://github.com/alphagov/whitehall/blob/267d615c1fb84b31b353128ce7cee49a745a952c/lib/tasks/scheduled_publishing.rake#L60) here, where it checks the current queue, which would reduce the chance of trying to queue a job that is already queued (but hasn't been processed yet)?

test/unit/tasks/publish_missed_scheduled_editions_test.rb
```diff
@@ -0,0 +1,27 @@
+require "test_helper"
+require "rake"
+
+class EditionsPublishMissedScheduledEditionsTaskTest < ActiveSupport::TestCase
```
mtaylorgds, line 4: What's the reasoning behind the extra `Editions` at the start of the class name here?

test/unit/tasks/publish_missed_scheduled_editions_test.rb
```diff
@@ -0,0 +1,27 @@
...
+    missed_edition = FactoryBot.create(:edition, state: "scheduled_for_publishing", publish_at: 1.hour.from_now)
+    missed_edition.update_attribute(:publish_at, 1.hour.ago) # rubocop:disable Rails/SkipsModelValidations
+
+    Rake::Task["editions:publish_missed_scheduled_editions"].invoke
+    missed_edition.reload
+
+    assert_equal "published", missed_edition.state
+  end
+
+  test "do not publish editions with publishing scheduled in the future" do
```
mtaylorgds, line 19: Consider renaming test to "does not…" in order to use the same tense as the other test.

lib/tasks/publish_missed_scheduled_editions.rake
baisa: As sidekiq seems to be a culprit as for why those editions were not published, and for the reasons we were not able to diagnose (see details in Trello card)  I think it is safer to publish them imitating manual action.

test/unit/tasks/publish_missed_scheduled_editions_test.rb
baisa: Yes it was because of the namespace, but actually it will be more readable without it 👍

mtaylorgds (approved):

PR #2386, opened 2024-10-22
Add route for browser extension to get routed to publisher

syed-ali-tw:
[Trello](https://trello.com/c/8XrdpwX2/484-browser-extension-doesnt-work-on-transitioned-publication-page-publications-page?filter=label:Design%20System%20MVP)

As Mainstream publisher moves to new design sytem, we would need to make changes to the current external link that navigates to filter publication page, we did discuss [different options](https://docs.google.com/document/d/1DPl6MuUPDMLqRHjvbt3ByJyGVQT_F8on0Ek_-udcgjE/edit?tab=t.0) we had to make this work.

This PR is coupled with the [changes in browser extension ](https://github.com/alphagov/govuk-browser-extension/pull/212)

### DIFF
--- a/app/controllers/content_item_controller.rb
+++ b/app/controllers/content_item_controller.rb
@@ -1,0 +1,21 @@
+class ContentItemController < ApplicationController
+  def by_content_id
+    artefact =
+      Artefact.find_by(content_id: params[:content_id])
+
+    if artefact
+      redirect_to edition_path(artefact.latest_edition)
+    else
+      redirect_to_root_path_with_error
+    end
+  rescue StandardError
+    redirect_to_root_path_with_error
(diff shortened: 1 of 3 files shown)


mtaylorgds (commented):
Couple of comments:

1. I'm not sure we should be referencing the browser extension in our naming. Although we're building it to support that, it doesn't actually have any dependency on the browser extension (could reference why we've created the endpoint in the commit message).
2. When I tried testing it locally I didn't see the error message. I checked both the bootstrap and Design System UIs.

mtaylorgds (commented):
LGTM

mtaylorgds (approved):

PR #2388, opened 2024-10-25
407 | Update the edit tab to the Design System for "Answer" and "Help" content types

davidtrussler:
[Trello](https://trello.com/c/CHePz7xP/407-edit-answer-help-content-types-edit-page)

This work adds the Edit tab to the Edit page. There are a few scenarios to consider: 
- A new draft: doesn't display the change note section
- It displays a success message when saved
- It displays an error message when the title field is left blank
- A new edition of a published draft: displays the change note section

Additionally it also makes a small change to the popular links page to create some generic styles for the sticky options panel. 

This only contains the code required to display Answer and Help content types: there will be further work to ensure that other content types are not displayed as Design System pages. 

|Scenario|Screenshot|
|-|-|
|New document|publisher dev gov uk_editions_6691241ea73bb8001d3916d8|
|New document with success message|publisher dev gov uk_editions_6691241ea73bb8001d3916d8 (1)|
|New document with error message|publisher dev gov uk_editions_6691241ea73bb8001d3916d8 (2)|
|New edition of a published document|publisher dev gov uk_editions_671135ff5aeeb50039816b77|


app/controllers/editions_controller.rb
```diff
@@ -27,6 +30,25 @@ def show
   alias_method :metadata, :show
   alias_method :unpublish, :show
 
+  def update
+    params = permitted_params
```
mtaylorgds, line 34: I'd be wary of calling this variable `params` as that is already the name of a field accessible in rails controllers (maybe use `update_params` instead).

app/controllers/editions_controller.rb
```diff
@@ -27,6 +30,25 @@ def show
...
+    @resource.assign_attributes(params[:edition])
+    success = @resource.update!
+
+    if success == true
```
mtaylorgds, line 38: Don't think we really need the `success` variable.
```suggestion
    if @resource.update!
```

app/controllers/editions_controller.rb
```diff
@@ -71,6 +93,14 @@ def setup_view_paths
     setup_view_paths_for(resource)
   end
 
+  def update_action_is_publish?
```
mtaylorgds, line 96: Since we don't have a way to publish yet, do we need this method?

If we do, I think this method can be made `private` (which it is in the legacy controller).

app/controllers/editions_controller.rb
```diff
@@ -71,6 +93,14 @@ def setup_view_paths
...
+    attempted_activity == :publish
+  end
+
+  def attempted_activity
```
mtaylorgds, line 100: I think this method can be made `private` (which it is in the legacy controller).

test/functional/editions_controller_test.rb
```diff
@@ -200,4 +200,66 @@ class EditionsControllerTest < ActionController::TestCase
       end
     end
   end
+
+  context "edit" do
+    setup do
+      @answer = FactoryBot.create(:answer_edition)
+    end
+
+    should "render the page ok" do
```
mtaylorgds, line 209: Do we need this test? Isn't it just a simpler form of the one on line 47? We could then get rid of the "edit" context and just have the "#update" one.

test/functional/editions_controller_test.rb
```diff
@@ -200,4 +200,66 @@ class EditionsControllerTest < ActionController::TestCase
...
+      get :show, params: { id: @answer.id }
+      assert_response :ok
+    end
+
+    context "#update" do
+      should "show update and success message and render show template when saved" do
+        get :update, params: {
```
mtaylorgds, line 216: I'm surprised this works as a `get`—I'd have expected an update to be a `post` request.

mtaylorgds (commented):
Added a few comments.

app/controllers/editions_controller.rb
ellohez: I agree, I think it will cause confusion. I have added a comment about the permitted_params method below and I think if the .permit was changed to the pattern I have mentioned, then you could maybe change the line below to:

`@resource.assign_attributes(permitted_params)`

app/controllers/editions_controller.rb
```diff
@@ -118,4 +153,8 @@ def redirect_url
   def description(resource)
     resource.format.underscore.humanize
   end
+
+  def permitted_params
+    params.permit(edition: %i[title overview in_beta body major_change change_note])
```
ellohez, line 158: Not sure if this matters but I have just been learning about 'strong parameters' and I have only seen this written as:
`params.require(:edition).permit(%i[title overview in_beta body major_change change_note])`
This will raise an error if the :edition attribute is missing. Maybe that isn't what you want here though?

[Use Strong Parameters](https://www.mintbit.com/blog/securing-ruby-on-rails-applications-part-3-use-strong-parameters)

ellohez (commented):
Just a couple of comments (and I could well be wrong!)

app/controllers/editions_controller.rb
davidtrussler: My feeling is that that makes a lot of sense and helps with the naming above. So I'm changing this accordingly. 👍

app/controllers/editions_controller.rb
davidtrussler: I believe we do not, nor do we need `UpdateWorker.perform_async(resource.id.to_s, update_action_is_publish?)` above that, so removing both.

test/functional/editions_controller_test.rb
davidtrussler: Yes that's a fair point. I've updated all of these requests to `post` for correctness.

app/controllers/editions_controller.rb
```diff
@@ -27,6 +30,23 @@ def show
   alias_method :metadata, :show
   alias_method :unpublish, :show
 
+  def update
+    @resource.assign_attributes(permitted_params)
+
+    if @resource.update!
```
mtaylorgds, line 36: I'm not sure what `@resource.update!` is actually doing/comes from—if `@resource` is the model object, shouldn't we be calling `save`/`save!`? I think the `InheritedResources` base class maybe causing confusion, and wonder whether we should stop using it.

It seems like it is throwing an exception when there is a validation error, as I'm seeing both the validation message and the error message from `render_show_page_with_error` when I try and save an Answer with a blank title, for example.

app/views/editions/secondary_nav_tabs/_edit.html.erb
```diff
@@ -0,0 +1,127 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "Edit",
+      heading_level: 2,
+      margin_bottom: 5,
+    } %>
+  </div>
+</div>
+
+<%= resource_form(@resource) do |f| %>
```
mtaylorgds, line 11: Should we be using `resource_form`? It appears to be a helper method that adds some "data-module" attributes to do with AJAX-saving.

app/views/editions/secondary_nav_tabs/_edit.html.erb
```diff
@@ -0,0 +1,127 @@
...
+            heading: "Add a public change note",
+            heading_level: 3,
+            heading_size: "m",
+            name: "edition[major_change]",
+            hint: "Telling users when published information has changed is important for transparency.",
+            items: [
+              {
+                hint_text: "A change note will be published on the page and emailed to users subscribed to email alerts. The ‘last updated’ date will change.",
+                text: "Yes - information has been added, updated or removed.",
+                checked: resource.major_change,
```
mtaylorgds, line 78: Should we use `@resource` instead of `resource` (the latter seems to come from the `InheritedBase` gem, and we're using `@resource` in most places in this file—it would be good to be consistent.

app/views/editions/secondary_nav_tabs/_edit.html.erb
```diff
@@ -0,0 +1,127 @@
...
+                  },
+                  name: "edition[change_note]",
+                  hint: ("<p class=\"govuk-!-margin-0\">Tell users what has changed, where and why. Write in full sentences, leading with the most important words. For example, \"College A has been removed from the registered sponsors list because its licence has been suspended.\"</p><a href=\"https://www.gov.uk/guidance/content-design/writing-for-gov-uk#change-notes\" target=\"_blank\">Guidance on change notes (opens in a new tab)</a>").html_safe,
+                }),
+              },
+              {
+                value: false,
+                text: "No",
+                bold: true,
+                checked: !resource.major_change,
```
mtaylorgds, line 94: Same comment about `@resource` vs `resource`.

app/controllers/editions_controller.rb
davidtrussler: @mtaylorgds I have updated this section of the controller to call `save` in a way that seems to work and doesn't show an error message with the validation method. It's not really addressing your concerns about the `InheritedResources` class though as I don't fully understand why that is being used here and the consequences of doing it differently. I'm feeling that may be a slightly separate piece of work if we can get this to work as expected for the present.

davidtrussler:
> Added some more comments after a fresh look.

@mtaylorgds I've updated stuff in line with your later comments though without considering the consequences of the `InheritedResources` class (as mentioned in one of my comments).

app/controllers/editions_controller.rb
```diff
@@ -27,6 +30,23 @@ def show
...
   alias_method :unpublish, :show
 
+  def update
+    @resource.assign_attributes(permitted_params)
+
+    if @resource.save
+      UpdateWorker.perform_async(resource.id.to_s)
+      flash.now[:success] = "Edition updated successfully."
+    else
+      @resource = resource
```
mtaylorgds, line 40: I know it was in the legacy controller, but I'm not sure this is actually doing anything/is needed—I tried commenting it out and it still seems to work. Thoughts?

test/functional/editions_controller_test.rb
```diff
@@ -200,4 +200,59 @@ class EditionsControllerTest < ActionController::TestCase
...
+        edition: {
+          title: "",
+        },
+      }
+
+      assert_template "show"
+      assert_select ".gem-c-error-summary__list-item", "Enter a title"
+    end
+
+    should "show error message and render show template when the edition could not be updated" do
```
mtaylorgds, line 234: I think we're missing a test for when `save` returns `false`.

mtaylorgds (commented):
A couple more minor comments.

app/controllers/editions_controller.rb
davidtrussler: Yes, makes sense, have deleted that line.

test/functional/editions_controller_test.rb
davidtrussler: I've added a test which might (or might not) do what we'd expect: If we stub save to return false without triggering a validation error we get the standard error rather than a validation or other message. This seems to make sense but if we add a specific message in the controller we don't see this, which would be my expectation. 🤔

test/functional/editions_controller_test.rb
```diff
@@ -200,4 +200,69 @@ class EditionsControllerTest < ActionController::TestCase
...
+        edition: {
+          title: "A title",
+        },
+      }
+
+      assert_template "show"
+      assert_select ".gem-c-error-summary__list-item", "Due to a service problem, the edition couldn't be updated"
+    end
+
+    should "show error message and render show template when save returns false" do
```
mtaylorgds, line 247: Sorry David, I think I was wrong when I said we needed this test—the "show error message and render show template when title field is blank" test is testing this case already.

I'm curious how this test is actually passing, since it _shouldn't_ be showing the service problem error message when save returns false (since there isn't an exception).

davidtrussler: OK, though I understood we were testing subtly different cases here, but I guess not if passing a blank string for title is the only way to make save return false. 

I am equally confused though as to how the test passes WRT to the "Service problem" error: this doesn't display in the browser if we call `@resource.save` rather than `@resource.save!` - which I think is what you'd expect. But if I add an assertion to the test for a blank title string that the "Service problem" error is *not* rendered we get a failure. 😕 

This is clearly wrong but I'm not sure how much to worry about it. It just feels a bit wrong to ignore the apparent contradiction here.

mtaylorgds: It'd be _nice_ to have an understanding of why it's unexpectedly passing in case it's a symptom of something bigger being wrong, but also don't want to invest too much time into it if we're deleting the test…

davidtrussler: Yes, agreed. I've reverted the last commit that adds that test.

davidtrussler:
@mtaylorgds Can you have another look over this? I've updated the two tests a bit, to simplify them a bit and making use of the added `visit_edition_in_draft` and suchlike methods and adding a new one for a new edition of a published one.

test/integration/edition_edit_test.rb
```diff
@@ -225,10 +225,86 @@ class EditionEditTest < IntegrationTest
...
+
+      assert page.has_field?("edition[title]", with: "Changed Title")
+      assert page.has_field?("edition[overview]", with: "Changed Meta tag description")
+      assert find(".gem-c-radio input[value='1']").checked?
+      assert page.has_field?("edition[body]", with: "Changed body")
+      assert page.has_text?("Edition updated successfully.")
+    end
+  end
+
+  context "edit tab" do
```
mtaylorgds, line 285: We should probably avoid having two contexts with the same name ("edit tab")…

Maybe we should add some sub-contexts so we have something like:
```ruby
  context "edit tab" do
    context "editing a draft edition of a new publication" do
      setup do
        visit_edition_in_draft
        click_link("Edit") # Do we need this? We don't have it for the previously-published test
      end

      …tests
    end

    context "editing a draft edition of a previously-published publication" do
    end
  end
```

With those extra contexts, we probably then don't need to duplicate that context in the test names.

mtaylorgds (commented):
One minor comment with the test structure, but otherwise looking good!

test/integration/edition_edit_test.rb
davidtrussler: > We should probably avoid having two contexts with the same name ("edit tab")…
> 
> Maybe we should add some sub-contexts so we have something like:
> 
> ```ruby
>   context "edit tab" do
>     context "editing a draft edition of a new publication" do
>       setup do
>         visit_edition_in_draft
>         click_link("Edit") # Do we need this? We don't have it for the previously-published test
>       end
> 
>       …tests
>     end
> 
>     context "editing a draft edition of a previously-published publication" do
>     end
>   end
> ```
> 
> With those extra contexts, we probably then don't need to duplicate that context in the test names.

Oh yes, didn't even notice we had that duplication of context ... I've added the two sub-contexts to the main one. (I'll squash this commit if it looks good.)

mtaylorgds (approved):
LGTM

PR #2429, opened 2024-11-21
Enable editing the person assigned to an edition

syed-ali-tw:
Add an edit link to the document summary (at the top of the page) when viewing an edition with the GOV.UK Design System template.

On the Bootstrap UI this was managed as just another dropdown field on the "Edit" tab. With the new Design System views, the edit link takes the user to a separate page for selecting the user to be assigned.

### Document summary when viewing an edition that may be edited

image

### Document summary when viewing an edition that may not be edited

image

### Edit assigned person page when nobody is currently assigned

image

### Edit assigned person page when a user is already assigned

image

### Old Bootstrap UI for editing the assigned person

image

## Trello

[Trello card](https://trello.com/c/NMCUGhvR)

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -11,11 +11,15 @@ class EditionsController < InheritedResources::Base
   before_action only: %i[unpublish confirm_unpublish process_unpublish] do
     require_govuk_editor(redirect_path: edition_path(resource))
   end
-  before_action only: %i[progress admin update confirm_destroy] do
+  before_action only: %i[progress admin update confirm_destroy edit_assignee update_assignee] do
     require_editor_permissions
   end
   before_action only: %i[confirm_destroy destroy] do
     destroyable_edition?
+  end
+
+  before_action only: %i[edit_assignee update_assignee] do
+    require_assignee_editable
   end
 
   helper_method :locale_to_language
@@ -35,7 +39,7 @@ class EditionsController < InheritedResources::Base
   alias_method :admin, :show
 
   def update
-    @resource.assign_attributes(permitted_params)
+    @resource.assign_attributes(permitted_update_params)
 
     if @resource.save
       UpdateWorker.perform_async(resource.id.to_s)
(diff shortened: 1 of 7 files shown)


ellohez (approved):
LGTM!

app/controllers/editions_controller.rb
```diff
@@ -176,4 +202,31 @@ def destroyable_edition?
     flash[:danger] = "Cannot delete a #{description(@resource).downcase} that has ever been published."
     redirect_to edition_path(@resource)
   end
+
+  def require_assignee_editable
+    return if can_update_assignee?(@resource)
+
+    flash[:danger] = "Cannot edit assignee that has ever been published."
```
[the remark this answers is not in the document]
mtaylorgds: Good catch, I agree that this wording isn't great. Have changed it to "Cannot edit the assignee of an edition that has been published.".

app/assets/stylesheets/editions.scss
```diff
@@ -14,3 +14,10 @@
     margin-top: govuk-spacing(3)
   }
 }
+
+.edition__edit_assignee {
+  p {
+    padding-left: govuk-spacing(2);
+    font-size: 20px;
```
davidtrussler, line 21: We should avoid using explicit pixel sizes for font-size. Can we use the [design system type scale](https://design-system.service.gov.uk/styles/type-scale/) instead? Also I don't think 20px is a size we generally use.

app/views/editions/secondary_nav_tabs/_edit_assignee.html.erb
```diff
@@ -0,0 +1,47 @@
...
+      <% unless @resource.assignee.nil? %>
+        <%= render "govuk_publishing_components/components/radio", {
+          name: "assignee_id",
+          items: [
+            { value: @resource.assigned_to_id, text: @resource.assignee, checked: true },
+            { value: "none" , text: "None" },
+          ],
+        } %>
+        <% if enabled_users(@resource).any? %>
+          <p class="govuk-body"> or </p>
```
davidtrussler, line 17: It would be better if we can use the [with or divider](https://components.publishing.service.gov.uk/component-guide/radio#with_or_divider) variation on the radio component (here) in this case so that the radio controls all remain within the same fieldset. Otherwise this could cause some confusion for assistive technology. 
In which case I think the custom CSS added for this can be removed.

app/assets/stylesheets/editions.scss
mtaylorgds: Ah, very good catch. I'll take a look at this.

davidtrussler: Actually this comment is probably rendered irrelevant by the one below (in the `edit_assignee` partial) if that is applicable.

app/views/editions/secondary_nav_tabs/_edit_assignee.html.erb
davidtrussler: Also once the radios are combined in a single fieldset I think we should add a legend for that to explain (for screen readers) what these controls are doing. Ideally we would wrap the page heading in a `<legend>` tag but I think that would cause a problem for the alignment of the sticky panel so we can repeat the page title in a visually hidden legend as provided in the [With visually hidden heading](https://components.publishing.service.gov.uk/component-guide/radio/with_visually_hidden_heading) option.

davidtrussler (approved):
Looks good! 👍

PR #2431, opened 2024-11-22
472 admin tab update content type

davidtrussler:
[Trello](https://trello.com/c/qLZmHXIu/472-admin-update-content-type-edit-admin?filter=label:Design%20System%20MVP)

Adds the functionality for the user to update the content type of a document on the Admin tab. The work involves
- adding a new partial for the Admin tab
- adding the `duplicate` method to the editions controller
- updating the editions controller test 
- updating the edit editions test
- a bit of renaming of methods in the editions helper and where these are used in the legacy code

|View|Screenshot|
|-|-|
|Admin tab|Screenshot 2024-11-22 at 12 43 36|
|Redirect to edit page on success|Screenshot 2024-11-22 at 12 47 38|
|Redirect to edit with error when another user has already created a new version|Screenshot 2024-11-22 at 12 58 20|

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -30,13 +30,34 @@ class EditionsController < InheritedResources::Base
 
   def show
     @artefact = @resource.artefact
-
     render action: "show"
   end
 
   alias_method :metadata, :show
   alias_method :unpublish, :show
   alias_method :admin, :show
+
+  def duplicate
+    command = EditionDuplicator.new(@resource, current_user)
+    target_edition_class_name = "#{params[:to]}_edition".classify if params[:to]
+
+    if !@resource.can_create_new_edition?
+      flash[:warning] = "Another person has created a newer edition"
+      redirect_to edition_path(resource)
+    elsif command.duplicate(target_edition_class_name, current_user)
+      new_edition = command.new_edition
+      UpdateWorker.perform_async(new_edition.id.to_s)
+      flash[:success] = "New edition created"
+      redirect_to edition_path(new_edition)
+    else
+      flash[:danger] = command.error_message
+      redirect_to edition_path(resource)
+    end
+  rescue StandardError => e
+    Rails.logger.error "Error #{e.class} #{e.message}"
+    @resource.errors.add(:show, "Due to a service problem, the edition couldn't be duplicated")
+    render "show"
+  end
 
   def update
     @resource.assign_attributes(permitted_update_params)
--- a/app/helpers/editions_helper.rb
+++ b/app/helpers/editions_helper.rb
@@ -19,9 +19,30 @@ module EditionsHelper
     nested_form_for resource, as: :edition, url: edition_path(resource), html: html_options, &form_definition
   end
 
-  def format_conversion_select_options(edition)
+  def legacy_format_conversion_select_options(edition)
     possible_target_formats = Edition.convertible_formats - [edition.artefact.kind]
     possible_target_formats.map { |format_name| [format_name.humanize, format_name] }
+  end
+
+  def conversion_items(edition)
+    radio_options_hints = {
+      "answer" => "One page guidance",
+      "completed_transaction" => "Done page for end of a service",
+      "guide" => "Multi-page guidance",
+      "help_page" => "Info about GOV.UK website, for example Privacy",
+      "place" => "Postcode look-up for places/services near you",
+      "simple_smart_answer" => "Simple questions and answers that route users to relevant outcomes",
+      "transaction" => "Start page for a service",
+    }
+
+    possible_target_formats = Edition.convertible_formats - [edition.artefact.kind]
+    possible_target_formats.map do |format_name|
+      {
+        text: format_name.humanize,
+        value: format_name,
+        hint_text: radio_options_hints[format_name],
+      }
+    end
   end
 
   def legacy_format_filter_selection_options
--- a/app/views/editions/secondary_nav_tabs/_admin.html.erb
+++ b/app/views/editions/secondary_nav_tabs/_admin.html.erb
@@ -1,12 +1,36 @@
 <% @edition = @resource %>
+
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds editions__admin__tab">
     <%= header_for("Admin") %>
-    <%= render "govuk_publishing_components/components/list", {
-      items: [
-        ( primary_button_for(@edition, skip_fact_check_edition_path(@edition), "Skip fact check") if @edition.fact_check? ),
-        ( link_to("Delete edition #{@edition.version_number}", confirm_destroy_edition_path(@resource), class: "govuk-link gem-link--destructive") if @edition.can_destroy? ),
-      ],
-    } %>
+
+    <% if @edition.can_create_new_edition? && @edition.published? %>
+      <p class="govuk-body">No content will be lost, but content in some fields might not make it into the new edition. If it can't be copied to a new content type it will still be available in the previous edition. Content in multiple fields might be combined into a single field.</p>
+
+      <% if @edition.respond_to?(:parts) %>
+        <p class="govuk-body">All parts of Guide Editions will be copied across. If the format you are converting to doesn't have parts, the content of all the parts will be copied into the body, with the part title displayed as a top-level sub-heading.</p>
+      <% end %>
+
+      <%= form_for @resource, url: duplicate_edition_path(@resource), method: "post" do %>
+        <%= render "govuk_publishing_components/components/radio", {
+          heading: "Update content type",
+          heading_level: 0,
+          heading_size: "s",
+          name: "to",
+          items: conversion_items(@resource),
+        } %>
+
+        <%= render "govuk_publishing_components/components/button", {
+          text: "Update",
+        } %>
+      <% end %>
+    <% else %>
+      <%= render "govuk_publishing_components/components/list", {
+        items: [
+          ( primary_button_for(@edition, skip_fact_check_edition_path(@edition), "Skip fact check") if @edition.fact_check? ),
+          ( link_to("Delete edition #{@edition.version_number}", confirm_destroy_edition_path(@resource), class: "govuk-link gem-link--destructive") if @edition.can_destroy? ),
+        ],
+      } %>
+    <% end %>
   </div>
 </div>
--- a/test/functional/editions_controller_test.rb
+++ b/test/functional/editions_controller_test.rb
@@ -264,6 +264,50 @@ class EditionsControllerTest < ActionController::TestCase
     end
   end
 
+  context "#duplicate" do
+    setup do
+      @answer = FactoryBot.create(:answer_edition)
+      @help = FactoryBot.create(:help_page_edition)
+    end
+
+    should "redirect to the edit tab of the newly created edition and show success message when user saves successfully" do
+      EditionDuplicator.any_instance.expects(:duplicate).returns(true)
+      EditionDuplicator.any_instance.expects(:new_edition).returns(@help)
+
+      post :duplicate, params: { id: @edition.id, to: "help_page" }
+
+      assert_redirected_to edition_path(@help)
+      assert_equal "New edition created", flash[:success]
+    end
+
+    should "redirect to the edit tab and show a failure message when saving the new edition fails" do
+      EditionDuplicator.any_instance.expects(:duplicate).returns(false)
+
+      post :duplicate, params: { id: @edition.id, to: "help_page" }
+
+      assert_response :found
+      assert_equal "Failed to create new edition: couldn't initialise", flash[:danger]
+    end
+
+    should "redirect to the edit tab and show a failure message when another user has already created a new edition" do
+      FactoryBot.create(:edition, panopticon_id: @edition.artefact.id)
+
+      post :duplicate, params: { id: @edition.id, to: "help_page" }
+
+      assert_response :found
+      assert_equal "Another person has created a newer edition", flash[:warning]
+    end
+
+    should "render the edit tab and show a failure message when there is a service problem" do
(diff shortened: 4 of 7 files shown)


test/functional/editions_controller_test.rb
```diff
@@ -264,6 +264,46 @@ class EditionsControllerTest < ActionController::TestCase
     end
   end
 
+  context "#duplicate" do
+    setup do
+      @answer = FactoryBot.create(:answer_edition, panopticon_id: FactoryBot.create(:artefact).id)
```
mtaylorgds, line 269: I don't believe it's necessary to set the `panopticon_id` for the purposes of these tests.
```suggestion
      @answer = FactoryBot.create(:answer_edition)
```

test/functional/editions_controller_test.rb
```diff
@@ -264,6 +264,46 @@ class EditionsControllerTest < ActionController::TestCase
...
+    end
+
+    should "redirect to edit tab and show success message when user saves successfully" do
+      EditionDuplicator.any_instance.expects(:duplicate).returns(true)
+      EditionDuplicator.any_instance.expects(:new_edition).returns(@answer)
+      post :duplicate, params: { id: @edition.id, to: "help_page" }
+
+      assert_response :found
```
mtaylorgds, line 277: Maybe we should assert on the redirect to match the test name?
```suggestion
      assert_redirected_to edition_path(@answer)
```

test/functional/editions_controller_test.rb
```diff
@@ -264,6 +264,46 @@ class EditionsControllerTest < ActionController::TestCase
     end
   end
 
+  context "#duplicate" do
+    setup do
+      @answer = FactoryBot.create(:answer_edition, panopticon_id: FactoryBot.create(:artefact).id)
+    end
+
+    should "redirect to edit tab and show success message when user saves successfully" do
```
mtaylorgds, line 272: Should we include in the test name that it's the edit tab of the **new edition** that is redirected to?

test/functional/editions_controller_test.rb
```diff
@@ -264,6 +264,46 @@ class EditionsControllerTest < ActionController::TestCase
...
+      EditionDuplicator.any_instance.expects(:duplicate).returns(true)
+      EditionDuplicator.any_instance.expects(:new_edition).returns(@answer)
+      post :duplicate, params: { id: @edition.id, to: "help_page" }
```
mtaylorgds, line 275: Consider adding a blank line above the `post` to separate the "setup" code from the "doing" code (applies to all tests).

mtaylorgds, line 275: I think it's a bit confusing for the reader that we're posting a request to create a help page, but stubbing the `EditionDuplicator` to return an answer.

test/functional/editions_controller_test.rb
```diff
@@ -264,6 +264,46 @@ class EditionsControllerTest < ActionController::TestCase
...
+    should "render the edit tab and show a failure message when another user has already created a new edition" do
+      FactoryBot.create(:edition, panopticon_id: @edition.artefact.id)
+      post :duplicate, params: { id: @edition.id, to: "help_page" }
+
+      assert_template "show"
+      assert_equal "Another person has created a newer edition", flash[:warning]
+    end
+
+    should "render the edit tab and show a failure message when there is a service problem" do
+      EditionDuplicator.any_instance.expects(:duplicate).returns(true)
```
mtaylorgds, line 298: I think this test is passing for the wrong reason. The line below doesn't do anything, as `Edition` doesn't have a `duplicate` method (and if you comment it out, the test still passes). Should this line be stubbing an exception raise instead of returning true?
```suggestion
      EditionDuplicator.any_instance.expects(:duplicate).raises(StandardError)
```

test/integration/edition_edit_test.rb
```diff
@@ -169,6 +170,19 @@ class EditionEditTest < IntegrationTest
         login_as(FactoryBot.create(:user, :govuk_editor, name: "Stub User"))
       end
 
+      %i[draft amends_needed in_review fact_check_received ready archived scheduled_for_publishing].each do |state|
+        context "When state is '#{state}'" do
```
mtaylorgds, line 174: For consistency, we should not start with a capital "W".
```suggestion
        context "when state is '#{state}'" do
```

test/integration/edition_edit_test.rb
```diff
@@ -253,6 +267,56 @@ class EditionEditTest < IntegrationTest
...
+            assert page.has_no_selector?(".gem-c-radio input[value='answer']")
+            assert page.has_selector?(".gem-c-radio input[value='completed_transaction']")
+            assert page.has_selector?(".gem-c-radio input[value='guide']")
+            assert page.has_selector?(".gem-c-radio input[value='help_page']")
+            assert page.has_selector?(".gem-c-radio input[value='place']")
+            assert page.has_selector?(".gem-c-radio input[value='simple_smart_answer']")
+            assert page.has_selector?(".gem-c-radio input[value='transaction']")
+          end
+
+          should "show explanatory text content for all content types except Guides" do
```
mtaylorgds, line 313: Maybe we can reword this as to me the name here implies that we're checking that the text is displayed for all content types other than guides, rather than we're checking that the text common to all content types is displayed and the guide-specific text is not displayed.

mtaylorgds (commented):
Added a few comments.

app/controllers/editions_controller.rb
```diff
@@ -26,14 +26,35 @@ def index
...
   alias_method :unpublish, :show
   alias_method :admin, :show
 
+  def duplicate
+    command = EditionDuplicator.new(@resource, current_user)
+    target_edition_class_name = "#{params[:to]}_edition".classify if params[:to]
+
+    if !@resource.can_create_new_edition?
+      flash.now[:warning] = "Another person has created a newer edition"
+      render "show"
```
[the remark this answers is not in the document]
davidtrussler: > What's the logic behind some of the branches of the if statement calling `render show`, and some doing a `redirect_to`? The legacy controller seems to always to a redirect, so I'm curious about the reason for changing this.
> 
> This particular line also contradicts the PR comment, which says that it does a redirect in this scenario.

I started with all of those calling `render show` until I found it wasn't displaying the new edition when it was created so reverted to `redirect_to` for that. But I agree it makes more sense to be consistent on this. I have left the final `rescue` instance at the minute, which feels more appropriate (rightly or wrongly).

app/views/editions/secondary_nav_tabs/_admin.html.erb
```diff
@@ -1,12 +1,41 @@
...
+
+      <%= form_for @resource, url: duplicate_edition_path(@resource), method: "post" do |f| %>
+        <%= render "govuk_publishing_components/components/radio", {
+          heading: "Update content type",
+          heading_level: 0,
+          heading_size: "s",
+          name: "to",
+          items: format_conversion_radio_options(@resource).map do |option|
+            {
+              value: option[1],
```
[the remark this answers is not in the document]
davidtrussler: > We seem to be missing the [hint text](https://components.publishing.service.gov.uk/component-guide/radio#with_hint_text_on_radios) on the radio buttons that are in the trello card?

Oh yeah, don't know how that got missed. I've put something in for this that works though there may be a more elegant way to do it.

test/functional/editions_controller_test.rb
davidtrussler: > I think it's a bit confusing for the reader that we're posting a request to create a help page, but stubbing the `EditionDuplicator` to return an answer.

Good point. I think I've sorted that by creating a new (help) edition and stubbing that as the expected return.

app/controllers/editions_controller.rb
```diff
@@ -26,14 +26,35 @@ def index
...
   alias_method :metadata, :show
   alias_method :unpublish, :show
   alias_method :admin, :show
 
+  def duplicate
+    command = EditionDuplicator.new(@resource, current_user)
+    target_edition_class_name = "#{params[:to]}_edition".classify if params[:to]
+
+    if !@resource.can_create_new_edition?
+      flash.now[:warning] = "Another person has created a newer edition"
```
mtaylorgds, line 41: I think if we're doing a redirect, we don't want to use `flash.now`.

app/controllers/editions_controller.rb
```diff
@@ -26,14 +26,35 @@ def index
...
+      redirect_to edition_path(resource)
+    elsif command.duplicate(target_edition_class_name, current_user)
+      new_edition = command.new_edition
+      UpdateWorker.perform_async(new_edition.id.to_s)
+      flash[:success] = "New edition created"
+      redirect_to edition_path(new_edition)
+    else
+      flash.now[:danger] = command.error_message
```
mtaylorgds, line 49: I think if we're doing a redirect, we don't want to use `flash.now`.

app/views/editions/secondary_nav_tabs/_admin.html.erb
```diff
@@ -1,12 +1,42 @@
...
+              text: option[0],
+              hint_text: radio_options_hints[option[1]],
```
mtaylorgds, line 24: Would it make more sense/be simpler to have `format_conversion_radio_options` return the list of items, rather than call a method that we then have to loop through and then call another method? If we do that we should probably rename the method (e.g. `conversion_items`).

test/integration/edition_edit_test.rb
```diff
@@ -253,6 +267,56 @@ class EditionEditTest < IntegrationTest
...
+            assert page.has_no_selector?(".gem-c-radio input[value='answer']")
+            assert page.has_selector?(".gem-c-radio input[value='completed_transaction']")
+            assert page.has_selector?(".gem-c-radio input[value='guide']")
+            assert page.has_selector?(".gem-c-radio input[value='help_page']")
+            assert page.has_selector?(".gem-c-radio input[value='place']")
+            assert page.has_selector?(".gem-c-radio input[value='simple_smart_answer']")
+            assert page.has_selector?(".gem-c-radio input[value='transaction']")
+          end
+
+          should "show common explanatory text for all content types and not show explanatory text spedific to Guides" do
```
mtaylorgds, line 313: Typo—"spedific".

app/views/editions/secondary_nav_tabs/_admin.html.erb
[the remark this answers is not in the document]
mtaylorgds: Have suggested an alternative to simplify the code a bit.

app/helpers/editions_helper.rb
```diff
@@ -19,11 +19,28 @@ def legacy_resource_form(resource, &form_definition)
     nested_form_for resource, as: :edition, url: edition_path(resource), html: html_options, &form_definition
   end
 
-  def format_conversion_select_options(edition)
+  def legacy_format_conversion_select_options(edition)
     possible_target_formats = Edition.convertible_formats - [edition.artefact.kind]
     possible_target_formats.map { |format_name| [format_name.humanize, format_name] }
   end
 
+  def format_conversion_radio_options(edition)
```
mtaylorgds, line 27: Rather than have these two methods (`format_conversion_radio_options` and `radio_options_hints`), consider having just the one that returns the list of items (i.e. an array `[{value: "", text: "", hint_text: ""}…]`).

mtaylorgds (commented):
A few more comments.

app/helpers/editions_helper.rb
davidtrussler: Right, that would be more elegant. I think I have understood and updated accordingly.

app/views/editions/secondary_nav_tabs/_admin.html.erb
```diff
@@ -1,12 +1,42 @@
...
+        <p class="govuk-body">All parts of Guide Editions will be copied across. If the format you are converting to doesn't have parts, the content of all the parts will be copied into the body, with the part title displayed as a top-level sub-heading.</p>
+      <% end %>
+
+      <%= form_for @resource, url: duplicate_edition_path(@resource), method: "post" do %>
+        <%= render "govuk_publishing_components/components/radio", {
+          heading: "Update content type",
+          heading_level: 0,
+          heading_size: "s",
+          name: "to",
+          items: conversion_items(@resource).map do |text, value, hint|
```
mtaylorgds, line 20: I think we can simplify this even more by having `conversion_items` return the items in the format that the radio component is expecting so this line becomes
```suggestion
          items: conversion_items(@resource)
```

app/helpers/editions_helper.rb
```diff
@@ -19,11 +19,26 @@ def legacy_resource_form(resource, &form_definition)
...
+      "completed_transaction" => "Done page for end of a service",
+      "guide" => "Multi-page guidance",
+      "help_page" => "Info about GOV.UK website, for example Privacy",
+      "place" => "Postcode look-up for places/services near you",
+      "simple_smart_answer" => "Simple questions and answers that route users to relevant outcomes",
+      "transaction" => "Start page for a service",
+    }
+
+    possible_target_formats = Edition.convertible_formats - [edition.artefact.kind]
+    possible_target_formats.map { |format_name| [format_name.humanize, format_name, radio_options_hints[format_name]] }
```
mtaylorgds, line 39: If we update what's returned here, we can avoid the caller needing to do a `map` call.
```suggestion
    possible_target_formats.map do |format_name|
      {
        value: format_name,
        text: format_name.humanize,
        hint_text: radio_options_hints[format_name],
      }
```

mtaylorgds (commented):
One minor further improvement.

app/views/editions/secondary_nav_tabs/_admin.html.erb
davidtrussler: Ah right, nice, looks much better.

mtaylorgds (approved):
LGTM

PR #2440, opened 2024-11-27
Move secrets to credentials to stop secrets deprecation warning

syed-ali-tw:
Rails 7 has begun to deprecate Rails.application.secrets in favour of Rails.application.credentials, this would ideally need to be fixed by getting secrets from environment variables probably through app config though its a bigger change that we intend to do and for now we have taken this approach of merging secrets with credentials and then just using credentials to stop the noise around deprecation warnings.


baisa (approved):
Helpful change, thank you @syed-ali-tw

PR #2449, opened 2024-11-29
(703) Add Content Block guidance to Whitehall in non-production environments

pezholio:
https://trello.com/c/UZJt5Je1/703-update-mainstream-publisher-to-allow-insert-journey

To allow user testing of the "insert" journey for Content Modelling, we need to add a link to the Content Block Manager in the sidebar when editing content. This is enabled by a feature flag, which is set to true by default in Staging/Integration.

## Screenshot

image


davidtrussler (approved):
LGTM! 👍

PR #2488, opened 2025-01-08
Update access and permissions rake task

mtaylorgds:
Update the rake task for adding owning organisations to publications so that the "updated_at" field is not updated, which given published and archived editions are being modified will avoid potential confusion of them looking like they were updated after publishing/archival.


Tetrino (approved):
Sensible way to avoid the update loop.

PR #2508, opened 2025-01-23
Update publications filter search box

mtaylorgds:
Update the search box on the Design System-ed publications (root) page filter to search by title and slug, instead of just searching the title. This brings the behaviour more inline with how it works on the Bootstrap UI.

Also update some labels on the Design System-ed publications (root) page filter, following a review by the team.

[Trello card](https://trello.com/c/Ur6nAo1J)

### DIFF
--- a/app/presenters/filtered_editions_presenter.rb
+++ b/app/presenters/filtered_editions_presenter.rb
@@ -3,17 +3,15 @@
 class FilteredEditionsPresenter
   ITEMS_PER_PAGE = 20
 
-  def initialize(user, states_filter: [], assigned_to_filter: nil, content_type_filter: nil, title_filter: nil, page: nil)
+  attr_reader :search_text
+
+  def initialize(user, states_filter: [], assigned_to_filter: nil, content_type_filter: nil, search_text: nil, page: nil)
     @user = user
     @states_filter = states_filter || []
     @assigned_to_filter = assigned_to_filter
     @content_type_filter = content_type_filter
-    @title_filter = title_filter
+    @search_text = search_text
     @page = page
-  end
-
-  def title
-    @title_filter
   end
 
   def content_types
@@ -68,7 +66,7 @@ private
     result = editions_by_content_type
     result = apply_states_filter(result)
     result = apply_assigned_to_filter(result)
-    result = apply_title_filter(result)
+    result = apply_search_text(result)
     result = result.accessible_to(user)
     result = result.where.not(_type: "PopularLinksEdition")
     result.order_by(%w[updated_at desc]).page(@page).per(ITEMS_PER_PAGE)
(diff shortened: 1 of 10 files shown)


app/presenters/filtered_editions_presenter.rb
```diff
@@ -3,19 +3,17 @@
 class FilteredEditionsPresenter
   ITEMS_PER_PAGE = 20
 
-  def initialize(user, states_filter: [], assigned_to_filter: nil, content_type_filter: nil, title_filter: nil, page: nil)
+  attr_reader :search_text
```
davidtrussler, line 6: @mtaylorgds Just a query really: can you explain the addition of this line here (as opposed to changing `title_filter` to `search_text` below)? I mean I can see it breaks and I guess it's because it's in a private method but I don't see the difference to how it was before.

test/integration/root_overview_test.rb
```diff
@@ -67,6 +67,20 @@ class RootOverviewTest < IntegrationTest
     assert page.has_no_content?("YYY")
   end
 
+  should "filter by slug" do
```
syed-ali-tw, line 70: just wondering, if we could now say something like 'search by slug' instead and same for title.

test/integration_test_helper.rb
```diff
@@ -33,14 +33,14 @@ def login_as(user)
   def filter_by_user(option, from: "Assigned to")
     within ".publications-filter form" do
       select(option, from:)
-      click_on "Update filter"
+      click_on "Search"
     end
   end
 
   def filter_by_content(substring)
```
syed-ali-tw, line 40: similar to the above comment, shall we call something on the line of "search_by"?

app/presenters/filtered_editions_presenter.rb
mtaylorgds: Originally there was a private readonly attribute, `:title_filter`, and a `title` method (which was effectively a public readonly attribute). I renamed `:title_filter` to `:search_text` and it made sense to me to just make the `:search_text` attribute public rather than have a private attribute and a public method.

Does that make sense? Happy to jump on a quick call if it doesn't.

test/integration/root_overview_test.rb
mtaylorgds: Have updated the "filter by slug" and "filter by title" tests to use "search by".

test/integration_test_helper.rb
mtaylorgds: Have renamed to "search_by_title_or_slug".

davidtrussler (approved):
LGTM! 👍

app/presenters/filtered_editions_presenter.rb
davidtrussler: Yes, I think that makes sense! 👍

PR #2513, opened 2025-01-27
(Content modelling/785) Show content block updates in Mainstream

pezholio:
Trello card: https://trello.com/c/4UrVJbKH/785-show-content-block-updates-in-mainstream

This updates the history component to list when content blocks being used by a particular artefact have been updated.

This works by calling the Publishing API to find all events of a `HostContentUpdateJob` type for the artefact's content ID. We then use the new Signon API to fetch user details for each author and marshal this into a list of `HostContentUpdateEvent` objects

We call this method each time we load an edition, when showing history entries for each previous / subsequent edition, we check if an event took place during the lifecycle of this edition by checking if it occurred after an edition was published and (optionally) before the next edition was published.

We have a similar approach in Whitehall (see https://github.com/alphagov/whitehall/pull/9754) which is now live and in production.

We may need to iterate some elements of the content design, but this puts all the various pieces into play for us to iterate on.

## Screenshot

image

### DIFF
--- a/app/lib/services.rb
+++ b/app/lib/services.rb
@@ -4,7 +4,14 @@ module Services
   def self.publishing_api
     @publishing_api ||= GdsApi::PublishingApi.new(
       Plek.find("publishing-api"),
-      bearer_token: ENV["PUBLISHING_API_BEARER_TOKEN"] || "example",
+      bearer_token: ENV.fetch("PUBLISHING_API_BEARER_TOKEN", "example"),
+    )
+  end
+
+  def self.signon_api
+    @signon_api ||= GdsApi::SignonApi.new(
+      Plek.find("signon", external: true),
+      bearer_token: ENV.fetch("SIGNON_API_BEARER_TOKEN", "example"),
     )
   end
 end
--- a/app/models/host_content_update_event.rb
+++ b/app/models/host_content_update_event.rb
@@ -5,7 +5,7 @@ class HostContentUpdateEvent < Data.define(:author, :created_at, :content_id, :c
       action: "HostContentUpdateJob",
     })
 
-    user_uuids = events.map { |event| event["payload"]["source_block"]["updated_by_user_uid"] }
+    user_uuids = events.map { |event| event["payload"]["source_block"]["updated_by_user_uid"] }.uniq
     users = users_with_uuids(user_uuids)
 
     events.map do |event|
@@ -23,5 +23,15 @@ class HostContentUpdateEvent < Data.define(:author, :created_at, :content_id, :c
     Services.signon_api.get_users(uuids:).map { |user|
       [user["uid"], Author.new(user["name"], user["email"])]
     }.to_h
+  end
+
+  def is_for_edition?(edition)
+    if edition.published?
+      created_at.after?(edition.published_at)
+    elsif edition.archived? && edition.superseded_at
+      created_at.between?(edition.published_at, edition.superseded_at)
+    else
+      false
+    end
   end
 end
--- a/app/models/host_content_update_event.rb
+++ b/app/models/host_content_update_event.rb
@@ -34,4 +34,8 @@ class HostContentUpdateEvent < Data.define(:author, :created_at, :content_id, :c
       false
     end
   end
+
+  def to_action
+    Action.new(self)
+  end
 end
--- a/app/models/host_content_update_event.rb
+++ b/app/models/host_content_update_event.rb
@@ -1,0 +1,27 @@
+class HostContentUpdateEvent < Data.define(:author, :created_at, :content_id, :content_title, :document_type)
+  Author = Data.define(:name, :email)
+  def self.all_for_artefact(artefact)
+    events = Services.publishing_api.get_events_for_content_id(artefact.content_id, {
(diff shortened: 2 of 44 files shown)


app/models/host_content_update_event.rb
```diff
@@ -0,0 +1,45 @@
...
+        created_at: Time.zone.parse(event["created_at"]),
+        content_id: event["payload"]["source_block"]["content_id"],
+        content_title: event["payload"]["source_block"]["title"],
+        document_type: humanize_document_type(event["payload"]["source_block"]["document_type"]),
+      )
+    end
+  end
+
+  def self.users_with_uuids(uuids)
+    Services.signon_api.get_users(uuids:).map { |user|
```
pezholio, line 23: This needs the Signon API key to be present in Publisher - I've already done this (see https://github.com/alphagov/govuk-helm-charts/pull/2946 and https://github.com/alphagov/govuk-helm-charts/pull/2947). Confirmed this is working in Integration.

mtaylorgds: The app falls over when running locally on this call.

mtaylorgds (changes requested):
Not done a proper review yet—just tried to run locally and hit an error, so thought I'd raise if for you to look into.

pezholio:
I've got a PR for signon that should fix this https://github.com/alphagov/signon/pull/3579

mtaylorgds:
> I've got a PR for signon that should fix this [alphagov/signon#3579](https://github.com/alphagov/signon/pull/3579)

We'll also need to add `signon-app` as a dependency in publisher's `docker-compose.yml` in `govuk-docker`.

test/integration/host_content_update_history_test.rb
```diff
@@ -0,0 +1,143 @@
...
+      some_time_passes
+      @edition2_content_update_event_1 = create_content_update_event(
+        updated_by_user_uid: user1["uid"],
+      )
+
+      some_time_passes
+      @edition3 = edition2.build_clone(GuideEdition)
+      @edition3.save!
+
+      @edition3.new_action(@author, Action::NEW_VERSION)
```
mtaylorgds, line 80: Do we need this action? If we do, should be recording a similar one for the other editions?

test/unit/helpers/admin/action_helper_test.rb
```diff
@@ -58,4 +58,55 @@ class ActionHelperTest < ActionView::TestCase
...
+      edition_actions[6],
+      edition_actions[3],
+      edition_actions[1],
+      edition_actions[2],
+      edition_actions[0],
+    ]
+
+    result = edition_actions(edition, update_events)
+
+    assert_equal result, expected_actions
```
mtaylorgds, line 110: I believe the arguments to `assert_equal` are the wrong way round.
```suggestion
    assert_equal expected_actions, result
```

mtaylorgds (commented):
Generally looks good to me—I've added a few comments/suggestions/questions.

We shouldn’t merge this PR until both:
- The [PR in signon](https://github.com/alphagov/signon/pull/3579) has been merged.
- We've updated publisher’s `docker-compose.yml` in `govuk-docker` to add `signon-app` as a dependency.

app/models/host_content_update_event.rb
```diff
@@ -0,0 +1,45 @@
...
+      [user["uid"], Author.new(user["name"], user["email"])]
+    }.to_h
+  end
+
+  def self.humanize_document_type(document_type)
+    document_type.delete_prefix("content_block_").humanize
+  end
+
+  def is_for_edition?(edition)
```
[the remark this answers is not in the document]
pezholio: Yep, that's right. It's basically like how a listing page gets updated when linked documents get updated, so it's completely separate from the editioning process. I'm going to write up some proper docs in the coming weeks.

app/lib/services.rb
```diff
@@ -7,4 +7,11 @@ def self.publishing_api
       bearer_token: ENV["PUBLISHING_API_BEARER_TOKEN"] || "example",
     )
   end
+
+  def self.signon_api
+    @signon_api ||= GdsApi::SignonApi.new(
+      Plek.find("signon", external: true),
+      bearer_token: ENV.fetch("SIGNON_API_BEARER_TOKEN", "example"),
```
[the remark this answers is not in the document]
pezholio: Yeah, good call - `ENV.fetch` basically does the same under to hood, but is a bit more specific to Env vars and (to me, anyway) a neater way to provide defaults

test/integration/host_content_update_history_test.rb
pezholio: Nah, I've got rid 👍

pezholio:
@mtaylorgds I've addressed all those changes, and got a PR for `govuk-docker` here https://github.com/alphagov/govuk-docker/pull/816 👍

mtaylorgds (commented):
Changes look good to me, @baisa is going to also take a look today just to check she's all happy before we officially approve it.

baisa (approved):

mtaylorgds (approved):

PR #2517, opened 2025-01-28
Add a check for departmental_editor? permission for user

baisa:
[Trello card](https://trello.com/c/HQmmcxJl/599-access-permissions-add-departmentaleditor-permission)

This is to enable specifying restriction to access Mainstream content by departments. It is an addition to giving access by matching owning_org_content_ids with users organisation_content_id, to ensure we do not block access for some departments.

### DIFF
--- a/test/functional/editions_controller_test.rb
+++ b/test/functional/editions_controller_test.rb
@@ -122,14 +122,22 @@ class EditionsControllerTest < ActionController::TestCase
   end
 
   context "when 'restrict_access_by_org' feature toggle is disabled" do
-    %i[show metadata history admin related_external_links unpublish].each do |action|
+    %i[show metadata history related_external_links].each do |action|
       context "##{action}" do
         setup do
           @edition = FactoryBot.create(:edition, owning_org_content_ids: %w[org-two])
         end
 
-        should "return a 200 when requesting an edition owned by a different organisation" do
-          login_as(FactoryBot.create(:user, :govuk_editor, organisation_content_id: "org-one"))
+        should "return a 200 when requesting the #{action} tab on an edition owned by a different organisation and user has departmental_editor permission" do
+          login_as(FactoryBot.create(:user, :departmental_editor, organisation_content_id: "org-one"))
+
+          get action, params: { id: @edition.id }
+
+          assert_response :ok
+        end
+
+        should "return a 200 when requesting the #{action} tab on an edition owned by a different organisation and user does not have departmental_editor permission" do
+          login_as(FactoryBot.create(:user, organisation_content_id: "org-one"))
 
           get action, params: { id: @edition.id }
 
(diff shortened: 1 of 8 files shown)


test/functional/editions_controller_test.rb
mtaylorgds, line 1: I think we should also update the test on line 55 ("return a 200 when requesting an edition owned by a different organisation") to log in with a user that has the "departmental_editor" permission. Otherwise we're not really testing that the restriction isn't being implemented when the feature toggle is disabled.

test/functional/legacy_editions_controller_test.rb
mtaylorgds, line 1: Similarly, I think we should also update the test on line 1318 ("return a 200 when requesting an edition owned by a different organisation") to log in with a user that has the "departmental_editor" permission. Otherwise we're not really testing that the restriction isn't being implemented when the feature toggle is disabled.

test/functional/legacy_editions_controller_test.rb
```diff
@@ -1343,8 +1343,8 @@ class LegacyEditionsControllerTest < ActionController::TestCase
           @edition = FactoryBot.create(:edition, owning_org_content_ids: %w[org-two])
         end
 
-        should "return a 404 when requesting an edition owned by a different organisation" do
-          login_as(FactoryBot.create(:user, :govuk_editor, organisation_content_id: "org-one"))
+        should "return a 404 when requesting an edition owned by a different organisation and user has departmental_editor permission" do
+          login_as(FactoryBot.create(:user, :departmental_editor, organisation_content_id: "org-one"))
```
mtaylorgds, line 1347: Maybe we should also now have a test that a 200 is returned when owned by a different organisation, but the user doesn't have the "departmental_editor" permission.

test/unit/presenters/filtered_editions_presenter_test.rb
```diff
@@ -160,8 +160,8 @@ def a_gds_user
         test_strategy.switch!(:restrict_access_by_org, false)
       end
 
-      should "filter out editions not accessible to the user" do
-        user = FactoryBot.create(:user, organisation_content_id: "an-org")
+      should "filter out editions not accessible to the user with departmental_editor permission" do
```
mtaylorgds, line 163: I'm actually not convinced we need to change this test name.

mtaylorgds (commented):
I think we could do with some more tweaks to the tests.

test/functional/editions_controller_test.rb
```diff
@@ -46,14 +46,22 @@ class EditionsControllerTest < ActionController::TestCase
...
-          login_as(FactoryBot.create(:user, :govuk_editor, organisation_content_id: "org-one"))
+        should "return a 200 when requesting the #{action} tab on an edition owned by a different organisation and user has departmental_editor permission" do
+          login_as(FactoryBot.create(:user, :departmental_editor, organisation_content_id: "org-one"))
+
+          get action, params: { id: @edition.id }
+
+          assert_response :ok
+        end
+
+        should "return a 200 when requesting the #{action} tab on an edition owned by a different organisation and user does not have departmental_editor permission" do
```
mtaylorgds, line 63: I think this second "return a 200" should probably be in the feature toggle enabled context (to demonstrate the difference that the `departmental_editor` permission has when the toggle is enabled).

test/functional/legacy_editions_controller_test.rb
```diff
@@ -1315,7 +1315,15 @@ class LegacyEditionsControllerTest < ActionController::TestCase
...
-        should "return a 200 when requesting an edition owned by a different organisation" do
+        should "return a 200 when requesting the #{action} tab on an edition owned by a different organisation and user has departmental_editor permission" do
+          login_as(FactoryBot.create(:user, :departmental_editor, organisation_content_id: "org-one"))
+
+          get action, params: { id: @edition.id }
+
+          assert_response :ok
+        end
+
+        should "return a 200 when requesting the #{action} tab on an edition owned by a different organisation and user does not have departmental_editor permission" do
```
mtaylorgds, line 1326: Likewise, I think this second "return a 200" should probably be in the feature toggle enabled context (to demonstrate the difference that the departmental_editor permission has when the toggle is enabled).

mtaylorgds (approved):
LGTM

mtaylorgds (approved):

PR #2519, opened 2025-01-28
Replace title component with heading

andysellick:

## What
Replace the title component with the heading component.

- should still be a h1 and visually identical
- heading has no margin top, so matches the previous margin_top: 0 option on title

Confession: I've not actually run this locally to check it visually, but I've done this in a lot of applications now and I'm reasonably confident it will work.

## Why

The title component is being decommissioned and we've already removed the margin_top option from it as part of some other work. This means that with the latest changes a big space will suddenly appear above it. But if we switch it to the heading component, it looks exactly the same and won't have this problem.

## Visual changes
None.

Trello card: https://trello.com/c/Y0pDWbHw/390-remove-margintop-option-from-shared-helper


AshGDS (approved):
LGTM

PR #2539, opened 2025-02-12
Prevent updated_at being changed when edition is archived

Tetrino:
This should fix most causes of behaviour we are seeing wherein the public_updated_at field is being set incorrectly in some circumstances, causing the documents to become visible in various time based search results and potentially causing confusion for the user.

This function is only called by edition when archiving previous versions of itself, and the fact check message processor. The latter does not care for the time stamp and thus this should be a safe change.


mtaylorgds:
I worry that although nothing in code might rely on the `updated_at` field changing when a fact check is received, users might expect it to be updated.

Tetrino:
> I worry that although nothing in code might rely on the `updated_at` field changing when a fact check is received, users might expect it to be updated.

This will have to be a content discussion I feel.

app/models/workflow.rb
```diff
@@ -170,6 +170,16 @@ def error_description
...
+    # http://rubydoc.info/github/pluginaweek/state_machine/StateMachine/Machine:event
+    # pass false to transition state without performing state machine actions
+    public_send(event, false)
+    # Set to not update the updated_at field as it causes issues with public_updated_at
+    # See https://github.com/alphagov/publisher/pull/2539
+    save!(validate: false, touch: false)
+  end
+
+  # As above, but retains change to updated_at field for cases where it's expected
+  def perform_timestamped_event_without_validations(event)
```
mtaylorgds, line 182: To me, it feels like updating the `updated_at` field should be the default, and you call a special method if you don't want that (even though in this case there'll be two calls to the "special case" method and only one to the "normal" one).

So I'd be tempted to keep `perform_event_without_validations` unchanged (and still called by `ReceiveFactCheckProcessor`), and have the new method be something like `perform_event_without_validations_or_updating_timestamps` and update the calls in `Artefact#archive_editions` and `Edition#was_published` to call the new method.

Alternatively, we could add an extra boolean parameter to `Workflow#perform_event_without_validations` that gets passed through to the `save!` call.

Tetrino: I'm always hesitant to add new parameters so I'll swap the functions. The reason why I kept it as I did was in case of any other areas we hadn't found.

But yes I'll swap the functions and make changes up the trace.

mtaylorgds (commented):
It's just occurred to me that we haven't added any tests for the new behaviour—do you think that would be possible, or is it too difficult a situation to recreate in tests?

Tetrino:
> It's just occurred to me that we haven't added any tests for the new behaviour—do you think that would be possible, or is it too difficult a situation to recreate in tests?

I did attempt to but behaviour in the test world seems different from the real world in this case, thus I came to the conclusion that it's too unreliable to consistently test.

mtaylorgds (approved):

PR #2555, opened 2025-02-24
Move Users table to postgres

syed-ali-tw:
Move User table to postgres and support any existing relation with other models using temporary foreign key.
All the temporary code changes are earmarked with comments to allow assistance when we would undo them.

Note: there is a test failure on test/models/edition_test.rb: should "have a database-level constraint on the uniqueness"

This has been hard to debug, the test passes locally and will change in near future, hence investing any more time trying to fix it might not be worth it.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -3,40 +3,11 @@ require "cgi"
 require "gds-sso/user"
 require_dependency "safe_html"
 
-class User
-  include Mongoid::Document
-  include Mongoid::Timestamps
+class User < ApplicationRecord
   include GDS::SSO::User
 
-  # Let an app configure the symbolized collection name to use,
-  # e.g. set a constant in an initializer.
-  if defined?(USER_COLLECTION_NAME)
-    store_in collection: USER_COLLECTION_NAME.to_sym
-  else
-    store_in collection: :users
-  end
-
-  field "name",                    type: String
-  field "uid",                     type: String
-  field "version",                 type: Integer
-  field "email",                   type: String
-  field "permissions",             type: Array, default: []
-  field "remotely_signed_out",     type: Boolean, default: false
-  field "organisation_slug",       type: String
-  field "disabled",                type: Boolean, default: false
-  field "organisation_content_id", type: String
-
-  index({ uid: 1 }, unique: true)
-  index disabled: 1
-
-  scope :alphabetized, -> { order_by(name: :asc) }
-  scope :enabled,
-        lambda {
-          any_of(
-            { :disabled.exists => false },
-            { :disabled.in => [false, nil] },
-          )
-        }
+  scope :alphabetized, -> { order(name: :asc) }
+  scope :enabled, -> { where("disabled IS NULL OR disabled = ?", false) }
 
   def to_s
     name || email || ""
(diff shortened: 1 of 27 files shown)


Tetrino (approved):
Looks good to me - only comment is to check that the default behaviour of `order` is ascending to match previous behaviour, if it matters in this case.

app/models/user.rb
```diff
@@ -3,40 +3,15 @@
...
+class User < ApplicationRecord
   include GDS::SSO::User
 
-  # Let an app configure the symbolized collection name to use,
-  # e.g. set a constant in an initializer.
-  if defined?(USER_COLLECTION_NAME)
-    store_in collection: USER_COLLECTION_NAME.to_sym
-  else
-    store_in collection: :users
+  def self.using_object_ids?
```
mtaylorgds, line 9: What's this method for? It doesn't seem to be referenced anywhere…

test/unit/helpers/admin/publications_table_helper_test.rb
```diff
@@ -180,19 +180,19 @@ class PublicationsTableHelperTest < ActionView::TestCase
         :edition,
         state: "in_review",
         review_requested_at: "2024-07-12",
-        assigned_to_id: "66911dbf2c88ee0001d8af62",
```
mtaylorgds, line 183: Should we be consistent with the test below and assign the edition to a different user that is created in the test?

app/lib/govuk_content_models/action_processors/assign_processor.rb
```diff
@@ -2,8 +2,10 @@ module GovukContentModels
   module ActionProcessors
     class AssignProcessor < BaseProcessor
       def process
-        edition.set(assigned_to_id: action_attributes[:recipient_id])
+        # rubocop:disable Rails/SaveBang
+        edition.update(assigned_to_id: action_attributes[:recipient_id])
```
mtaylorgds, line 6: As discussed, this change will cause validations and callbacks to trigger (for example, the `updated_at` value will get updated now).

app/models/workflow.rb
```diff
@@ -191,6 +202,17 @@ def important_note
     action if action.try(:request_type) == Action::IMPORTANT_NOTE
   end
 
+  # Temp-to-be-removed
+  # The method below are getters and setters for assigned_to that allows us to set the assigned_to_id and get assigned_to.
+  # We are unable to use belongs_to :assigned_to, class_name: "User as the User table is
```
mtaylorgds, line 207: Missing a closing quote
```suggestion
  # We are unable to use belongs_to :assigned_to, class_name: "User" as the User table is
```

app/models/workflow.rb
```diff
@@ -191,6 +202,17 @@ def important_note
...
+  # in postgres and using a combination of setter and getter methods with a assigned_to_id field
+  # to be able to achieve congruent result as  having a belongs to while we are moving other table to postgres
+  delegate :assigned_to_id=, to: self
```
mtaylorgds, line 210: I don't quite understand this line—how does delegating to `self` work?

syed-ali-tw: it is similar to saying 
def assigned_to_id=(id)
self[:assigned_to_id] = id
end

though it makes more sense to keep it similar to other changes.

mtaylorgds (approved):

baisa (approved):

PR #2559, opened 2025-02-27
Edit page for published editions

mtaylorgds:
[Trello card](https://trello.com/c/5EuyRleF)

Display a published edition's content as read-only on the "edit" tab when viewing an edition.

Also implements a sidebar, allowing users to view a later draft edition (if one exists), or create a new draft edition (if one doesn't already exist), and also view the published page on GOV.UK.

|Scenario|View|
|-|-|
|Viewing a published edition before the changes in this PR|published-view-before|
|Viewing a published edition when a draft edition does not exist|published-view-after-no-later-draft|
|Viewing a published edition when a draft edition exists|published-view-after-with-later-draft|

### DIFF
--- a/app/helpers/editions_sidebar_buttons_helper.rb
+++ b/app/helpers/editions_sidebar_buttons_helper.rb
@@ -10,4 +10,15 @@ module EditionsSidebarButtonsHelper
                end
     buttons << link_to("View on GOV.UK (opens in new tab)", view_homepage_path, rel: "noreferrer noopener", target: "_blank", class: "govuk-link")
   end
+
+  def non_published_sidebar_buttons(edition)
+    buttons = []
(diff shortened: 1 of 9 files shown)


davidtrussler (approved):
LGTM 👍 
Just a question (shouldn't hold up merging if it's not a thing): do we need the commented line at the top of the helper (`# frozen_string_literal: true`)?

mtaylorgds:
> LGTM 👍 Just a question (shouldn't hold up merging if it's not a thing): do we need the commented line at the top of the helper (`# frozen_string_literal: true`)?

It's something that RubyMine adds to new files. From what I can read about [it is a minor performance improvement](https://bugs.ruby-lang.org/issues/20205) and prevents strings being mutated (which they generally really shouldn't be).

PR #2560, opened 2025-02-27
Stub UpdateWorker#perform_async in some tests

mtaylorgds:
Stubbing this call out prevents attempts to use Sidekiq, and thus prevents output like the below appearing in the console when running these tests, which makes the test output noisy.

```
2025-02-27T14:18:03.340Z pid=1 tid=5vp INFO: Sidekiq 7.3.9 connecting to Redis with options {:size=>10, :pool_name=>"internal", :url=>"redis://publisher-redis", :reconnect_attempts=>[0.05, 0.25, 1, 5]}
```

This commit only applies the change to the `edition_edit_test.rb` integration test. We should look to do this in a test helper to avoid similar issues when running other tests.


syed-ali-tw (approved):
LGTM

syed-ali-tw (approved):

PR #2593, opened 2025-03-14
Log selenium webdriver and chrome versions

richardTowers:


richardTowers:
The chrome version is already logged by failing tests, and I can't see the rails logger output in the GHA test output. So I think this isn't worth pursuing.

PR #2594, opened 2025-03-14
Pin the chrome version in our minitest CI workflow

mtaylorgds:
There appears to be a bug in the latest version (v134) that is causing a lot of test flakiness. Pin to an older version of Chrome that does not have this issue.


mtaylorgds:
This PR is superseded by #2596

PR #2596, opened 2025-03-14
Pin chrome version for integration tests

syed-ali-tw:
Use an older version of chrome for integration test runs as there is an ongoing 
issue with more recent versions which is causing random test failures


mtaylorgds (approved):
Great work!

chao-xian:
Said issue https://github.com/teamcapybara/capybara/issues/2800

PR #2601, opened 2025-03-18
Simplify chrome version pinning in minitest action

mtaylorgds:
Remove the use of a GitHub action from an unverified publisher, and just
 use the "setup-chrome" action from a verified one.

Also pins the version of the action referenced to a specific sha as a
 security precaution, as recommended by the [developer docs](https://gds-way.digital.cabinet-office.gov.uk/standards/source-code/use-github.html#using-github-actions-and-workflows).


syed-ali-tw (approved):
LGTM

PR #2603, opened 2025-03-18
Undo the Chrome version pinning in CI

mtaylorgds:
This was introduced as a temporary measure until the issues with the latest version of Chrome have been resolved.


Tetrino (approved):
LGTM

PR #2607, opened 2025-03-20
624 edit page archived edition

davidtrussler:
[Trello](https://trello.com/c/oOBPngB2/624-edit-page-for-archived-edition-answer-and-help)

This implements the Edit page for an 'Archived' edition. Essentially it displays the data in read only mode rather than in form fields, using the same headings: 
- Edit
- Title
- Meta tag description
- Is this beta content?
- Body
- Public change note

Additional requirements are
- The 'Unpublish' tab should not be displayed in this scenario. This has already been implemented in a previous piece of work so there are no changes for that requirement here. 
- A message should be displayed to the user when all editions have been unpublished

Screenshot 2025-03-20 at 10 52 42

### DIFF
--- a/app/views/editions/secondary_nav_tabs/_edit.html.erb
+++ b/app/views/editions/secondary_nav_tabs/_edit.html.erb
@@ -9,7 +9,7 @@
 </div>
 
 <div class="govuk-grid-row">
-  <% if @resource.published? %>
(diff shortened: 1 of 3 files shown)


mtaylorgds (approved):
LGTM!

PR #2612, opened 2025-03-21
show Content Block update events on in progress Hosts

Harriethw:
https://trello.com/c/RxmJ0fe2/977-allow-updates-to-objects-to-appear-in-the-history-and-notes-for-mainstream-draft-editions-story

Previously an event would only show in the
Edition's "history and notes" if the event took
place after the Edition had been published (and if
it was not archived/superseded).

We now want to display event if it took place at
any time after an Edition's creation, IF it is not
archived.

---

### DIFF
--- a/app/models/host_content_update_event.rb
+++ b/app/models/host_content_update_event.rb
@@ -26,10 +26,14 @@ class HostContentUpdateEvent < Data.define(:author, :created_at, :content_id, :c
   end
 
   def is_for_edition?(edition)
-    if edition.published?
-      created_at.after?(edition.published_at)
-    elsif edition.archived? && edition.superseded_at && edition.published_at
-      created_at.between?(edition.published_at, edition.superseded_at)
+    if edition.archived?
+      if edition.superseded_at.present?
+        created_at.between?(edition.created_at, edition.superseded_at)
+      elsif edition.artefact.state == "archived"
+        created_at.between?(edition.created_at, edition.artefact.updated_at)
+      end
+    elsif edition.in_progress? || edition.published? || edition.scheduled_for_publishing?
+      created_at.after?(edition.created_at)
     else
       false
     end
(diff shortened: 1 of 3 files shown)


app/models/host_content_update_event.rb
```diff
@@ -26,10 +26,10 @@ def self.users_with_uuids(uuids)
   end
 
   def is_for_edition?(edition)
-    if edition.published?
-      created_at.after?(edition.published_at)
-    elsif edition.archived? && edition.superseded_at && edition.published_at
-      created_at.between?(edition.published_at, edition.superseded_at)
+    if edition.archived?
+      created_at.between?(edition.created_at, edition.superseded_at)
```
Harriethw, line 30: I'm assuming here that all archived Editions will have a superseded_at date, but I'm struggling to find the actual logic of how something gets archived in the code.

pezholio: We should probably have a check for `superseded_at` here, as if an edition doesn't have a subsequent edition, then this will error. If there's no `superseded_at` date, then any actions that occur after the `created_at` date will apply to that edition, as it'll be the last one.

Harriethw: In the UI if I understand correctly an Edition only appears as archived once it has been superseded, so I was wanting to understand here if that was the case, or if an Edition could be archived in another way. 
I think we discussed in the meeting that if an Edition has been archived then the updates don't apply to it, because it can never be previewed or published again. I think there might be a way to look at the created_at date of an Artefact or Action to determine when it was archived so will take a look...
I might also have a look at the db in Integration to see if there are any Editions without a `superseded_at` that are archived...

Harriethw: Ah I've just seen that an Edition can be archived if a user goes to "unpublish" it, so it would be possible to not have a `superseded_at`

Harriethw: @pezholio ok I've maybe gone way to in the weeds with this, but have updated the commit message with explanation - although there is a handy Artefact/Action created to set the state to `archived` when an Edition is unpublished, nothing is created when an Edition is superseded, even though it's state is set to `archived`.

pezholio (approved):
LGTM 👍

PR #2617, opened 2025-03-25
Compare editions page

mtaylorgds:
## Trello

[Trello card](https://trello.com/c/aC9DCeOB/607-compare-editions-diff-edit-history-and-notes)

## What
Adds a link to each edition's section (bar the first) on the history and notes tab that links to a page where a diff between that edition and the previous one can be viewed.

## History and notes link
image

## Compare editions page
compare-editions-page


app/assets/stylesheets/diff.scss
davidtrussler, line 1: We should be able to use the `background-color` attribute of `li.del::before` to set the darker colour in the design where it overlaps the grey line here. Likewise for `li.ins::before`.

app/assets/stylesheets/legacy_diff.scss
```diff
@@ -0,0 +1,80 @@
...
+$added-color: #ddffdd;
+$strong-added-color: #77f177;
+$removed-color: #ffdddd;
+$strong-removed-color: #ffaaaa;
+
+.diff {
+  border: 1px solid $gray-lighter;
+  border-left: 40px solid $gray-lighter;
+  border-radius: 3px;
+  padding: 15px;
```
davidtrussler, line 14: We should probably use the [Design System responsive spacing scale](https://design-system.service.gov.uk/styles/spacing/#responsive-spacing) variables for valuables like '40px' and '15px'. This would adapt to viewport sizes and allow for the values to potentially change in future keeping the layout consistent.

app/assets/stylesheets/diff.scss
davidtrussler: I know the styles for the layout of this are largely inherited but they feel a bit fragile to me and full of arbitrary values and could be expressed more elegantly. Not sure it's necessarily worth the effort to do anything about that but the one that is bugging me a bit is the `min-height` on the `<li>` items that doesn't relate to anything. It seems to be there to allow for blank lines but it displays those slightly smaller than lines with content.

mtaylorgds: @davidtrussler I guess the height of the non-blank lines will depend on the font size used—since we're using `govuk-body` this seems to be 25px high.

Are you suggesting we should change the `min-height` to `25px`, or something else?

davidtrussler: I think we can deal with blank lines and stop them collapsing by removing the `min-height` attribute and replacing it with hidden generated content like this: 
```
ul {
  ...
  
  li {
    ...

    &::after {
      content: ".";
      visibility: hidden;
    }
  }
}
```

davidtrussler: Also delete `min-height` from `li.del::before, li.ins::before`

davidtrussler:
All looks great to me except for some super-pedantic observations around the inherited CSS. Let me know if it's not clear but it's not going to make much difference that anyone will notice so let's not worry too much about it!

app/assets/stylesheets/diff.scss
davidtrussler: Yes, `govuk-body` resolves to 25px though it's strictly speaking defined in ems rather than pixels so could vary on some devices. I think adding generated content like above will ensure it always displays at the height of a line of `govuk-body` text across devices.

mtaylorgds: @davidtrussler I guess the line height of non-empty lines will depend on the font size, which given the current styling seems to put it at `25px`. 

Are you suggesting that we change the `min-height` to `25px`, or something else?

davidtrussler: My suggestion is to use something we already use elsewhere (see below) and probably copy that for consistency but apply it to the `<li>`. It's a pretty widely used technique, usually referred to as 'clearfix', that stops vertical blocks of content collapsing. I was trying to see if we had a useful mixin for that but I don't think we do. I think it's better than specifying an explicit height since that can change and is a bit device dependant, particularly if text values are specified in pixels (which they aren't really). Hope that's sort of clear. 

```
.govuk-grid-row::after {
    content: "";
    display: block;
    clear: both;
}
```

davidtrussler (approved):
Nice, LGTM 👍

PR #2620, opened 2025-03-26
624 edit page add custom styles

davidtrussler:
[Trello](https://trello.com/c/oOBPngB2/624-edit-page-for-archived-edition-answer-and-help)

The changes here (somewhat circuitously) updates how `govuk-body` styles are applied to all `<p>` elements within the read-only content on the Edit page for editions in 'Published' and 'Archived' states. This is necessary to remove the browser-applied top margin to these elements, specifically on the 'body' section. We can only do this by applying the `govuk-body` style directly to the `<p>` element rather than to its parent as it is currently implemented. Since this cannot be done in the 'body' section the alternative approach is to apply the style via a custom stylesheet. 

This is achieved by
- removing the `govuk-body` class from elements on the page
- grouping sections together with the `<section>` element (not strictly necessary but feels a semantically more elegant)
- adding a class name to the parent container to reflect the edition's state, e.g `edit--published`
- adding a custom stylesheet to extend. the `govuk-body` class to `<p>` elements within this container for editions whose state is either 'published' or 'archived'.

|Before|After|
|-|-|
|Screenshot 2025-03-26 at 13 02 03|Screenshot 2025-03-26 at 13 02 42|

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -38,6 +38,7 @@ $govuk-page-width: 1140px;
 @import 'govuk_publishing_components/components/textarea';
 @import 'govuk_publishing_components/components/warning-text';
 @import "downtimes";
+@import "edit";
 @import "editions";
 @import "history";
 @import "options_sidebar";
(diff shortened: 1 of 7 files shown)


mtaylorgds (approved):
What a nightmare!

Well figured out; it seems a good solution, given the circumstances.

PR #2626, opened 2025-03-31
Update page structure so row is contained within form

davidtrussler:
This PR makes a small change to the markup structure of the Edit page to fix the sticky sidebar (which currently is not sticky). 

In order to function the CSS attribute `position: sticky` requires the element to which it is applied and its immediate sibling from which its offset position is calculated to both be direct descendants of their "nearest scrolling ancestor", i.e. the parent `govuk-grid-row` element. In this case the `<form>` element is the direct ancestor of both elements. The fix here is to ensure that the `govuk-grid-row` is inside, rather than outside, the `<form>` element. 

Additionally the `edit--<state>` class has been deleted from the row within the `<form>` since this is only required when the content is read-only, i.e. editions whose state is either 'published' or 'archived'.

### DIFF
--- a/app/views/editions/secondary_nav_tabs/_edit.html.erb
+++ b/app/views/editions/secondary_nav_tabs/_edit.html.erb
@@ -8,8 +8,8 @@
   </div>
 </div>
 
-<div class="govuk-grid-row edit--<%= @resource.state %>">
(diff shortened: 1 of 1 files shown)


ellohez (approved):
LGTM! Makes sense moving the edit state.

PR #2627, opened 2025-03-31
Refactor send to 2i page to use partial

ellohez:
This was prompted by a discussion during the Desk Check for [Trello Card]( https://trello.com/c/dQepazi0)
In order to simplify the code, encourage reuse (make it DRY) and to make future formatting changes easier, the Send to 2i page has been adapted to use the same partial as 'Request amendments', 'Skip review' and 'No changes needed' pages. 

Also, in response to the feedback from the 'Send to 2i' desk check: 
- A label has been added to the comment textfield for the Send to 2i page
- The link in the copy above the textfield has been changed to the GOV.UK style

### DIFF
--- a/app/views/editions/secondary_nav_tabs/send_to_2i_page.html.erb
+++ b/app/views/editions/secondary_nav_tabs/send_to_2i_page.html.erb
@@ -4,7 +4,7 @@
 <% content_for :title, "Send to 2i" %>
 
 <%= render "editions/secondary_nav_tabs/progress_with_comment", {
-  form_url: "#{edition_path}/send_to_2i",
+  form_url: send_to_2i_edition_path(@edition),
   text: "Explain what changes you did or did not make and why. Include a link to the relevant Zendesk ticket and Trello card.
     If you’ve added a change note already, you do not need to add another one.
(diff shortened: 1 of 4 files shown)


app/views/editions/secondary_nav_tabs/send_to_2i_page.html.erb
```diff
@@ -3,28 +3,11 @@
...
-        <%= render "govuk_publishing_components/components/button", {
-          text: "Send to 2i",
-        } %>
-        <%= link_to("Cancel", edition_path, class: "govuk-link govuk-link--no-visited-state") %>
-      </div>
-    <% end %>
-  </div>
-</div>
+<%= render "editions/secondary_nav_tabs/progress_with_comment", {
+  form_url: "#{edition_path}/send_to_2i",
```
mtaylorgds, line 7: Looking at this now, I think we should probably be using the path helper method here (and fix the other pages to do likewise):
```suggestion
  form_url: send_to_2i_page_edition_path(@edition),
```

mtaylorgds (commented):
One minor comment, but otherwise looks good!

PR #2618, opened 2025-03-25
604 tab send to 2i link and page

ellohez:
[Trello card](https://trello.com/c/dQepazi0)

The changes have added a **Send to 2i** link, to the side bar, visible from the **Edit** tab

|Scenario|View|
|-|-|
|User visits the **Edit** tab for an edition in draft state and if they have editor permissions then the **Send to 2i** link will be shown in the sidebar.|Edition with **Send to 2i** link in sidebar|
|When the User clicks on the **Send to 2i** button the Send to 2i page is displayed, allowing them to enter a comment and submit the edition to 2i. Clicking cancel takes the user back to the previous Edit tab|Send to 2i page opens showing an empty text area and submit and cancel buttons|
|When the user clicks **Send to 2i** then the user is redirected back to the edit tab and a success message is displayed.|Edit tab with success message is displayed when the edition is successfully sent to 2i|

PR #2628, opened 2025-04-01
Fix sidebar buttons helper to return button array

ellohez:
Applies a fix for the missing sidebar buttons for an in-review edition, caused by [Send to 2i link and page](https://trello.com/c/dQepazi0) https://github.com/alphagov/publisher/pull/2618

Adds tests for an edition in the 'in review' (Sent to 2i) state to hopefully catch any similar issues in future.


mtaylorgds (approved):
LGTM

PR #2633, opened 2025-04-02
Publish now for scheduled edition

mtaylorgds:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #2636, opened 2025-04-03
Fix progress with comment methods

ellohez:
## What

This PR introduces fixes for the 'progress with comment' pages:
- Send to 2i
- Request amendments
- Skip review
- No changes needed

Send to 2i

- The Send to 2i had an issue where the added flash messages would not be displayed and it wasn't rendering/redirecting properly. 

All other pages
- The other pages were 'rendering' rather than 'redirecting' back to the edition which meant that the URL became the edition path with the addition of the method name as a slug. 
- They were only outputting an unhelpful message if the user was able to trigger the page even if the edition wasn't in the correct state for that action. 
- Tests were missing for both of the above instances.


app/controllers/editions_controller.rb
```diff
@@ -130,7 +136,10 @@ def send_to_2i
   def skip_review
     if skip_review_for_edition(@resource, params[:comment])
       flash[:success] = "2i review skipped"
-      redirect_to edition_path(resource) 
+      redirect_to edition_path(resource)
+    elsif !@resource.can_approve_review?
```
mtaylorgds, line 140: Should this check be `can_skip_review?`?

test/integration/edition_edit_test.rb
mtaylorgds, line 1: Nice bit of cleanup 👍

app/controllers/editions_controller.rb
mtaylorgds, line 1: It might be worth doing the `can_*` checks first to avoid attempting to do the transition when we don't need to, like I have done on the ["publish now" PR](https://github.com/alphagov/publisher/pull/2633/files).

ellohez: Ooh, good point!!

PR #2643, opened 2025-04-09
642 publish for ready edition

ellohez:
**WHAT**
Implement Publish button and process for Ready edition.

-------------
[Trello card](https://trello.com/c/ljJbaP3q)

|Scenario|View|
|-|-|
|The Publish button shows when the edition is in a ready state, and user has the right permissions.|Visible publish button|
|Clicking the button sends the user to the send to Publish confirmation page. User can enter optional comments, and click the Send to Publish button to publish the edition. Cancel button takes the user back to the Edit tab.|Visible publish button and textbox|
|Success message present and edition is in Published state. Edition is now edit-only.|Edition state updated and success message shows|

### DIFF
--- a/test/functional/editions_controller_test.rb
+++ b/test/functional/editions_controller_test.rb
@@ -520,84 +520,166 @@ class EditionsControllerTest < ActionController::TestCase
   end
 
   context "#send_to_publish" do
-    setup do
-      @requester = FactoryBot.create(:user, :govuk_editor, name: "Stub Requester")
-      @edition = FactoryBot.create(:edition, state: "scheduled_for_publishing", publish_at: Time.zone.now + 1.hour)
-      login_as(@requester)
-    end
-
-    context "user has govuk_editor permission" do
-      setup do
-        ScheduledPublisher.stubs(:cancel_scheduled_publishing)
-        PublishWorker.stubs(:perform_async)
-      end
-
-      should "update the edition status to 'published' and save the comment" do
-        post :send_to_publish, params: {
-          id: @edition.id,
-          comment: "Publishing with immediate effect",
-        }
-
-        assert_equal "Published", flash[:success]
-        @edition.reload
-        assert_equal "published", @edition.state
-        assert_equal "Publishing with immediate effect", @edition.latest_status_action.comment
-        assert_equal @requester.id, @edition.latest_status_action.requester_id
-      end
-
-      should "not update the edition state and render 'send_to_publish' template with an error when an error occurs" do
-        EditionProgressor.any_instance.expects(:progress).returns(false)
-
-        post :send_to_publish, params: {
-          id: @edition.id,
-        }
-
-        assert_template "secondary_nav_tabs/send_to_publish_page"
-        assert_equal "Due to a service problem, the request could not be made", flash[:danger]
-        @edition.reload
-        assert_equal "scheduled_for_publishing", @edition.state
-      end
-
-      should "notify the publishing API to publish the edition" do
-        PublishWorker.expects(:perform_async).with(@edition.id.to_s)
-
-        post :send_to_publish, params: {
-          id: @edition.id,
-        }
-      end
-    end
-
-    context "user does not have govuk_editor permission" do
-      setup do
-        user = FactoryBot.create(:user)
-        login_as(user)
-      end
-
-      should "render an error message" do
-        post :send_to_publish, params: {
-          id: @edition.id,
-        }
-
-        assert_equal "You do not have correct editor permissions for this action.", flash[:danger]
-      end
-    end
-
-    context "edition is not in a valid state to be published" do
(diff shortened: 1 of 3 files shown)


test/functional/editions_controller_test.rb
```diff
@@ -520,84 +520,167 @@ class EditionsControllerTest < ActionController::TestCase
...
+        should "not update the edition state and render 'send_to_publish' template with an error when an error occurs" do
+          EditionProgressor.any_instance.expects(:progress).returns(false)
+
+          post :send_to_publish, params: {
+            id: @edition.id,
+          }
+
+          assert_template "secondary_nav_tabs/send_to_publish_page"
+          assert_equal "Due to a service problem, the request could not be made", flash[:danger]
+          @edition.reload
```
davidtrussler, line 558: Is `@edition.reload` needed here? I think the state can be tested without that.

test/functional/editions_controller_test.rb
```diff
@@ -520,84 +520,167 @@ class EditionsControllerTest < ActionController::TestCase
...
+          assert_equal "scheduled_for_publishing", @edition.state
+        end
+
+        should "notify the publishing API to publish the edition" do
```
davidtrussler, line 562: This test has no assertions. Is that right?

test/functional/editions_controller_test.rb
```diff
@@ -520,84 +520,167 @@ class EditionsControllerTest < ActionController::TestCase
...
-        }
+        should "not update the edition state and render 'send_to_publish' template with an error" do
+          post :send_to_publish, params: {
+            id: @edition.id,
+          }
+
+          assert_equal "Edition is not in a state where it can be published", flash[:danger]
+
+          assert_template "secondary_nav_tabs/send_to_publish_page"
+          @edition.reload
```
davidtrussler, line 600: Again, I'm not sure if `@edition.reload` does anything useful here.

test/functional/editions_controller_test.rb
```diff
@@ -520,84 +520,167 @@ class EditionsControllerTest < ActionController::TestCase
...
+            id: @edition.id,
+          }
+
+          assert_template "secondary_nav_tabs/send_to_publish_page"
+          assert_equal "Due to a service problem, the request could not be made", flash[:danger]
+          @edition.reload
+          assert_equal "ready", @edition.state
+        end
+
+        should "notify the publishing API to publish the edition" do
```
davidtrussler, line 643: As above, this has no assertions. Is that correct?

test/functional/editions_controller_test.rb
```diff
@@ -520,84 +520,167 @@ class EditionsControllerTest < ActionController::TestCase
...
+        end
       end
 
-      should "not update the edition state and render 'send_to_publish' template with an error" do
-        post :send_to_publish, params: {
-          id: @edition.id,
-        }
+      context "edition is not in a valid state to be published" do
+        setup do
+          # @requester = FactoryBot.create(:user, :govuk_editor, name: "Stub Requester")
```
davidtrussler, line 669: Commented line should be deleted.

test/functional/editions_controller_test.rb
PeterHattyar: Yes, that is correct.
**PublishWorker** is a stub here, and it is expecting that the code in the test will call the PublishWorker.perform_asynch method with the edition id as a string. With any other value it would fail. So it works as intended.

test/functional/editions_controller_test.rb
PeterHattyar: Indeed, that is correct, no need, as it is tested via a stub, so incorrect string values not being the edition id would cause it to fail.

test/functional/editions_controller_test.rb
PeterHattyar: You're correct, it doesn't seem to require it, we have just followed suit in the test suite, as other tests present for editions in **scheduled_for_publishing** state utilise it in the same way, even though they similarly don't fail if removed. So we did the same in our copy of those tests. (Like on [this line](https://github.com/alphagov/publisher/blob/main/test/functional/editions_controller_test.rb#L672))

test/functional/editions_controller_test.rb
PeterHattyar: Like on the other comment, we might be better just leaving it to keep integrity with the other tests.

test/functional/editions_controller_test.rb
PeterHattyar: Good spot, I shall remove this comment.

test/functional/editions_controller_test.rb
davidtrussler: Yes, I think leaving it for consistency is fine. 👍

davidtrussler (approved):
Nice work, all looking good to me! 👍

PR #2657, opened 2025-04-24
671 remove the 6 default status checkbox

ellohez:
## What
Change the default from 6 of the 9 status checkboxes selected to all 9 of them unselected. This affects the initial view and the 'Reset all fields' link. Further manual testing is required to ensure that there are no regressions for the Browser extension. 

## [Trello card](https://trello.com/c/w4LP1YRC)

| |Original|New|
|-|-|-|
|Default (initial view) with no user changes|6 checkboxes checked - current user selected|No checkboxes checked -current user selected|
|Reset all fields if assignee changed|6_checkboxes_selected|no checkboxes selected|
|View when navigating from browser extension|8 checkboxes selected|8 checkboxes selected|

### DIFF
--- a/test/functional/root_controller_test.rb
+++ b/test/functional/root_controller_test.rb
@@ -13,7 +13,7 @@ class RootControllerTest < ActionController::TestCase
       assert_template "root/index"
     end
 
-    should "default the state filter checkboxes to unchecked when no filters are specified" do
+    should "default all state filter checkboxes to unchecked" do
       get :index
 
       assert_select "input.govuk-checkboxes__input", 9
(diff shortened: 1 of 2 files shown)


test/functional/root_controller_test.rb
```diff
@@ -13,58 +13,62 @@ class RootControllerTest < ActionController::TestCase
       assert_template "root/index"
     end
 
-    should "default the state filter checkboxes when no filters are specified" do
+    should "default the state filter checkboxes to unchecked when no filters are specified" do
```
syed-ali-tw, line 16: not an english expert.. just read this a few times and the test name still feels that we could word it a bit better..
"default all state filter checkboxes to unchecked when no filters are specified"
"default state filter checkboxes to all unchecked when no filters are specified"
"uncheck all state filter checkboxes as a default"

also probably 'when no filters are specified' is redundant as we are talking about default status?

syed-ali-tw (commented):
LGTM, only a minor opinionated comment

test/functional/root_controller_test.rb
ellohez: Well spotted, thank you. Have updated this accordingly and hopefully made a couple of other test names clearer.

syed-ali-tw (approved):
LGTM

PR #2659, opened 2025-04-25
649 schedule for ready edition

davidtrussler:
[Trello](https://trello.com/c/8V0CiJv6/649-schedule-for-ready-edition-answer-and-help-page)

These changes: 
- Adds a "Schedule" button to the Edit page when: 
  - the user has the correct permissions
  - the edition state is 'ready'
- Adds a new "Schedule publication page" page view
- Adds the logic to 
  - progress the edition to the "Schedule for publishing" state
  - add success and error messages as applicable

See screenshots for UI and messaging. 

NB: There are two validations here (checking that date/time fields are not blank and are not set to a date in the past) that have been implemented by using flash messages rather than more standard validation and the error message component. There are complications around validation due to how the Edition Progressor is set up to deal with the current way this works that does not need to make the same validations. This has made the controller a bit clunky within the `schedule` action. This will be dealt with in a future piece of work to use more standard validation here. 

NNB: this is using a partial in the downtimes page to render the time fields which is not ideal. The decision was made to do this for now whilst scheduling some future work to make a date/time component available in the Publishing Components Gem and then use that in both the "Schedule downtimes" and "Schedule publications" pages.

|Page/scenario|View|
|-|-|
|Edit page: edition state is "In review", user has correct permissions. "Schedule" button is rendered|Screenshot 2025-04-28 at 11 14 15|
|Schedule publication page on page load|Screenshot 2025-04-28 at 15 28 11|
|Edit page: user without correct permissions attempts to save|Screenshot 2025-04-28 at 11 18 53|
|Edit page: user attempts to schedule page that is not in the correct state|Screenshot 2025-04-28 at 11 22 59|
|Schedule publication page: user attempts to save with blank fields|Screenshot 2025-04-28 at 11 27 11|
|Schedule publication page: user attempts to save with a date in the past|Screenshot 2025-04-28 at 11 28 48|
|Edit page: user saves successfully|Screenshot 2025-04-28 at 11 30 05|
|Schedule publication page: a server problem occurs|Screenshot 2025-04-28 at 11 34 29|

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -11,7 +11,7 @@ class EditionsController < InheritedResources::Base
   before_action only: %i[unpublish confirm_unpublish process_unpublish] do
     require_govuk_editor(redirect_path: edition_path(resource))
   end
-  before_action only: %i[progress admin update confirm_destroy edit_assignee update_assignee edit_reviewer update_reviewer request_amendments request_amendments_page no_changes_needed no_changes_needed_page send_to_2i send_to_2i_page send_to_publish send_to_publish_page cancel_scheduled_publishing cancel_scheduled_publishing_page] do
+  before_action only: %i[progress admin update confirm_destroy edit_assignee update_assignee edit_reviewer update_reviewer request_amendments request_amendments_page no_changes_needed no_changes_needed_page send_to_2i send_to_2i_page send_to_publish send_to_publish_page cancel_scheduled_publishing cancel_scheduled_publishing_page schedule schedule_page] do
     require_editor_permissions
   end
   before_action only: %i[confirm_destroy destroy] do
@@ -25,6 +25,9 @@ class EditionsController < InheritedResources::Base
   end
   before_action only: %i[edit_reviewer update_reviewer] do
     require_reviewer_editable
+  end
+  before_action only: %i[schedule_page] do
+    require_schedulable
   end
 
   helper_method :locale_to_language
@@ -58,6 +61,10 @@ class EditionsController < InheritedResources::Base
 
   def skip_review_page
     render "secondary_nav_tabs/skip_review_page"
+  end
+
+  def schedule_page
+    render "secondary_nav_tabs/schedule_page"
   end
 
   def send_to_publish_page
@@ -155,6 +162,29 @@ class EditionsController < InheritedResources::Base
     else
       flash.now[:danger] = "Due to a service problem, the request could not be made"
       render "secondary_nav_tabs/skip_review_page"
+    end
+  end
+
+  def schedule
+    if !@resource.can_schedule_for_publishing?
+      flash.now[:danger] = "Edition is not in a state where it can be scheduled for publishing"
+      render "secondary_nav_tabs/schedule_page"
+    elsif params[:publish_at_1i].empty? || params[:publish_at_2i].empty? || params[:publish_at_3i].empty? || params[:publish_at_4i].empty? || params[:publish_at_5i].empty?
+      flash.now[:danger] = "Select a future time and/or date to schedule publication."
+      render "secondary_nav_tabs/schedule_page"
+    else
+      publish_at = Time.zone.local(params[:publish_at_1i].to_i, params[:publish_at_2i].to_i, params[:publish_at_3i].to_i, params[:publish_at_4i].to_i, params[:publish_at_5i].to_i)
+
+      if publish_at.present? && publish_at < Time.zone.now
+        flash.now[:danger] = "Select a future time and/or date to schedule publication."
+        render "secondary_nav_tabs/schedule_page"
+      elsif schedule_for_edition(@resource, params[:comment], publish_at)
+        flash[:success] = "Scheduled to publish at #{publish_at.to_fs(:govuk_date)}"
+        redirect_to edition_path(resource)
+      else
+        flash.now[:danger] = "Due to a service problem, the request could not be made"
+        render "secondary_nav_tabs/schedule_page"
+      end
     end
   end
 
@@ -364,6 +394,11 @@ private
     @command.progress({ request_type: "skip_review", comment: comment })
   end
 
+  def schedule_for_edition(resource, comment, publish_at)
+    @command = EditionProgressor.new(resource, current_user)
+    @command.progress({ request_type: "schedule_for_publishing", comment: comment, publish_at: publish_at })
+  end
+
   def send_to_publish_for_edition(resource, comment)
     @command = EditionProgressor.new(resource, current_user)
     publish_succeeded = @command.progress({ request_type: "publish", comment: comment })
@@ -435,6 +470,13 @@ private
     return if @resource.can_destroy?
 
     flash[:danger] = "Cannot delete a #{description(@resource).downcase} that has ever been published."
+    redirect_to edition_path(@resource)
+  end
+
+  def require_schedulable
+    return if @resource.can_schedule_for_publishing?
+
+    flash[:danger] = "Cannot schedule an edition that is not ready."
     redirect_to edition_path(@resource)
   end
 
--- a/test/functional/editions_controller_test.rb
+++ b/test/functional/editions_controller_test.rb
@@ -487,6 +487,241 @@ class EditionsControllerTest < ActionController::TestCase
     end
   end
 
+  context "#schedule_page" do
+    setup do
+      @ready_edition = FactoryBot.create(:answer_edition, :ready)
+    end
+
+    context "user has govuk_editor permission" do
+      should "be able to navigate successfully to schedule page path" do
+        get :schedule_page, params: { id: @ready_edition.id }
+
+        assert_response :success
+      end
+    end
+
+    context "user does not have govuk_editor permission" do
+      setup do
+        user = FactoryBot.create(:user)
+        login_as(user)
+      end
+
+      should "show permission error and redirect to edition path" do
+        get :schedule_page, params: { id: @ready_edition.id }
+
+        assert_redirected_to edition_path(@ready_edition)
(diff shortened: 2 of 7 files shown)


app/models/workflow.rb
```diff
@@ -145,6 +145,10 @@ def can_destroy?
     !scheduled_for_publishing? && !published? && !archived?
   end
 
+  def can_schedule?
```
ellohez, line 148: I think this already exists as `edition.can_schedule_for_publishing?`. Just found it in the (legacy?) `EditionActivityButtonsHelper`

test/functional/editions_controller_test.rb
```diff
@@ -482,6 +482,196 @@ class EditionsControllerTest < ActionController::TestCase
...
+            publish_at_2i: 5,
+            publish_at_1i: 2024,
+            publish_at_4i: 0,
+            publish_at_5i: 1,
+          }
+
+          assert_equal "Select a future time and/or date to schedule publication.", flash[:danger]
+        end
+
+        should "update the edition state to 'ready' and save the comment" do
```
ellohez, line 612: Should the test name be 'update the edition state to 'scheduled_for_publishing' and save the comment'?

test/functional/editions_controller_test.rb
```diff
@@ -482,6 +482,196 @@ class EditionsControllerTest < ActionController::TestCase
...
+          post :schedule, params: {
+            id: @edition.id,
+            comment: "Scheduling for publish",
+            publish_at_3i: 24,
+            publish_at_2i: 5,
+            publish_at_1i: 2025,
+            publish_at_4i: 0,
+            publish_at_5i: 1,
+          }
+
```
ellohez, line 668: Not sure if this is an issue but do we need to test where the user is at this point? I.e. which page the error is displayed on?

ellohez (commented):
Just a couple of minor questions!

app/models/workflow.rb
davidtrussler: Looks like you're right: have removed the `can_schedule?` method from here and replaced in the controller with `can_schedule_for_publishing?`.

test/functional/editions_controller_test.rb
davidtrussler: Yes it certainly should! Have updated that.

test/functional/editions_controller_test.rb
davidtrussler: Yeah, it's a slightly odd one this to my mind: it's kind of a bit abstract in that if the user doesn't have permissions they shouldn't be there so I guess it would display at whatever the current location is. It would only really occur if the user's permissions changed somehow between navigating to the schedule page and trying to schedule the edition. I've based this test on the one on [line 683 below](https://github.com/alphagov/publisher/pull/2659/files#diff-60e62045ebb6b58d548098b54dfd1b2e96d65438e325af47892befae1c528274R683).

app/controllers/editions_controller.rb
```diff
@@ -158,6 +165,29 @@ def skip_review
...
+    elsif params[:publish_at_1i].empty? || params[:publish_at_2i].empty? || params[:publish_at_3i].empty? || params[:publish_at_4i].empty? || params[:publish_at_5i].empty?
+      flash.now[:danger] = "Select a future time and/or date to schedule publication."
+      render "secondary_nav_tabs/schedule_page"
+    else
+      publish_at = Time.zone.local(params[:publish_at_1i].to_i, params[:publish_at_2i].to_i, params[:publish_at_3i].to_i, params[:publish_at_4i].to_i, params[:publish_at_5i].to_i)
+
+      if publish_at.present? && publish_at < Time.zone.now
+        flash.now[:danger] = "Select a future time and/or date to schedule publication."
+        render "secondary_nav_tabs/schedule_page"
+      elsif schedule_for_edition(@resource, params[:comment], publish_at)
```
TonyGDS, line 181: I think this might read a little better if it was a variable, because the function name doesn't sound like a boolean. Similar to what is done in `send_to_publish_for_edition`:

`schedule_succeeded = schedule_for_edition(@resource, params[:comment], publish_at)`

or return the command object from the schedule_for_edition function and query it's status

test/functional/editions_controller_test.rb
```diff
@@ -487,6 +487,198 @@ class EditionsControllerTest < ActionController::TestCase
...
+      context "govuk editor and non-Welsh edition" do
+        should "be able to navigate to the edit reviewer path" do
+          @ready_edition = FactoryBot.create(:answer_edition, :ready)
+          get :schedule_page, params: { id: @ready_edition.id }
+
+          assert_response :success
+        end
+      end
+
+      %i[draft in_review amends_needed fact_check fact_check_received scheduled_for_publishing published archived].each do |edition_state|
```
TonyGDS, line 543: This might be better in another context `context "Edition is not in 'Ready' state"`

test/functional/editions_controller_test.rb
```diff
@@ -487,6 +487,198 @@ class EditionsControllerTest < ActionController::TestCase
...
+      end
+
+      context "user has govuk_editor permission" do
+        setup do
+          @requester = FactoryBot.create(:user, :govuk_editor, name: "Stub Requester")
+          login_as(@requester)
+        end
+
+        should "show an error if any date/time fields are left blank" do
+          post :schedule, params: {
```
TonyGDS, line 592: I think we should be testing for all of these fields given the code checks each on individually

test/functional/editions_controller_test.rb
```diff
@@ -487,6 +487,198 @@ class EditionsControllerTest < ActionController::TestCase
...
+          }
+
+          assert_template "secondary_nav_tabs/schedule_page"
+          assert_equal "Due to a service problem, the request could not be made", flash[:danger]
+          @edition.reload
+          assert_equal "ready", @edition.state
+        end
+      end
+
+      context "user does not have govuk_editor permission" do
```
TonyGDS, line 659: should this be `context "user does not have govuk_editor or welsh_editor permissions"`?

test/integration/edition_edit_test.rb
```diff
@@ -925,6 +925,127 @@ class EditionEditTest < IntegrationTest
...
+    context "Schedule page" do
+      setup do
+        create_ready_edition
+        visit schedule_page_edition_path(@ready_edition.id)
+      end
+
+      should "render the 'Schedule' page" do
+        within :css, ".gem-c-heading" do
+          assert page.has_css?("h1", text: "Schedule publication")
+          assert page.has_css?(".gem-c-heading__context", text: @ready_edition.title)
```
TonyGDS, line 996: are we using `assert page.has_css?` to test the css or just if the text was rendered? If it's just the text would it be better to use `assert page.has_text?` so that the tests don't break if we update the design system gem?

test/functional/editions_controller_test.rb
```diff
@@ -487,6 +487,198 @@ class EditionsControllerTest < ActionController::TestCase
...
+
+        should "show permission error and redirect to edition path" do
+          get :schedule_page, params: { id: @edition.id }
+
+          assert_redirected_to edition_path(@edition)
+          assert_equal "You do not have correct editor permissions for this action.", flash[:danger]
+        end
+      end
+
+      context "non-govuk editor and non-Welsh edition" do
```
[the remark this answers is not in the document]
davidtrussler: Yes, I think we have not been very consistent on this in other tests, but looking over them I think this is probably best expressed within categories of user following this pattern, which I have used to update this section of the test. I think it covers all variations of users and editions that we need to test: 
- user has govuk_editor permissions: they can edit all editions (no need to test for Welsh or non-Welsh)
- user does not have govuk_editor permissions: they can't edit any editions (no need to test for Welsh or non-Welsh)
- user has welsh_editor permissions only: they can edit Welsh editions but not non-Welsh editions

test/functional/editions_controller_test.rb
davidtrussler: Yes, that makes sense. I've just copied and pasted the fields and set one to nil because I couldn't see a quick way of doing this more dynamically - happy to see any suggestions on that! It looks a bit cumbersome but I'm hoping that when we get back to doing the validations in a better way - which is scheduled as a future piece of work - this will naturally be tidied up somewhat.

test/integration/edition_edit_test.rb
davidtrussler: I think I'd leave this as is though I see your point. Probably just being overly frontend orientated but I kind of prefer testing that text is shown in the context of the component containing it. We tend to follow this pattern (to a greater or lesser extent) in other tests.

app/controllers/editions_controller.rb
davidtrussler: Not sure about this: I think it is pretty much in line with the way methods are named and used in other tests, though I see what you are saying. I think the only reason the `send_to_publish_for_edition` method uses a variable is so that it can be re-used.

test/integration/edition_edit_test.rb
TonyGDS: Sure, the reason I brought this up is that I tend to follow the [principles of Testing Library](https://testing-library.com/docs/guiding-principles/) where we interact with the page more like a user would. It promotes more accessibility focus in terms of the user, but also can make the tests less brittle e.g. change to css can break a test. Testing Library does have more [selectors](https://testing-library.com/docs/queries/about/#priority) than Capybara though.

test/functional/editions_controller_test.rb
TonyGDS: Yeah, I understand it doesn't feel optimal, but at least it guarantees its tested. A dynamic solution for this is probably a bit convoluted and not worth the effort. As you say hopefully we can improve it when we update the validations.

TonyGDS (approved):
LGTM

PR #2663, opened 2025-04-29
Add 'Request amendments link' for editions in fact check state

ellohez:
## What
A link to 'Request amendments' for editions that are out for fact check. The existing Request amendments page is then displayed when the link is clicked, a comment can be saved by the user and the edition is changed to 'Amends needed' state.

## [Trello card](https://trello.com/c/CRzIyejJ)

|Current|Updated|
|-|-|
|Existing edit page for an edition in fact check state with no link|New edit page with 'Request amendments' link for fact check state|

- **NOTE:** The success message has been changed to a more generic 'Amendments requested' to cover all conditions/state changes 

|Scenario|View|
|-|-|
|User clicks 'Request amendments' link and is taken to the new 'Request amendments' page|Screenshot 2025-02-17 at 17 02 58|
|User clicks 'Request amendments' button and an error occurs|Screenshot 2025-02-17 at 17 06 19|
|User clicks 'Request amendments' button and the request is successful|Screenshot 2025-04-30 at 11 29 15|

### DIFF
--- a/app/views/editions/show.html.erb
+++ b/app/views/editions/show.html.erb
@@ -45,10 +45,10 @@
         <% end %>
       <% end %>
     </div>
-  <% elsif @edition.ready? %>
+  <% elsif @edition.ready? || @edition.fact_check? %>
     <div class="govuk-grid-column-two-thirds">
-      <%= render "govuk_publishing_components/components/inset_text", {} do %>
-        <% if current_user.has_editor_permissions?(@edition) %>
+      <% if current_user.has_editor_permissions?(@edition) %>
+        <%= render "govuk_publishing_components/components/inset_text", {} do %>
           <p class="govuk-body"><%= link_to("Request amendments", request_amendments_page_edition_path) %></p>
         <% end %>
       <% end %>
--- a/test/functional/editions_controller_test.rb
+++ b/test/functional/editions_controller_test.rb
@@ -99,7 +99,7 @@ class EditionsControllerTest < ActionController::TestCase
               comment: "This is a comment",
             }
 
-            assert_equal "2i amendments requested", flash[:success]
+            assert_equal "Amendments requested", flash[:success]
             @edition.reload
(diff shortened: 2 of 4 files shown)


app/views/editions/show.html.erb
```diff
@@ -45,10 +45,10 @@
         <% end %>
       <% end %>
     </div>
-  <% elsif @edition.ready? %>
+  <% elsif @edition.ready? || @edition.fact_check? %>
     <div class="govuk-grid-column-two-thirds">
-      <%= render "govuk_publishing_components/components/inset_text", {} do %>
-        <% if current_user.has_editor_permissions?(@edition) %>
+      <% if current_user.has_editor_permissions?(@edition) %>
+        <%= render "govuk_publishing_components/components/inset_text", {} do %>
```
ellohez, line 51: I moved this render inside the `if` statement to prevent a situation where an empty inset text is rendered as this would display just the grey box on the left hand side.

test/functional/editions_controller_test.rb
```diff
@@ -64,70 +64,75 @@ class EditionsControllerTest < ActionController::TestCase
...
       end
     end
 
-    context "user does not have govuk_editor permission" do
-      setup do
-        user = FactoryBot.create(:user)
-        login_as(user)
-      end
+    # NOTE: Eventually this list will also contain the 'fact_check_received' state
+    %i[in_review ready fact_check].each do |edition_state|
```
ellohez, line 85: I wasn't sure if this would make the tests too DRY but I hope that it will save duplication in the future.

davidtrussler: I think this - as a means to avoid duplicating the test for each state -  is a valid way of doing this here. For me the main thing is that it remains easily human readable.

davidtrussler (commented):
I think this looks fine and ready to approve. Do we want to wait for some feedback on the spacing issue before proceeding?

davidtrussler (approved):
LGTM! 👍

PR #2678, opened 2025-05-12
Css important note inset text improvement

ellohez:
## What
A large gap between the Important note element and the Inset text was noticed as part of the desk check for [Fact check - [Request amendments] Answer and Help Page](https://github.com/alphagov/publisher/pull/2663). 
This has now been addressed so that with or without the inset text, the gap below the important note should be 30px. 

## [Trello card](https://trello.com/c/CRzIyejJ)

Out of scope for this is the gap which still exists between the inset text and the summary list which will be addressed as part of a design sweep. 

|Scenario|View|
|-|-|
|New gap between the important note and the inset text|New gap between the important note and inset text|
|Gap between the important note and tab headers with no inset text|Gap between the important note and tab headers with no inset text|
|Important note with larger inset text|Important note with larger inset text|
|Inset text with no important note|Inset text with no important note|

### DIFF
--- a/app/assets/stylesheets/editions.scss
+++ b/app/assets/stylesheets/editions.scss
@@ -14,3 +14,7 @@
     margin-top: govuk-spacing(3)
   }
 }
+
+.gem-c-notice + .gem-c-inset-text {
+  padding-top: 0;
+}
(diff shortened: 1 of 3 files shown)


davidtrussler:
This looks good to me but I probably shouldn't formally approve it because I was involved in the solution.

baisa (approved):

PR #2685, opened 2025-05-21
Migrate Fact Check email retrieval to Gmail API

Tetrino:
Google removed traditional auth methods for accessing inboxes, migrating to OAuth2. The result of this is that we have to migrate our method of accessing the inboxes from the original username/password system to a service account calling the GMail API.

Notable changes:

- We no longer need MailFetcherConfig as the Mail gem is no longer used to access the email inbox. This config was only ever used in our automated script and is replaced by the Gmail auth setup in FactCheckEmailHandler

- We no longer use email and password, instead using a JSON API secrets key that is stored in AWS secrets and called via ENV.

Functionally the resulting logic is the same as previously, by converting calls to from gmail message objects to Mail objects for actual handling logic.

https://trello.com/c/r8cUPjj3/


test/integration/fact_check_email_test.rb
```diff
@@ -1,58 +1,68 @@
...
+        BCC: attrs.fetch(:bcc, nil),
+        Subject: attrs.fetch(:subject, "This is a fact check response [#{edition.present? ? edition.id : ''}]"),
+      },
+      attrs.fetch(:body, "I like it. Good work!"),
+      base_mail.to_s,
+    )
+  end
+
+  def handler(messages)
+    handler = FactCheckEmailHandler.new(fact_check_config, stub.stubs(:set))
```
baisa, line 30: stub.stubs is a bit confusing, maybe we could define it separately?

app/lib/fact_check_email_handler.rb
```diff
@@ -43,4 +48,49 @@ def process
...
+      message_content = retrieve_message_content(message.id)
+      processed_mails[message.id] = Mail.read_from_string(message_content.raw)
+    end
+    processed_mails
+  end
+
+  def authenticate(gmail)
+    # Gmail insists on taking an IO for account credentials in Ruby
+    parsed_key = StringIO.new(ENV.fetch("FACT_CHECK_API_KEY"))
+    parsed_key.rewind # A little safety to avoid passing a nil string
```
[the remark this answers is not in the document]
Tetrino: Fun fact I had a race condition wherein sometimes it would need rewinding. I refactored that out and forgot this. 😅

test/gmail_test_helper.rb
```diff
@@ -0,0 +1,79 @@
...
+  Google::Apis::GmailV1::MessagePart.new(
+    headers: header_array,
+    parts: [
+      Google::Apis::GmailV1::MessagePart.new(
+        body: create_body(body),
+        headers: [
+          create_header("Content-Type", "text/plain; charset=\"UTF-8\""),
+        ],
+        mime_type: "text/html",
+        part_id: "0",
```
baisa, line 75: is it ok to have it hardcoded?

app/lib/fact_check_email_handler.rb
```diff
@@ -43,4 +48,49 @@ def process
...
+
+    scopes = ["https://www.googleapis.com/auth/gmail.labels", "https://www.googleapis.com/auth/gmail.modify"]
+    authorizer = Google::Auth::ServiceAccountCredentials.make_creds(json_key_io: parsed_key, scope: scopes)
+
+    gmail.authorization = authorizer.dup
```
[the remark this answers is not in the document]
Tetrino: No - fixed in next push.

test/integration/fact_check_email_test.rb
Tetrino: Despite this being marked as new code, this is 1:1 what currently exists in our test framework. I just moved its location https://github.com/alphagov/publisher/blob/5325b16b97bf180609711494c93f4a09f36c2310/test/integration/fact_check_email_test.rb#L28 

I was trying not to rock the boat too much - this stub was put in place by Richard when putting in the prometheus metrics and is purely there to "eat" that call as we're not considering prometheus here.

test/gmail_test_helper.rb
```diff
@@ -0,0 +1,79 @@
+require "test_helper"
+require "google/apis/gmail_v1"
+require "securerandom"
+require "time"
+require "webmock"
+
+# Building a gmail message to match that from a server is a complicated affair, this helper places the logic somewhere reasonable.
```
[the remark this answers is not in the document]
Tetrino: Now I am more awake I want to clarify a bit my thought process here.

While I stubbed functions that call outwards, the structure of a gmail message is fundamental to how the new logic works. Creating a GMail response object is more verbose than a straight Mail object, but my logic is fundamentally not actually that much different than what the test was already doing - [creating a fake Mail object](https://github.com/alphagov/publisher/blob/main/test/integration/fact_check_email_test.rb#L5) in its entirety then working on it. The only real difference is that a Mail object is a fundamentally simple thing, whereas Google give us quite a nice, if verbose, collection of things to work with.

With a little effort we could double a lot of it, but it flies in the face of what we're actually trying to test here imho.

test/integration/fact_check_email_test.rb
richardTowers: Oh yeah, sorry for the bad variable name 😓  - I think let's leave it if it was already there. I might have a look and see if it can be made better in a separate PR.

app/lib/fact_check_email_handler.rb
```diff
@@ -43,4 +48,49 @@ def process
...
+    processed_mails
+  end
+
+  def authenticate(gmail)
+    # Gmail insists on taking an IO for account credentials in Ruby
+    parsed_key = StringIO.new(ENV.fetch("FACT_CHECK_API_KEY"))
+    parsed_key.rewind # A little safety to avoid passing a nil string
+
+    scopes = ["https://www.googleapis.com/auth/gmail.labels", "https://www.googleapis.com/auth/gmail.modify"]
+    authorizer = Google::Auth::ServiceAccountCredentials.make_creds(json_key_io: parsed_key, scope: scopes)
```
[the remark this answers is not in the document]
Tetrino: Re: API key formatted string, in other language implementations (Java and Python for instance) this is indeed the case. For some reason they did it differently in Ruby.

app/lib/fact_check_email_handler.rb
```diff
@@ -43,4 +48,45 @@ def process
...
+    retrieve_message_list.each do |message|
+      message_content = retrieve_message_content(message.id)
+      processed_mails[message.id] = Mail.read_from_string(message_content.raw)
+    end
+    processed_mails
+  end
+
+  def authenticate_gmail
+    # Gmail insists on taking an IO for account credentials in Ruby
+    parsed_key = StringIO.new(ENV.fetch("FACT_CHECK_API_KEY"))
```
syed-ali-tw, line 66: can we fetch this from config?

app/lib/fact_check_email_handler.rb
```diff
@@ -43,4 +48,45 @@ def process
...
+
+    scopes = ["https://www.googleapis.com/auth/gmail.labels", "https://www.googleapis.com/auth/gmail.modify"]
+    @gmail.authorization = Google::Auth::ServiceAccountCredentials.make_creds(json_key_io: parsed_key, scope: scopes)
+
+    @gmail.authorization.sub = ENV.fetch("FACT_CHECK_USERNAME")
```
syed-ali-tw, line 71: can we fetch this from config?

richardTowers (approved):
Looks good to me!

I added another couple of nits, but they're not blocking so you can take them or leave them :)

PR #2700, opened 2025-06-02
Editions reviewer fix

syed-ali-tw:
[Trello](https://trello.com/c/PpxCshge/718-fix-reviewer-issue-to-work-with-both-new-design-system-and-bootstrap)

- Fix claiming 2i by saving name of the user instead of id, as it was always saved by name and not id.
- fetch reviewer from db to show correctly for history and notes and related external links tab.
- Undo the changes to hide claim 2i changes in https://github.com/alphagov/publisher/pull/2698


TonyGDS (approved):
LGTM :shipit:

PR #2708, opened 2025-06-09
710 Fix metadata permissions

ellohez:
## What
Follows the existing permissions for Artefact. This update shows the tab as read-only with no 'update' button if the user does not have the correct permission to update the metadata for the artefact. 

## [Trello card](https://trello.com/c/XPnx8J2j)

|Scenario|View|
|-|-|
|Previous view for all users with any permission|Screenshot 2025-06-09 at 17 36 31|
|Current view if user does not have permission|Screenshot 2025-06-09 at 17 21 40|

### DIFF
--- a/test/functional/artefacts_controller_test.rb
+++ b/test/functional/artefacts_controller_test.rb
@@ -19,5 +19,59 @@ class ArtefactsControllerTest < ActionController::TestCase
       assert_redirected_to root_path
       assert_includes flash[:danger], "do not have permission"
     end
+
+    should "not allow creation if no permissions" do
+      user = FactoryBot.create(:user, name: "Non-editor")
+      login_as(user)
+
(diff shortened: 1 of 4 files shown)


test/functional/artefacts_controller_test.rb
```diff
@@ -19,5 +19,59 @@ class ArtefactsControllerTest < ActionController::TestCase
...
+
+    should "not allow creation if no permissions" do
+      user = FactoryBot.create(:user, name: "Non-editor")
+      login_as(user)
+
+      get :new
+
+      assert_response :redirect
+      assert_redirected_to root_path
+      assert_includes flash[:danger], "do not have permission"
```
TonyGDS, line 31: Maybe we could check for the full message here

PR #2709, opened 2025-06-10
Extend `govuk-link` styles to 'action' messages within 'history & not…

davidtrussler:
[Trello](https://trello.com/c/JLczRnEb/714-issue-publisher-text-link-colours-on-history-and-notes-tab-are-not-matching-govuk-guidelines)

The problem identified here is that links on the "History and notes" tab are not styled according to GOVUK guidelines. This is because that content it generated from user actions and is not able to be directly styled. 

The changes in this PR applies the styles which have already been used to style the links within fact check messages to links within all the "action" content on the "History and notes" page by extending the `govuk-link` and `govuk-link--no-visited-state` styles to all `<a>` elements within the required areas. 

The screenshots below show the updated styles applied to all the links in a variety of different sections in default, hover and focus states. Note that this is not really a realistic scenario since all links can't have hover and focus states at one time. 

|Scenario|Current|Updated|
|-|-|-|
|Links: default state|publisher dev gov uk_editions_68388cca99686e005f7da1b3_history (3)|publisher dev gov uk_editions_68388cca99686e005f7da1b3_history|
|Links: focus state|publisher dev gov uk_editions_68388cca99686e005f7da1b3_history (4)|publisher dev gov uk_editions_68388cca99686e005f7da1b3_history (1)|
|Links: hover state|publisher dev gov uk_editions_68388cca99686e005f7da1b3_history (5)|publisher dev gov uk_editions_68388cca99686e005f7da1b3_history (2)|


TonyGDS (approved):
LGTM :shipit: 

I was wondering why we suppress stylelint checks but then I saw this directive on the sass website.
> As a rule of thumb, if you’re writing a stylesheet just for your own app, it’s often better to just extend a class selector if one is available.

PR #2720, opened 2025-06-13
Fix permissions for the external links tab

Tetrino:
User can no longer see the Add, Delete, or Save options. Also hidden the advisory
text indicating that a save can occur to avoid confusion for the user.

### DIFF
--- a/app/views/editions/secondary_nav_tabs/_related_external_links.html.erb
+++ b/app/views/editions/secondary_nav_tabs/_related_external_links.html.erb
@@ -18,21 +18,23 @@
         end
       %>
 
-      <%= render "govuk_publishing_components/components/add_another", {
-        fieldset_legend: "Link",
-        add_button_text: "Add related external link",
-        empty_fields: true,
-        items: items,
-        empty: empty,
-      } %>
+      <% if current_user.has_editor_permissions?(@resource) %>
+        <%= render "govuk_publishing_components/components/add_another", {
+          fieldset_legend: "Link",
+          add_button_text: "Add related external link",
+          empty_fields: true,
+          items: items,
+          empty: empty,
+        } %>
 
-      <%= render "govuk_publishing_components/components/inset_text", {
-        text: "After saving, changes to related external links will be visible on the site the next time this publication is published.",
-      } %>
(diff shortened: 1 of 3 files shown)


app/views/editions/secondary_nav_tabs/related_external_links/_add-another_checkbox.html.erb
```diff
@@ -1,5 +1,7 @@
-<%= render "govuk_publishing_components/components/checkboxes", { 
-  name: "artefact[external_links_attributes][#{index}][_destroy]",
-  id: "artefact_external_links_attributes_#{index}__destroy",
-  items: [{label: "Delete", value: "1" }],
-} %>
+<% if current_user.has_editor_permissions?(@resource) %>
```
TonyGDS, line 1: I think it would be better if we put the check in the parent page `_related_external_links.html.erb`. It is responsible for deciding to show the partial or not, rather than the partial itself.

TonyGDS (commented):
I think we are missing a check + tests on the `update_related_external_links` route to ensure users without editor permissions are unable to save i.e. manual save through the URL.

TonyGDS (commented):
I think we're missing tests in `edition_edit_test.rb` to check for `govuk_editor` vs `welsh_editor` vs `signon` for external links

app/views/editions/secondary_nav_tabs/_related_external_links.html.erb
```diff
@@ -9,30 +9,40 @@
           empty = render(partial: "secondary_nav_tabs/related_external_links/add-another_fieldset", locals: { index: 0, id: nil })
         else
           items = @edition.artefact.external_links.each_with_index.map do | external_link, index |
+            fields = render(partial: "secondary_nav_tabs/related_external_links/add-another_fieldset", locals: { index:, id: external_link.id })
+            destroy_checkbox = if current_user.has_editor_permissions?(@resource)
```
davidtrussler, line 13: Small query: do we need to put a condition on `destroy_checkbox` here? May have missed something but this seems to apply the same condition twice (including the broader one below).

PR #2724, opened 2025-06-16
727 bug fact check toggle

davidtrussler:
[Trello](https://trello.com/c/XeUFymb8/727-bug-show-earlier-messages-toggle-should-change-to-hide-earlier-messages-when-selected)

This deals with a bug on the "History and Notes" tab. Where there are "Receive fact check by GOV.UK Bot" actions listed, there is a toggle in the email text  that reads "Show earlier messages". When clicked this should expand the section below that contains extra content and should then read "Hide earlier messages". The bug is that the text of the toggle was not changing. 

The main problem was that some class names of elements were being used for both styling and behaviour. Some of these were changed for style purposes co that they were no longer useable by the JavaScript module. To follow best practice on this those have been changed so that the JS refers to one set of classes and the CSS another. 

Instances of `text` have also been updated to `textContent` within the module. This may have been a reason for the failure since `text` is not a valid property and may not have been being set as required. 

The screenshots below show the toggle on load, expanded and contracted. The only change is in the text of the toggle itself: either "Show earlier messages" or "Hide earlier messages".

Scenario|Current|Updated|
|-|-|-|
|Fact check action on load|Screenshot 2025-06-16 at 16 45 20|Screenshot 2025-06-16 at 16 41 41|
|Fact check action when expanded|Screenshot 2025-06-16 at 16 45 25|Screenshot 2025-06-16 at 16 41 32|
|Fact check action when contracted|Screenshot 2025-06-16 at 16 45 30|Screenshot 2025-06-16 at 16 41 55|

### DIFF
--- a/app/assets/javascripts/modules/history_and_notes.js
+++ b/app/assets/javascripts/modules/history_and_notes.js
@@ -17,7 +17,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     let i = 1
 
     actions.forEach(element => {
-      const commentEarlier = element.querySelector('.action--receive_fact_check--earlier') || null
+      const commentEarlier = element.querySelector('.js-earlier') || null
 
       if (commentEarlier) {
         const toggle = document.createElement('button')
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/modules/history_and_notes.js
```diff
@@ -17,7 +17,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     let i = 1
 
     actions.forEach(element => {
-      const commentEarlier = element.querySelector('.action--receive_fact_check--earlier') || null
+      const commentEarlier = element.querySelector('.js-earlier') || null
```
TonyGDS, line 20: Just curious, what was the reason the class name was changed to js?

TonyGDS (approved):
LGTM

app/assets/javascripts/modules/history_and_notes.js
davidtrussler: Just to indicate that it's only used for JavaScript and not styling, to keep them separate. That way it shouldn't get changed if there's a need to change a class name for styling.

PR #2747, opened 2025-06-26
Revert toggle for design system publications page

ellohez:
Revert design system toggle for the publications page back to false to continue use of legacy bootstrap page until improvements are made.


syed-ali-tw (approved):
LGTM

PR #2766, opened 2025-07-10
feat: exclude archived LocalTransactionEditions from blocking LocalServiceCleaner

KludgeKML:
- When all LocalTransactionEditions mentioning a given LGSL code are archived, we should allow it to be deleted so that it wont be a valid option for making new editions.
- https://trello.com/c/lGHQaN0h/694-delete-dataset-from-local-links-manager-for-specific-lgsl

### DIFF
--- a/app/lib/local_service_cleaner.rb
+++ b/app/lib/local_service_cleaner.rb
@@ -28,7 +28,7 @@ private
   end
 
   def in_local_transaction?(local_service)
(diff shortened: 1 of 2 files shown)


KludgeKML:
Have tested on integration - it correctly removes the local services that are now missing from the data/local_service.csv file but are only mentioned in archived LocalTransactionEdition records, and when you look at them in the UI they display correctly. You also can't add a new Local Transaction with the removed LGSL codes.

PR #2809, opened 2025-08-01
Make the maintenance mode message more generic

mtaylorgds:
Since the date/times for maintenance are in flux, keep the maintenance
 message more generic so we don't need to keep changing it.


syed-ali-tw (approved):
LGTM

PR #2834, opened 2025-08-20
Disable auto-merge of dependabot updates

mtaylorgds:
Prevent auto-merges so that deployments don't get triggered while we migrate data from MongoDB to PostgreSQL


ellohez (approved):
LGTM

PR #2862, opened 2025-09-05
Fix flakey test in PrimaryListingPresenterTest

mtaylorgds:
The test only seems to care that the same items are returned, irregardless of order (since the sorting for both arrays is done in the test). Since all three editions that are created will have the same, default, title, sorting by title is non-deterministic—use the `assert_same_elements` matcher instead of sorting and using the `assert_equal` one.

### DIFF
--- a/test/unit/primary_listing_presenter_test.rb
+++ b/test/unit/primary_listing_presenter_test.rb
@@ -150,13 +150,13 @@ class PrimaryListingPresenterTest < ActiveSupport::TestCase
     assert c.draft?
 
     presenter = PrimaryListingPresenter.new(Edition, bob)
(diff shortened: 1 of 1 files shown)


test/unit/primary_listing_presenter_test.rb
```diff
@@ -150,13 +150,13 @@ def setup_users
     assert c.draft?
 
     presenter = PrimaryListingPresenter.new(Edition, bob)
-    assert_equal [b, c].sort_by(&:title), presenter.draft.to_a.sort_by(&:title)
+    assert_same_elements [b, c], presenter.draft.to_a
```
ellohez, line 153: Good spot! That's a handy assertion to know about.

ellohez (approved):
Great work!

PR #2856, opened 2025-09-01
769 upgrade place edit page

farahTW:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #2865, opened 2025-09-08
Main 2118 upgrade transaction edit page

AlexRyanShep:
WIP to test github integration

### DIFF
--- a/test/integration/edition_edit_test.rb
+++ b/test/integration/edition_edit_test.rb
@@ -1832,6 +1832,76 @@ class EditionEditTest < IntegrationTest
       end
     end
 
+    context "transaction edition" do
+      setup do
+        visit_transaction_edition(state: "draft", in_beta: true)
+      end
+
+      should "show fields for transaction edition" do
+        assert page.has_field?("edition[title]", with: "Edit page title")
+        assert page.has_css?(".govuk-label", text: "Title")
+
+        assert page.has_field?("edition[overview]", with: "metatags")
+        assert page.has_css?(".govuk-label", text: "Meta tag description")
+        assert page.has_css?(".govuk-hint", text: "Some search engines will display this if they cannot find what they need in the main text")
+
+        assert page.has_field?("edition[introduction]", with: "Transaction introduction")
+        assert page.has_css?(".govuk-label", text: "Introduction")
+        assert page.has_css?(".govuk-hint", text: "Set the scene for the user. What is about to happen? For example, “you will need to fill in a form, print it out and take it to the post office”. Refer to the Govspeak guidance (opens in new tab).")
+
+        assert page.has_field?("edition[start_button_text]")
+        assert page.has_css?(".govuk-label", text: "Start button text")
+        assert find(".gem-c-radio input[value='Start now']").checked?
+
+        assert page.has_field?("edition[will_continue_on]", with: "To be continued...")
+        assert page.has_css?(".govuk-label", text: "Text below the start button (optional)")
+        assert page.has_css?(".govuk-hint", text: "Following ‘on’, for example “the HMRC website”")
+
+        assert page.has_field?("edition[link]", with: "http://continue.com")
+        assert page.has_css?(".govuk-label", text: "Link to start of transaction")
+        assert page.has_css?(".govuk-hint", text: "Link as deep as possible")
+
+        assert page.has_field?("edition[more_information]", with: "Transaction more information")
+        assert page.has_css?(".govuk-label", text: "More information (optional)")
+
+        assert page.has_field?("edition[alternate_methods]", with: "Method A or B")
+        assert page.has_css?(".govuk-label", text: "Other ways to apply (optional)")
+        assert page.has_css?(".govuk-hint", text: "Alternative ways of completing this transaction")
+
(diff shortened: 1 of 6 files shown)


AlexRyanShep:
This will need rebasing to main after https://github.com/alphagov/publisher/pull/2856 is merged

app/views/editions/secondary_nav_tabs/edit/draft/_transaction.html.erb
```diff
@@ -0,0 +1,85 @@
...
+    text: "Introduction",
+  },
+  name: "edition[introduction]",
+  value: edition.introduction,
+  rows: 16,
+  hint: ("Set the scene for the user. What is about to happen? For example, “you will need to fill in a form, print it out and take it to the post office”. Refer to the #{link_to("Govspeak guidance (opens in new tab)", "https://govspeak-preview.publishing.service.gov.uk/guide", target: "_blank", rel: "noopener", class: "govuk-link")}").html_safe,
+} %>
+
+<%= render "govuk_publishing_components/components/radio", {
+  heading: "Start button text",
```
davidtrussler, line 17: We need to add `heading_level: 0` here to prevent the legend text being rendered as a `<h2>`

davidtrussler: We should also make the same change in `_beta_content_radio.html.erb`, which is not part of this PR.

app/views/editions/secondary_nav_tabs/edit/published/_transaction.html.erb
```diff
@@ -0,0 +1,80 @@
...
+<%= render partial: "editions/secondary_nav_tabs/edit/published/components/meta_tag_description", locals: { edition: @resource } %>
+
+<section>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Introduction",
+    heading_level: 3,
+    font_size: "s",
+  } %>
+  <p>
+    <%= edition.introduction %>
```
davidtrussler, line 12: i think we should have some alternative text here (and in the other sections that follow) for where there is no value for this. In other content types we have used "None added".

test/integration/edition_edit_test.rb
```diff
@@ -1783,6 +1783,123 @@ class EditionEditTest < IntegrationTest
...
+        assert page.has_css?(".govuk-hint", text: "Link as deep as possible")
+
+        assert page.has_field?("edition[more_information]", with: "Transaction more information")
+        assert page.has_css?(".govuk-label", text: "More information (optional)")
+
+        assert page.has_field?("edition[alternate_methods]", with: "Method A or B")
+        assert page.has_css?(".govuk-label", text: "Other ways to apply (optional)")
+        assert page.has_css?(".govuk-hint", text: "Alternative ways of completing this transaction")
+
+        assert page.has_field?("edition[need_to_know]", with: "Transaction need to")
```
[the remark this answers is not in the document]
AlexRyanShep: Not sure what happened there, I must have accidentally deleted it while neatening the formatting! Thanks for the catch

app/views/editions/secondary_nav_tabs/edit/published/_transaction.html.erb
AlexRyanShep: I've added this message in for my place view, and also for the title, body and meta tag description views that were all also missing it

mtaylorgds: Maybe we should run this past Jonny/Chirag (could do that as part of the desk check, if you like).

app/views/editions/secondary_nav_tabs/edit/published/_transaction.html.erb
```diff
@@ -0,0 +1,80 @@
...
+<%= render partial: "editions/secondary_nav_tabs/edit/published/components/meta_tag_description", locals: { edition: @resource } %>
+
+<section>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Introduction",
+    heading_level: 3,
+    font_size: "s",
+  } %>
+  <p>
+    <%= edition.introduction != "" ? edition.introduction : "None added" %>
```
davidtrussler, line 12: I am finding that this needs to check for `nil` rather than an empty string to render the alternate text. Is that right? 😕

AlexRyanShep: @mtaylorgds could you also try this please? (create an `TransactionEdition` with an empty introduction, then see what `.introduction` returns against it?) Had a chat with David just now and we're definitely getting different results (`nil` for him, `""` for me). We're wondering if the difference could be he's running it against the MongoDB version, and I'm running against the PG version

app/views/editions/secondary_nav_tabs/edit/published/_transaction.html.erb
AlexRyanShep: I think that makes sense to do in the desk check yeah. It's easy enough to remove if they don't want it.

AlexRyanShep:
@mtaylorgds @davidtrussler New commits added covering the logic change, expanding that to the place type and other components as discussed, and some additional tests

davidtrussler (approved):
LGTM! 👍

PR #2877, opened 2025-09-12
(CM-441) Update links to new Content Block Manager

pezholio:
We are moving Content Block Manager to its own app, so we need to update the URLs.

~~I'll leave this in draft until the move is complete.~~


mtaylorgds (approved):
LGTM

PR #2884, opened 2025-09-17
Answer and help edit page read only for non editors

PeterHattyar:
[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/MAIN/boards/1555?selectedIssue=MAIN-2098)

These changes 
- make sure the document in any given state is not-editable for a user without editor permissions.
- also makes sure a user without editor permissions doesn't see the Send to 2i link in the relevant edition states.
- however it makes the preview link available for them to peruse while a document is not published or archived.


davidtrussler (approved):
Looks good to me - nice work! 👍

PR #2891, opened 2025-09-19
2080 upgrade local transaction edit page

farahTW:


AlexRyanShep (dismissed):
Left a few comments for review, mostly just little style tweaks, a fix for the LGSL code not displaying in the proper font, and a fix for the failing unit tests

app/views/editions/secondary_nav_tabs/edit/editable/components/_administration_availability.html.erb
```diff
@@ -0,0 +1,30 @@
+<%= render "govuk_publishing_components/components/radio", {
+  heading: heading,
+  name: "edition[#{administration}_availability_attributes][authority_type]",
+  id_prefix: "conditional-checked",
```
davidtrussler, line 4: We should delete the `id_prefix` parameter: this is giving us multiple identical IDs and isn't used for anything. The component will ensure that we have unique IDs for all the radios.

app/views/editions/secondary_nav_tabs/edit/editable/_local_transaction.html.erb
```diff
@@ -0,0 +1,96 @@
...
+<%= render "govuk_publishing_components/components/input", {
+  label: {
+    text: "LGIL code",
+  },
+  id: "lgil",
+  name: "edition[lgil_code]",
+  value: edition.lgil_code,
+  heading_size: "s",
+  width:2,
+  class: "govuk-input--extra-letter-spacing",
```
davidtrussler, line 23: We should add an error message parameter here and link it to the error via the id `lgil_code` rather than the value used here.

app/views/editions/secondary_nav_tabs/edit/editable/_local_transaction.html.erb
```diff
@@ -0,0 +1,96 @@
...
+} %>
+
+<%= render "govuk_publishing_components/components/input", {
+  label: {
+    text: "Button text (optional)",
+  },
+  id: "button text",
```
davidtrussler, line 30: It is probably better not to supply an `id` parameter here and for the other components where we do not need a specific value since the component itself will create a unique id in each case.

app/views/editions/secondary_nav_tabs/edit/editable/components/_administration_availability.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    {
+      value: "devolved_administration_service",
+      text: "Service available from devolved administration (or a similar service is available)",
+      checked: edition.editionable.send("#{administration}_availability").authority_type == "devolved_administration_service",
+      conditional: render("govuk_publishing_components/components/input", {
+        label: {
+          text: "URL of the devolved administration website page",
+        },
+        name: "edition[#{administration}_availability_attributes][alternative_url]",
+        value: edition.editionable.send("#{administration}_availability").alternative_url,
```
davidtrussler, line 20: We should add an error message value here and an id that matches the `href` in the error message at the top of the page.

app/views/editions/secondary_nav_tabs/edit/editable/_local_transaction.html.erb
```diff
@@ -0,0 +1,96 @@
...
+  name: "edition[cta_text]",
+  value: edition.editionable.cta_text,
+  heading_size: "s",
+  class: "govuk-input--extra-letter-spacing",
```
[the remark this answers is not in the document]
davidtrussler: We still need to delete this class parameter. I don't think we want it here anyway but it won't be used by the component.

app/views/editions/secondary_nav_tabs/edit/readonly/_local_transaction.html.erb
```diff
@@ -0,0 +1,81 @@
...
+  <p><%= edition.editionable.need_to_know.presence || "None added" %></p>
+</section>
+
+<section>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Above results content (optional)",
+    heading_level: 3,
+    font_size: "s",
+  } %>
+  <p> <%= edition.editionable.before_results.presence || "None added" %></p>
```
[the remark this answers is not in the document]
davidtrussler: Looks like this still needs doing.

app/views/editions/secondary_nav_tabs/edit/editable/components/_administration_availability.html.erb
```diff
@@ -0,0 +1,32 @@
+<%= render "govuk_publishing_components/components/radio", {
+  heading: heading,
+  heading_level: 0,
```
mtaylorgds, line 3: Is there a reason we're setting to level `0`, rather than `3`, like we do for the other headings? @davidtrussler should it not be a heading (albeit an `h3`, not an `h2`)?

app/views/editions/secondary_nav_tabs/edit/readonly/components/_administration_availability.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  <% if edition.editionable.send("#{administration}_availability").alternative_url.presence %>
+  <section>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "URL of the devolved administration website page",
+      heading_level: 3,
+      font_size: "s",
+    } %>
+    <p><%= edition.editionable.send("#{administration}_availability").alternative_url || "None added" %></p>
+  </section>
+    <% end %>
```
mtaylorgds, line 17: This line should be unindented one level.
```suggestion
  <% end %>
```

app/views/editions/secondary_nav_tabs/edit/editable/_local_transaction.html.erb
```diff
@@ -0,0 +1,91 @@
+<%= render partial: "editions/secondary_nav_tabs/edit/editable/components/title", locals: { edition: @resource } %>
+
+<div class="govuk-form-group">
+<%= render "govuk_publishing_components/components/label", {
```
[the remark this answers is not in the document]
farahTW: initially I have used the same thing based on the David's comment I have changed this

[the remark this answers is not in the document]
davidtrussler: Yes I pondered this but I'm happy for it to be a `<label>` since that keeps it consistent with the rest of the page. That element doesn't need to be used in conjunction with a form input (though it obviously usually is) so I think this is the least confusing markup in the context of the page as a whole, particularly for assistive technology.

app/views/editions/secondary_nav_tabs/edit/editable/components/_administration_availability.html.erb
davidtrussler: I think this is better to not mark this up as a heading to be consistent with the other labels on the page. I think it is more semantically confusing to use headings for these inputs and not the others.

PR #2895, opened 2025-09-24
MAIN-4926 edition history bug

AlexRyanShep:
Updates the new diff view to use `@resource.history.first` for the back button, so that it correctly takes the user back to the most recent revision after viewing a diff


davidtrussler (approved):
LGTM! 👍

PR #2896, opened 2025-09-24
Add GA4 parameters for user

davidtrussler:
[Jira](https://gov-uk.atlassian.net/browse/MAIN-2122) 

This is a part of the larger piece of work to implement GA4 tracking across Publisher and augments the work that has already been done to implemented `page_view` events across the whole app. These changes add some values for the user to that event. 

To add these we need to 
- update the Gemfile.lock to ensure we have the latest changes for `gds-sso` which provides a value for the `anonymous_user_id` parameter
- update the User model to add a value for the user `role` parameter
- update the design system layout to add the relevant metatags to provide the user data for the GA4 analytics code


app/models/user.rb
```diff
@@ -49,6 +49,17 @@ def unassign(edition)
     GovukContentModels::ActionProcessors::AssignProcessor.new(self, edition).processed_edition
   end
 
+  def role
```
ellohez, line 52: Does it matter that users can have multiple permissions? This seems to reduce that to just one role?

ellohez (approved):
Questions to be discussed at a later date, happy to approve for now.

ellohez:
[Jira](https://gov-uk.atlassian.net/browse/MAIN-2122)

This is a part of the larger piece of work to implement GA4 tracking across Publisher and augments the work that has already been done to implemented page_view events across the whole app. These changes add some values for the user to that event.

To add these we need to

update the Gemfile.lock to ensure we have the latest changes for gds-sso which provides a value for the anonymous_user_id parameter
update the User model to add a value for the user role parameter
update the design system layout to add the relevant metatags to provide the user data for the GA4 analytics code

PR #2899, opened 2025-09-26
Main 1203 postgres cleanup remove migration task and files

hezzimo:
[Jira](https://gov-uk.atlassian.net/browse/MAIN-1203?atlOrigin=eyJpIjoiZWM1NGM0MjQ3MzAzNDhkZDk4NGE3OGExMzg5YzYzNzUiLCJwIjoiaiJ9)

Now that the Mongo to PostgreSQL migration is complete, any now unnecessary migration and setup code needs to be removed. This will make the codebase cleaner, safer and reduce test suite run times. 

This PR removes the migration helper files, tests and associated test data, the import task and the logging code. A copy of which will be archived in case needed.


Tetrino (approved):
Looks like you got everything!

PR #2909, opened 2025-10-01
Update error message at local transaction page.

farahTW:
This PR updates the existing error message for local transaction page.

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/MAIN/boards/1555?selectedIssue=MAIN-2080)

It simply amends one of the error messages that incorrectly starts with a lowercase letter, to make sure it starts with a capital letter – so “Must” instead of “must”

**Before:**
image

**After:**
image


mtaylorgds (approved):
Approved, but are we missing test coverage around this validation?

PR #2916, opened 2025-10-06
Upgrade Guide edition page to design system

farahTW:
This PR upgrades guide edition page to design system.
[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/MAIN/boards/1555?selectedIssue=MAIN-1290)

It adds a new view for guide edition.
Updates the existing tests.
Adds the new integration tests.

**Screen shots**

**Draft version**

image

**Published version:**
image

**When user has no editor permission:**
image


farahTW:
**Note:** There is less space or no space between heading Chapters and the radio option ,this will be gone once Add Chapter button will get added  believe.

app/views/editions/secondary_nav_tabs/edit/readonly/_guide.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  font_size: "m",
+} %>
+</section>
+
+<section>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Is every chapter part of a step by step?",
+    heading_level: 3,
+    font_size: "s",
+  } %>
```
davidtrussler, line 18: I don't think we should render this content ("Is every chapter part of a step by step?") as a `<h3>` element given that it is (if I understand the page structure correctly) a subsection of ("Chapters"). I think I'd be happy to render this as a `<label>` as in a previous similar instance but I think we should consider all these pages together from an accessibility perspective as a further piece of work.

PR #3188, opened 2026-03-10
Content Block and HTML Renderer service

AlexRyanShep:
[MAIN-7277](https://gov-uk.atlassian.net/browse/MAIN-7277)

This PR creates a new HtmlRenderer service class which can take in a govspeak document, use Govspeak::Document to render it as HTML, and then use ContentBlockTools to replace all valid content block embed codes with their HTML value.

The GenericEditionPresenter is also updated to use this new service class, and the FCM Api Adapters have been updated with new documentation detailing the payload standard for current_content and previous_content.

This should not be merged before [MAIN-7353](https://gov-uk.atlassian.net/browse/MAIN-7353), otherwise it will prevent valid test data from being sent over.

[MAIN-7277]: https://gov-uk.atlassian.net/browse/MAIN-7277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAIN-7353]: https://gov-uk.atlassian.net/browse/MAIN-7353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/lib/gds_api/fact_check_manager.rb
+++ b/app/lib/gds_api/fact_check_manager.rb
@@ -10,8 +10,13 @@ class GdsApi::FactCheckManager < GdsApi::Base
   # @option [string] source_url The url locating the content on the source application
   # @param [string] requester_name The username of the source app user submitting the request
   # @param [string] requester_email The email address of the source app user submitting the request
-  # @param [hash] current_content Hash containing HTML content being fact checked
-  # @option [hash] previous_content Hash containing HTML content of previous content version to check against
+  # @param [hash] current_content Hash of current content to be compared in HTML format
+  #   For simple single-part documents format: { id: { heading: "heading string", body: "HTML content string" } }
+  #   For complex multi-part/multi-chapter documents format: { id: { heading: "heading string", body: "HTML content string" }, id2: { heading: "heading string", body: "HTML content string" }, ...}.
+  #   Heading or body may change between current_content and previous_content, but the ID for the parts must match to allow for comparison.
+  #   If a part has been deleted entirely, do not provide its ID in current_content.
+  # @option [hash] previous_content Same format as current_content
+  #   If a new part has been added to the document, do not provide its ID in previous_content.
   # @option [string] deadline Date a response is requested by. Use iso8601 date format: "2026-02-09"
   # @param [array] recipients Array of emails to be notified of the request
   # @option [uuid] draft_auth_bypass_id The edition's auth_bypass_id for draft origin preview access
@@ -58,7 +63,10 @@ class GdsApi::FactCheckManager < GdsApi::Base
   # Keyword Arguments:
   # @param [string] source_app identifier for the application sending the request
   # @param [uuid] source_id The unique ID for the content
-  # @param [hash] current_content
+  # @param [hash] current_content Hash of current content to be compared in HTML format
+  #   For simple single-part documents format: { id: { heading: "heading string", body: "HTML content string" } }
+  #   For complex multi-part/multi-chapter documents format: { id: { heading: "heading string", body: "HTML content string" }, id2: { heading: "heading string", body: "HTML content string" }, ...}.
+  #   If a part has been deleted entirely, do not provide its ID in current_content.
   # @option [string] source_title The title of the content (optional)
   # @option [uuid] draft_auth_bypass_id The edition's auth_bypass_id for draft origin preview access (optional)
   # @option [string] draft_slug The edition's slug for the draft origin preview URL path (optional)
--- a/app/presenters/formats/generic_edition_presenter.rb
+++ b/app/presenters/formats/generic_edition_presenter.rb
@@ -3,7 +3,7 @@ module Formats
     def render_for_fact_check_manager_api
       return unless @edition.respond_to?(:whole_body)
 
-      { body: @edition.whole_body.presence || "" }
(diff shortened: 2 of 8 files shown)


test/unit/services/html_renderer_test.rb
```diff
@@ -0,0 +1,70 @@
+require "test_helper"
+
+class HtmlRendererTest < ActiveSupport::TestCase
+  context ".render_hash" do
```
young-rob, line 4: It could be worth having a couple of assertions for the results of `.render_hash` when called with single and multi-item hashes

young-rob (approved):
Looking great!

test/unit/services/html_renderer_test.rb
AlexRyanShep: Thanks Rob! I've changed these tests so that they're just testing the output.

It's impossible to combine the original assertion (spying on method calls) and actually testing the output because Minitest has no equivalent to RSpec's `.and_call_original` nor inbuilt spy functionality. I'm so glad we've moved to RSpec for FCM!

I could expand it to four tests I guess, but that feels like overkill. The tests as they are give full coverage of the service class.

app/lib/gds_api/fact_check_manager.rb
```diff
@@ -10,19 +10,20 @@ class GdsApi::FactCheckManager < GdsApi::Base
   # @option [string] source_url The url locating the content on the source application
   # @param [string] requester_name The username of the source app user submitting the request
   # @param [string] requester_email The email address of the source app user submitting the request
-  # @param [hash] current_content Hash containing HTML content being fact checked
-  # @option [hash] previous_content Hash containing HTML content of previous content version to check against
+  # @param [hash] current_content Hash of current content to be compared in HTML format
+  #   For simple single-part documents format: { id: {heading: "heading string", bodu:'HTML content string'} }
+  #   For complex multi-part/multi-chapter documents format: { id: {heading: "heading string", bodu:'HTML content string'}, id2: {heading: "heading string", bodu:'HTML content string'}, ...}.
```
young-rob: Couple of small typos in here and maybe we should be consistent with double/single quote usage and spacing

```suggestion
  # @param [hash] current_content Hash of current content to be compared in HTML format
  #   For simple single-part documents format: { id: { heading: "heading string", body: "HTML content string" } }
  #   For complex multi-part/multi-chapter documents format: { id: { heading: "heading string", body: "HTML content string" }, id2: { heading: "heading string", body: "HTML content string" }, ...}.
```

app/lib/gds_api/fact_check_manager.rb
```diff
@@ -58,16 +56,15 @@ def post_resend_emails(source_app:, source_id:)
   # Keyword Arguments:
   # @param [string] source_app identifier for the application sending the request
   # @param [uuid] source_id The unique ID for the content
-  # @param [hash] current_content
+  # @param [hash] current_content Hash of current content to be compared in HTML format
+  #   For simple single-part documents format: { id: {heading: "heading string", bodu:'HTML content string'} }
+  #   For complex multi-part/multi-chapter documents format: { id: {heading: "heading string", bodu:'HTML content string'}, id2: {heading: "heading string", bodu:'HTML content string'}, ...}.
```
young-rob: Same as above around typos/spacing

```suggestion
  #   For simple single-part documents format: { id: { heading: "heading string", body: "HTML content string" } }
  #   For complex multi-part/multi-chapter documents format: { id: { heading: "heading string", body: "HTML content string" }, id2: { heading: "heading string", body: "HTML content string" }, ...}.
```

app/lib/gds_api/fact_check_manager.rb
AlexRyanShep: Agh, good catch thanks! Fixed and pushed

app/lib/gds_api/fact_check_manager.rb
AlexRyanShep: As above, fixed and pushed

app/presenters/formats/generic_edition_presenter.rb
```diff
@@ -3,7 +3,7 @@ class GenericEditionPresenter < EditionFormatPresenter
     def render_for_fact_check_manager_api
       return unless @edition.respond_to?(:whole_body)
 
-      { body: @edition.whole_body.presence || "" }
+      HtmlRenderer.render_hash({ content: { heading: "Body", body: @edition.whole_body.presence || "" } })
```
young-rob: I think the way you've implemented `HtmlRenderer.render_html` [handles the case](https://github.com/alphagov/publisher/pull/3188/changes#diff-eb66ecfae967da9709e67e3f34fafd0f0f5da46d3f6cb7e0b69fa9029a0b90ddR9) that `whole_body` is blank so I think you should be able to remove the fallback?
```suggestion
      HtmlRenderer.render_hash({ content: { heading: "Body", body: @edition.whole_body.presence } })
```

young-rob (approved):
LGTM 👍

app/presenters/formats/generic_edition_presenter.rb
AlexRyanShep: I've made this change 👍

PR #3229, opened 2026-04-07
[MAIN-7358] Bug fix - "In review" message

young-rob:
[MAIN-7358](https://gov-uk.atlassian.net/browse/MAIN-7358)

### Changes

In places, the view code calling `latest_status_action` assumes that the action returned has a certain `request_type` and associated attributes, when this is not always the
case. This ensures that the most recent action of the assumed
`request_type` is retrieved, preventing errors and incorrectly displayed messages (e.g. "blank sent this to be reviewed").

A [composite index](https://www.postgresql.org/docs/current/indexes-multicolumn.html) is added to the `actions` table on the columns: `edition_id`, `request_type` and `created_at`. [The index is added using `algorithm: :concurrently` to prevent write-locks whilst it is being built](https://thoughtbot.com/blog/how-to-create-postgres-indexes-concurrently-in). Queries using the following combinations of these columns should benefit from this index:
- `edition_id`
- `edition_id` and `request_type`
- `edition_id`, `request_type` and `created_at`.

The `latest_status_action` method is updated to leverage this new index. It also uses `order(:created_at)` rather than the default ordering via ID (the IDs are sequential so the ordering should be the same but I was able to find some records for which this was not the case in integration, possibly due to timing issues when they were imported from mongodb).

[MAIN-7358]: https://gov-uk.atlassian.net/browse/MAIN-7358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/traits/recordable_actions_test.rb
+++ b/test/traits/recordable_actions_test.rb
@@ -63,4 +63,28 @@ class RecordableActionsTest < ActiveSupport::TestCase
       assert_nil object_under_test.superseded_at
     end
   end
+
+  describe "#latest_status_action" do
+    it "returns the action of that type even when a newer action of a different type exists" do
+      edition = FactoryBot.create(:edition, :in_review)
+      edition.new_action(user, Action::SEND_FACT_CHECK)
+
+      assert_equal Action::REQUEST_REVIEW, edition.latest_status_action(Action::REQUEST_REVIEW).request_type
+    end
+
+    it "returns nil when no action of that type exists" do
+      edition = FactoryBot.create(:edition, :in_review)
+
+      assert_nil edition.latest_status_action(Action::SEND_FACT_CHECK)
+    end
+
+    it "returns the latest status action when no type argument provided" do
+      edition = FactoryBot.create(:edition, :in_review)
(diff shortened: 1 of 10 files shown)


test/traits/recordable_actions_test.rb
```diff
@@ -63,4 +63,19 @@ class RecordableActionsTest < ActiveSupport::TestCase
...
+      edition.new_action(user, Action::SEND_FACT_CHECK)
+
+      assert_equal Action::REQUEST_REVIEW, edition.latest_status_action(Action::REQUEST_REVIEW).request_type
+    end
+
+    it "returns nil when no action of that type exists" do
+      edition = FactoryBot.create(:edition, :in_review)
+
+      assert_nil edition.latest_status_action(Action::SEND_FACT_CHECK)
+    end
```
hezzimo: Maybe I am missing something but do we need a test for the else case if there are actions but type is not specified?

hezzimo (approved):
LGTM! Just one query?

PR #3232, opened 2026-04-08
Main 7352 ga4 form error events

davidtrussler:
[MAIN-7352](https://gov-uk.atlassian.net/browse/MAIN-7352)

This adds GA4 tracking for form_error events and involves the changes below. Primarily it
- puts all instances of the "Error summary" component into the single existing `error_summary` partial
- updates the `error_summary` partial to add the required data-attributes, also putting the `ga4-auto-tracker` module behind the `GA4 Form Tracking` feature flag
- makes some small changes to the views involved to use consistent values for input fields required by use of the partial
- references the required `ga4-auto-tracker` module
- updates existing integration tests to add tests for data fired by error messages
- adds new integration tests for areas not so far covered

Expected data values are listed in the Jira ticket. 

Tracking can be tested locally with these steps: 
- ensure that the "Ga4 form tracking" feature is set to "On" via FlipFlop
- run `window.dataLayer` in the browser console after submitting the form. The console will display all the data that will be sent to be tracked when the user submits the form. In this work we are interested in the "event_data" events. There should be one of these for each error listed in the error summary.

[MAIN-7352]: https://gov-uk.atlassian.net/browse/MAIN-7352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/editions/secondary_nav_tabs/tagging_remove_breadcrumb_page.html.erb
+++ b/app/views/editions/secondary_nav_tabs/tagging_remove_breadcrumb_page.html.erb
@@ -6,18 +6,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <% unless @edition.errors.empty? %>
-      <% content_for :error_summary do %>
-        <%= render("govuk_publishing_components/components/error_summary", {
-          id: "error-summary",
-          title: "There is a problem",
-          items: @edition.errors.map do |error|
-            {
-              text: error.message,
-              href: "##{error.attribute.to_s}",
-            }
-          end,
-        }) %>
-      <% end %>
+      <% content_for :error_summary, render("shared/error_summary", { object: @edition, full_messages: false }) %>
     <% end %>
 
     <%= form_with model: @tagging_update_form_values, url: remove_breadcrumb_edition_path(@resource), data: { module: "" } do |f| %>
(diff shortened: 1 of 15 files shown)


app/views/editions/secondary_nav_tabs/tagging_remove_breadcrumb_page.html.erb
young-rob, line 1: I think in the shared error summary, the error links are generated like this:
`href: "##{model.to_s.underscore + "_" + error.attribute.to_s}"`
which seems to be a bit different than how they were being generated here and in the other places you've swapped it in as part of this PR - are they still generated correctly (as in, do they match the form input id)?

davidtrussler: Yes, that was a slightly annoying side effect of using the shared error_summary, but I have changed the value for `id` on the matching form inputs. Mostly that has meant changing "title" to "edition_title" or "part_title" and so on, but I checked they were still linking the summary to the input field in each case. There are some (like "Tag related content") where this is already not working as it should so in those cases it has been updated but it still doesn't work: there should be a card in the backlog somewhere to fix that.

young-rob (approved):
LGTM 👍

PR #3251, opened 2026-04-15
Update `flash` partial to add GA4 data-attributes to flash messages

davidtrussler:
[MAIN-7345](https://gov-uk.atlassian.net/browse/MAIN-7345)

Add GA4 event tracking to Flash messages by adding the following data-attributes to each instance of a flash message across Publisher as described in the Jira ticket: 
- the `ga4-auto-tracker` data-module
- parameters for the `data-ga4-auto` attribute: 
  - event_name
  - text
  - action

It feels like this should include some tests but I am unsure what tests we would add for this. Adding integration test for every instance of a flash message (there are probably around 150 across Publisher) to check that each instance fires the correct event on page load seems a little excessive. For reference there are existing functional test associated with various controllers in the tests below which could be repurposed as integration tests but would involve alot of work. 
- artefacts_controller_test.rb
- content_item_controller_test.rb
- downtimes_controller_test.rb
- editions_controller_test.rb
- guide_parts_controller_test.rb
- homepage_controller_test.rb
- notes_controller_test.rb
- tagging_controller_test.rb

This can tested manually as below: 
- run `window.dataLayer` in the browser console after a page with a flash message loads. 
- In this work we are interested in the "event_data" events. There should be one of these for the flash message with the values listed in the Jira ticket.

[MAIN-7345]: https://gov-uk.atlassian.net/browse/MAIN-7345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/shared/_flash.html.erb
+++ b/app/views/shared/_flash.html.erb
@@ -1,15 +1,35 @@
+<% component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns) %>
+
 <% [:success, :info, :warning, :danger, :notice, :alert].select { |k| flash[k].present? }.each do |k| %>
+  <% data_attributes = {
+    module: "ga4-auto-tracker",
+    "ga4-auto": {
+      event_name: "flash_#{k.to_s}",
+      action: "#{k.to_s}_alerts",
+      text: sanitize(flash[k]),
+    },
+  } %>
+
   <% if k == :warning || k == :danger || k == :alert %>
-    <%= render "govuk_publishing_components/components/error_alert", {
-      message: sanitize(flash[k]),
-    } %>
+    <%= tag.div(**component_helper.all_attributes) do %>
+      <%= render "govuk_publishing_components/components/error_alert", {
+        message: sanitize(flash[k]),
+        data_attributes: data_attributes,
+      } %>
+    <% end %>
   <% elsif k == :notice || k == :info %>
-    <%= render "govuk_publishing_components/components/notice", {
-      description: sanitize(flash[k]),
-    } %>
+    <%= tag.div(**component_helper.all_attributes) do %>
+      <%= render "govuk_publishing_components/components/notice", {
+        description: sanitize(flash[k]),
+        data_attributes: data_attributes,
+      } %>
+    <% end %>
   <% elsif k == :success %>
-    <%= render "govuk_publishing_components/components/success_alert", {
-      message: sanitize(flash[k]),
-    } %>
+    <%= tag.div(**component_helper.all_attributes) do %>
+      <%= render "govuk_publishing_components/components/success_alert", {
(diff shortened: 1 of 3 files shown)


AlexRyanShep:
The code here looks good to me, though as per your slack message not actually approving this PR yet without tests.

As far as tests go, my personal opinion would be to agree with you that refactoring all existing tests to be integration style checking for these new data attributes feels excessive.

The way I've always thought about things is to isolate what exactly you're trying to test in each situation. In those existing tests we are testing that a flash is raised, and sometimes we're checking the content of the visual message provided, but in the context of those tests we don't really care about this GA4 tracking.

Personally, the approach I would take would be:

- Some isolated unit style tests directly against this view. Maybe rendering the view in isolation and manually inserting/invoking some test cases for each of the event_data events? In other work I'd be doing this kind of testing against a controller but there's not one in play here.
- A selection of integration style tests in a more realistic environment, verifying that this works properly in more realistic scenarios/when rendered in a full page using a simple scenario we know will provoke a flash. Rather than trying to test all different places we use these flashes, maybe just spot-pick a scenario of two for each type of event_data?

But yeah, regardless of whether you go that direction or not, I agree that adding new tests for all 150 scenarios here feels excessive. Given that flash is a built in feature of rails and all of ours use this same shared template, verifying that the template itself works + a couple of integration tests just in case feels like enough to me.

davidtrussler:
Thanks for your comment @AlexRyanShep 

I've gone through it again and added a new integration test to capture the main flash message events we get and some tests for each of the "success", "warning" and "danger" flash messages. Actually there's no "Alert" messages and only one "Warning" message, the main ones are "Success" and "Danger" so I've captured a variety of those. 

I'm not sure of the best way to add a unit test to this scenario but I'm thinking it may not be critical with the tests that are there now but let me know if you think we should still cover that.

AlexRyanShep:
Looking with fresh eyes today, I think I probably went too much into my default frame of thinking (which is very backend focused generally) about the unit tests 😅 What you've got here looks good to me!

AlexRyanShep (approved):
Looks good to me!

Am I understanding this right (pretty sure I am) that the datalayer/get_event_data stuff is a non-visible layer which is what GA4 picks up, so separate from the normal integration tests which are testing against the text visible on the screen?

davidtrussler:
> Looks good to me!
> 
> Am I understanding this right (pretty sure I am) that the datalayer/get_event_data stuff is a non-visible layer which is what GA4 picks up, so separate from the normal integration tests which are testing against the text visible on the screen?

Yes, user actions or - in this case - tracking data on page load get pushed to a `dataLayer` object which is not visible to the user. The tests run a `evaluate_script("window.dataLayer")` method to get that data and tests that it is what we expect.

PR #3252, opened 2026-04-15
[MAIN-7370] Add draft preview fields to fact-check-manager API payloads

young-rob:
[MAIN-7370](https://gov-uk.atlassian.net/browse/MAIN-7370)

### Changes
An edition's `auth_bypass_id`, `content_id` and `slug` are required to generate the draft origin preview link in fact-check-manager. These fields are added to the API payload for `post_fact_check`. The `auth_bypass_id` and `slug` are added to `patch_fact_check` to allow changes to their values to be pushed to fact-check-manager.

These fields are optional for the gds-api methods but always sent from Publisher via the service that calls them.

[MAIN-7370]: https://gov-uk.atlassian.net/browse/MAIN-7370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AlexRyanShep (approved):
LGTM!

PR #3263, opened 2026-04-20
Remove legacy root controller and views

young-rob:
### Changes

The legacy root has now been retired in favour of the design system interface, and can no longer be accessed. This code can safely be deleted.


stacy-hayes (approved):
LGTM!

PR #3267, opened 2026-04-21
Increase Rack QueryParser params_limit

young-rob:
We have some SimpleSmartAnswer records ([example](https://publisher.integration.publishing.service.gov.uk/editions/e31d89cf-605e-45cb-b152-82265195bda5)) which have a very large number of nodes. When attempting to save these records, the user sees an error because the size of the params exceeds the default limit put in place by Rack's QueryParser (4096). Increasing the `params_limit` will prevent this.


PeterHattyar (approved):
Amazing, looks good!

PR #3275, opened 2026-04-27
Refactor and add new fields to Fact Check Manager API calls

AlexRyanShep:
[MAIN-7360](https://gov-uk.atlassian.net/browse/MAIN-7360)

This PR:

- Refactors the FactCheckManagerApiService class into a FactCheckRequestForm, using the Form Object pattern to extract validation logic out of the EditionsController
- Updates the send_to_fact_check flow and api adapter method/documentation with the new fields `reason_for_change` and `zendesk_ticket`
- Updates the send_to_fact_check_page view with **minimal** new UI to enable testing and validation. Full UI to follow in a follow-up ticket
- Refactors the resend and update flows to use the new Form Object
- Removes the old service class and tests
- Fixes a bug with the old email validation regex for both new API and legacy Gmail flows

[MAIN-7360]: https://gov-uk.atlassian.net/browse/MAIN-7360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/editions_controller.rb
young-rob, line 1: For consideration:

I don't think we currently need to use a form object when performing `update_fact_check` and `resend_fact_check_email`; neither action accepts user input which we need to validate and both have fairly simple payloads. We could probably still get away with deleting the service class and just call the API adapters from the controller?

The validations that run for these actions are:

`validates :edition, :user, presence: true` (I'm not sure we need this given we set them in the controller?)
`validate :user_has_editor_permissions` (already runs in the controller)

Having a form object validate and construct their payloads might well be something we want to do later on if we add user input, but right now I'm wary of premature optimisation, giving the new form object too many responsibilities, and increasing the scope of this piece of work.

It could be simpler to keep the scope of form object to `send_to_fact_check` to begin with (as we definitely need all the defined validations for that) and add the others later if we need to.

app/lib/govuk_content_models/action_processors/send_fact_check_processor.rb
```diff
@@ -4,17 +4,28 @@ class SendFactCheckProcessor < BaseProcessor
       def process
         return false if action_attributes[:email_addresses].blank?
 
-        if Flipflop.enabled?(:fact_check_manager_api) && FactCheckManagerApiService.request_fact_check(@edition, @actor, action_attributes[:email_addresses]).is_a?(GdsApi::HTTPErrorResponse)
-          return false
+        if Flipflop.enabled?(:fact_check_manager_api)
+          form = action_attributes[:fact_check_request_form]
+          return false unless form.valid?(:send) && Services.fact_check_manager_api.post_fact_check(**form.post_new_request_payload)
```
young-rob: I don't think we need to validate the form again after having done it in the controller

app/controllers/editions_controller.rb
AlexRyanShep: Thanks Rob 🙂 Copying my response over from the DMs

On this point, I think I do disagree. I like the neatness of the Form object having its clear responsibility of preparing and validating those payloads for FCM API calls in general, and I think as it stands its set up in a way that will make any future expansion easy. They wouldn't need to write much in the way of new methods or tests, mostly just modify what's there.

If we were back in planning stage and the work hadn't been done yet I think i'd agree with you for the sake of time 😂  But as it is the work has been done, it would be extra work to take it out and re-work the tests to account for that, and I don't see that we would really gain anything from it vs losing that ease of expanding things later?

(I don't really feel too strongly, if you do)

app/lib/govuk_content_models/action_processors/send_fact_check_processor.rb
AlexRyanShep: Thanks Rob 🙂 Copying my response over from the DMs

My only concern with not running valid? again before submission is just what if something somehow changes with the edition or user between the form page being submitted, and that running. I don't believe in Publisher we have any concept of locking down objects like that while someone else is doing something with them.  It's an extremely unlikely scenario but it technically could happen I guess? 

I think it's best/common practice if we're trying to stick to the pattern with form objects to have that last minute validity check just in case, and it's a fairly low resource operation happening very infrequently so I don't think it'd have a performance impact.

Happy to take your call on that 🙂  What do you think?

app/controllers/editions_controller.rb
young-rob: Sounds good, let's leave it in (I do agree for the resend emails we will likely need to validate some user input in phase 2 or later, and the update payload is big enough to move).

What do you think about removing those 2 validations - perhaps we don't need to run any validation at all when updating or resending as there is no user input to validate, and the user permissions are checked in the controller?

app/lib/govuk_content_models/action_processors/send_fact_check_processor.rb
young-rob: I think the things we're re-validating are still in the same request cycle and still in-memory (i.e. not reloaded) so I'm not sure how the result of the validation could change between the two calls, even if the edition got deleted or the user lost permissions or whatever at a database level.

This discussion has made me start thinking: the processor just picks the payload off the form object so maybe we shouldn't be passing the form to the processor, but rather just the payload itself?

e.g. 
```
def send_to_fact_check_for_edition(resource, fact_check_request_form, comment)
  progress_edition(resource, { request_type: "send_fact_check", comment:, email_addresses: fact_check_request_form.email_addresses, api_payload: fact_check_request_form.post_new_request_payload })
end
```
(and the code in the processor wouldn't really change much)

app/lib/fact_check_request_form.rb
```diff
@@ -0,0 +1,127 @@
...
+                                             message: "Zendesk number must be a number at least 7 digits long",
+                                             allow_blank: true }, on: :send
+
+  validate :user_has_editor_permissions
+  validate :valid_email_addresses, on: :send
+  validate :deadline_in_range, on: :send
+
+  def initialize(*args)
+    super
+  end
```
young-rob: I think this could be removed as it's just calling the super

app/controllers/editions_controller.rb
```diff
@@ -476,10 +502,17 @@ def setup_view_paths
 private
 
   def update_fact_check
-    FactCheckManagerApiService.update_fact_check_content(@resource)
+    form = FactCheckRequestForm.new({ edition: @resource, user: current_user })
+
+    unless form.valid?(:update) && Services.fact_check_manager_api.patch_update_content(**form.update_content_payload)
+      Rails.logger.error "Request form validation errors: #{form.errors.full_messages.join(', ')}"
+      flash[:danger] = "Due to a service problem, the fact check request could not be updated. The edition was successfully saved"
```
young-rob: I think this should be a `flash.now` (so it only persisted for the one request)

app/controllers/editions_controller.rb
```diff
@@ -476,10 +502,17 @@ def setup_view_paths
...
+      render "show"
+      return
+    end
   rescue GdsApi::HTTPErrorResponse => e
-    Rails.logger.error "Error #{e.class} #{e.message}"
-    @resource.errors.add(:show, "Due to a service problem, the fact check request could not be updated. The edition was successfully saved")
+    Rails.logger.error "API Error Response for Edition id #{@resource.id}: #{e.class} #{e.message}"
+    flash[:danger] = "Due to a service problem, the fact check request could not be updated. The edition was successfully saved"
```
young-rob: Same as above for `flash.now`

app/lib/fact_check_request_form.rb
```diff
@@ -0,0 +1,127 @@
...
+  def user_has_editor_permissions
+    return if user.blank?
+
+    errors.add(:user, "You do not have permission to edit this content") unless user.has_editor_permissions?(edition)
+  end
+
+  def valid_email_addresses
+    return if email_addresses.blank?
+
+    email_regex = /\A[\w+\-%.]+@[a-z\d-]+(\.[a-z\d-]+)*\.[a-z]+\z/
```
young-rob: Might be worth defining this as a constant. I know we also seem to define email regex's elsewhere in the code - could/should they be shared?

app/lib/fact_check_request_form.rb
```diff
@@ -0,0 +1,127 @@
...
+  end
+
+  def update_content_payload
+    {
+      source_app: SOURCE_APP,
+      source_id: edition.id,
+      source_title: edition.title,
+      current_content: current_content_presenter.render_for_fact_check_manager_api,
+      draft_auth_bypass_id: @edition.auth_bypass_id,
+      draft_slug: @edition.slug,
```
young-rob: These two can use `edition` rather than `@edition`
```suggestion
      draft_auth_bypass_id: edition.auth_bypass_id,
      draft_slug: edition.slug,
```

app/controllers/editions_controller.rb
young-rob: Might be worth saving this error message in a constant too

app/lib/gds_api/fact_check_manager.rb
```diff
@@ -10,23 +10,25 @@ class GdsApi::FactCheckManager < GdsApi::Base
...
   #   For complex multi-part/multi-chapter documents format: { id: { heading: "heading string", body: "HTML content string" }, id2: { heading: "heading string", body: "HTML content string" }, ...}.
   #   Heading or body may change between current_content and previous_content, but the ID for the parts must match to allow for comparison.
   #   If a part has been deleted entirely, do not provide its ID in current_content.
   # @option [hash] previous_content Same format as current_content
   #   If a new part has been added to the document, do not provide its ID in previous_content.
-  # @option [string] deadline Date a response is requested by. Use iso8601 date format: "2026-02-09"
-  # @param [array] recipients Array of emails to be notified of the request
+  # @param [string] deadline Date a response is requested by. Use iso8601 date format: "2026-02-09"
+  # @option [string] reason_for_change optional free-text response of why the change has been made, to populate the notification email. Line breaks are retained.
+  # @option [integer] zendesk_number optional number for zendesk ticket linked to the change
```
young-rob, line 23: I think the form object has the zendesk number as a string at the mo - do we need to convert it in the payload so it's an integer?

app/views/editions/secondary_nav_tabs/send_to_fact_check_page.html.erb
```diff
@@ -6,22 +6,106 @@
...
+          label: {
+            heading_size: "m",
+            text: "Email addresses",
+          },
+          name: "fact_check_request_form[email_addresses]",
+          id: "email_addresses",
+          value: @form.email_addresses,
+          type: "email",
+          hint: "Separate multiple email addresses with a comma or semi-colon, followed by a space",
+          error_items: errors_for(@form&.errors, :email_addresses, use_full_message: false),
```
young-rob: I think we could get rid of the safe navigation operators here as the form object is assigned in the controller. If you want to keep them to be sure, you'll probably want to add them in all the places in this form that `@form` is being accessed (e.g. `@form.email_addresses`)

app/views/editions/secondary_nav_tabs/send_to_fact_check_page.html.erb
```diff
@@ -6,22 +6,106 @@
...
+            text: "Zendesk number",
+          },
+          name: "fact_check_request_form[zendesk_number]",
+          id: "zendesk_number",
+          value: @form.zendesk_number,
+          prefix: "https://govuk.zendesk.com/tickets/",
+          error_items: errors_for(@form&.errors, :zendesk_number, use_full_message: false),
+        } %>
+
+        <% deadline_autofill = @form.deadline_autofill || working_days_after(Date.current, how_many: 5) %>
```
young-rob: Would it be simpler if we let the date fields in the view deal with the date inputs separately e.g. using and submitting params `deadline_3i`, `deadline_2i`, `deadline_1i` (with their `value`s accessing attributes we've defined with those names on the form object) and let the form object worry about stitching together `deadline` for the payload, throwing validation errors if it can't do that with the params the user has submitted?

It might mean we don't have to worry about setting `deadline_autofill` and all the logic around it. In the form object you could have something like:

```
  def deadline
    return unless deadline_1i.present? && deadline_2i.present? && deadline_3i.present?

    Time.zone.local(deadline_1i.to_i, deadline_2i.to_i, deadline_3i.to_i).to_date
  rescue ArgumentError, TypeError
    nil
  end
```
and the presence validation for `deadline` would be triggered if the params submitted are bad?

test/functional/editions_controller_test.rb
```diff
@@ -773,6 +773,15 @@ class EditionsControllerTest < ActionController::TestCase
...
           assert_equal "fact_check", @edition.state
         end
+
+        should "not call Services.fact_check_manager_api" do
+          patch :resend_fact_check_email, params: {
+            id: @edition.id,
+          }
+
+          assert_equal "Fact check email re-sent", flash[:success]
+          Services.fact_check_manager_api.expects(:post_resend_emails).never
```
young-rob: Does `Services.fact_check_manager_api.expects(:post_resend_emails).never` need to be before the patch call?

test/functional/editions_controller_test.rb
```diff
@@ -1119,6 +1137,9 @@ class EditionsControllerTest < ActionController::TestCase
     end
 
     context "#send_to_fact_check" do
+      setup do
+      end
```
young-rob: Empty setup block, can delete

test/functional/editions_controller_test.rb
```diff
@@ -1139,17 +1163,33 @@ class EditionsControllerTest < ActionController::TestCase
           end
         end
 
+        should "call Services.fact_check_manager_api if the edition to be saved is in 'ready' state" do
+          edition = FactoryBot.create(:edition, :ready)
+          Services.fact_check_manager_api.expects(:post_fact_check).with(has_entries(source_id: edition.id)).returns(GdsApi::Response.new(code: 200))
```
young-rob: Should this expectation also verify the new fields we're sending across (deadline, zendesk number etc)?

test/functional/editions_controller_test.rb
```diff
@@ -2575,4 +2665,10 @@ class EditionsControllerTest < ActionController::TestCase
   def description(edition)
     edition.format.underscore.humanize.downcase
   end
+
+  def parametrised_deadline
```
young-rob: nit: typo in this name
```suggestion
  def parameterised_deadline
```

test/unit/lib/fact_check_request_form_test.rb
```diff
@@ -0,0 +1,280 @@
...
+  context ".resend_emails_payload" do
+    should "build and format the payload" do
+      expected_payload = { source_app: "publisher",
+                           source_id: @edition.id }
+
+      assert_equal expected_payload, @form.resend_emails_payload
+    end
+  end
+
+  context ".update_content_payloaad" do
```
young-rob: nit: typo
```suggestion
  context ".update_content_payload" do
```

test/functional/editions_controller_test.rb
stacy-hayes: I think it definitely should have deadline since thats a required field, not sure on zendesk number etc since they're optional

app/lib/gds_api/fact_check_manager.rb
stacy-hayes: > If this is typecast to integer, then non-integer values would be cast to 0 and accepted. Numericality handles non-int strings.
>   attribute :zendesk_number, :string, default: nil
>   

  looking at the form it looks like numericality casts them to integers after submissions so I think it's fine as is?

app/controllers/editions_controller.rb
stacy-hayes: Yeah I'm happy to remove those validations

app/lib/govuk_content_models/action_processors/resend_fact_check_processor.rb
young-rob, line 1: I think we can also remove the form validation in this processor and pass it `api_payload` in the `action_attributes` rather than a form object, like you have done in `send_fact_check_processor.rb`

app/controllers/editions_controller.rb
```diff
@@ -476,10 +503,17 @@ def setup_view_paths
 private
 
   def update_fact_check
-    FactCheckManagerApiService.update_fact_check_content(@resource)
+    form = FactCheckRequestForm.new({ edition: @resource, user: current_user })
+
+    unless form.valid?(:update) && Services.fact_check_manager_api.patch_update_content(**form.update_content_payload)
```
young-rob, line 508: I think we could remove this call to `form.valid?(:update)` as there are no validations for this context now?

young-rob (approved):
Great work!

PR #3277, opened 2026-04-29
Pass edition to path helper in reviewer_edit_link

young-rob:
### Changes

If the `edit_reviewer_edition_path` path helper is called whilst creating a new guide part (e.g. if it is "in 2i"), the edition ID cannot be retrieved from `params[:id]`. The edition is now passed to the path helper to account for this.


stacy-hayes (approved):
Really nice catch!

PR #3285, opened 2026-05-05
Fix exception handler for Notify errors in `ApplicationController`

hezzimo:
- Logic for this contradicted the comment and would surely result in exceptions being raised in non-prod environments rather than being handled gracefully.
- User (dev) friendly 'Team only key' errors would be raised in prod, not in staging or integration

Hopefully, this PR can correct the above so that we can use the team-only Notify key more effectively in Integration.

Needs manual testing in integration to ensure that we can send emails to team members.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -34,19 +34,17 @@ class ApplicationController < ActionController::Base
   end
 
   def notify_bad_request(exception)
-    # TODO: control this via the log level rather than an environment variable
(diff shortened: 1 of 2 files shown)


young-rob (approved):
Looking good - worth deploying this to integration and seeing what happens when you try to submit to an invalid recipient with the team only api key

young-rob:
We're going to make a ticket for this and pick it up in the next maintenance sprint

PR #3288, opened 2026-05-06
Main 7209 correct section values

davidtrussler:
[MAIN-7209](https://gov-uk.atlassian.net/browse/MAIN-7209)

The changes here correct a wrong value that was being pushed to the dataLayer when forms are submitted: the "section" value of the "form_response" event. Originally this was set at a "global" level to replicate the page title but a different value is required for a number of forms, as specified in the Jira ticket. This value is inferred from the `data-ga4-section` attribute. 

To ensure the correct value is reported:
- the `data-ga4-section` attribute is removed from the base layout view
- the correct value is added to the `data-ga4-section` in each view that requires it
- the correct value replaces the current one in the appropriate integration tests

[MAIN-7209]: https://gov-uk.atlassian.net/browse/MAIN-7209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/editions/secondary_nav_tabs/_edit.html.erb
+++ b/app/views/editions/secondary_nav_tabs/_edit.html.erb
@@ -48,7 +48,7 @@
     <% end %>
   </div>
 <% else %>
-  <%= form_for @resource, as: :edition, url: edition_path(@resource), data: {module: "unsaved-changes-prompt"} do %>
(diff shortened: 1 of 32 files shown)


hezzimo (approved):
LGTM!

PR #3293, opened 2026-05-08
Allow empty strings for `comment` in FactCheckResponseContract

young-rob:
Fact Check Manager responses send over an empty string for `comment` when there is no comment on the response (rather than omitting the key or sending `nil`). This change permits that behaviour rather than causing a validation error.


hezzimo (approved):
Looks good, nice fix

PR #3299, opened 2026-05-11
[MAIN-7413] Render Parted editions as hash of parts for fact check API

young-rob:
[MAIN-7413](https://gov-uk.atlassian.net/browse/MAIN-7413)

## Context
GenericEditionPresenter#render_for_fact_check_manager_api previously called whole_body for every edition, collapsing all parts of a Parted edition (e.g. GuideEdition) into a single body.

The Fact Check Manager API expects Parted content as a hash keyed by part slug, with each entry containing a part's title and rendered body, so that fact checkers can review each chapter separately.

## Changes

- Retrieve the content type-specific edition presenter for fact check content hashes via the existing `EditionPresenterFactory.get_presenter` method, rather than using GenericEditionPresenter.

- Move .render_for_fact_check_manager_api into EditionFormatPresenter, which is the base class for the edition presenters. We can now override this method in the content type-specific presenters if desired.

- Override .render_for_fact_check_manager_api for Guide editions in GuidePresenter to separate out the parts into a hash keyed by part slug.

- Include the edition title in the content hashes so that it is included in the diff in Fact Check Manager. The titles are sent as h2 elements to retain the heading hierarchy when rendered on FCM, and the `.edition-title` class is applied to edition titles so that FCM stylesheets can easily apply h1 stylings as required.

image

[MAIN-7413]: https://gov-uk.atlassian.net/browse/MAIN-7413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


davidtrussler (approved):
LGTM! 👍

PR #3303, opened 2026-05-14
Send to Fact Check Page (single page design)

AlexRyanShep:
[MAIN-7360](https://gov-uk.atlassian.net/browse/MAIN-7360)

This PR adds a semi-dynamic email preview to the Send to Fact Check page.This email is mostly static, following the one-page design, but does populate the edition title.

- Added a couple of new integration tests. Didn't exhaustively test for presence of all the content, but key dynamic elements.
- I've had to use direct hex codes rather than govuk-functional-colour due to incompatibilities with Publisher. That's been added to the tech debt backlog to investigate.

image

[MAIN-7360]: https://gov-uk.atlassian.net/browse/MAIN-7360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
```diff
@@ -0,0 +1,60 @@
...
+
+            Only one fact check response can be submitted.
+
+            Do not reply to this email.
+          GOV_SPEAK
+        %>
+      <% end %>
+    </div>
+  </div>
+</div>
```
PeterHattyar: I think this file might be missing a trailing empty line at the end?

AlexRyanShep: Thanks Peter, good catch! I've pushed up a fixed version with the newline

PeterHattyar (approved):

app/assets/stylesheets/edit.scss
```diff
@@ -28,6 +28,34 @@
...
+  &--send-fact-check-page {
+    @extend .govuk-body; // stylelint-disable-line scss/at-extend-no-missing-placeholder
+  }
+
+  &--send-fact-check-email-preview {
+    @extend .govuk-body; // stylelint-disable-line scss/at-extend-no-missing-placeholder
+
+    border: 1px solid #cecece; // Temporary workaround until govuk-functional-colour is working
+    background-color: #f4f8fb; // Temporary workaround until govuk-functional-colour is working
+    padding: 0 15px;
```
davidtrussler: We should try to use [design system units](https://design-system.service.gov.uk/styles/spacing/
) rather than specific pixel values where possible. Here it would be something like `@include govuk-responsive-padding(3, "left");` and so on.

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
```diff
@@ -0,0 +1,60 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds edit--send-fact-check-page">
```
davidtrussler: This could probably be contained in a semantic element like [article](https://html.spec.whatwg.org/multipage/sections.html#the-article-element) rather than a `<div>` element, given that it has its own structure and repeats a few headings that should be kep a bit distinct from those on the containing page.

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
```diff
@@ -0,0 +1,60 @@
...
+            Reason for change: Your input will go here.
+
+            Zendesk ticket: [https://govuk.zendesk.com/tickets/1234567](https://govuk.zendesk.com/tickets/1234567)
+
+            ^ Deadline: DD Month YYYY
+
+            ---
+
+            ## Review the changes
+            [View the fact check - #{@edition.title}](https://www.gov.uk/)
```
davidtrussler: I think links and email addresses should not be functional but should be styled like those elements. Might be a bit awkward to do though.

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
```diff
@@ -0,0 +1,60 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds edit--send-fact-check-page">
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "Example email",
+      heading_level: 2,
+      font_size: "m",
+    } %>
+
+    <p>This shows the format of the fact check email that will be sent.</p>
```
davidtrussler: We should probably add a `govuk-body` class to the `<p>` here and remove the `@extend` from the stylesheet.

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
AlexRyanShep: Not sure it would actually be possible with the nature of the rendering; I can't think of a way to make the Govspeak renderer apply the tags etc without actually creating a link?

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
```diff
@@ -0,0 +1,60 @@
...
+
+    <div class="edit--send-fact-check-email-preview">
+      <table class="logo-presentation">
```
davidtrussler: I see why you have used the `<table>` layout from the email template but while that's standard practice for emails I don't think we should reproduce it here. I had a quick look and I think, if we use an `<article>` element as a container we can use a `<header>` element to contain the logo and heading, something like this: 
```
<header>
  <p class="govuk-body">Government Digital Service</p>
</header>
```

Then we can add styles for that to render the logo as a background image something like below, using the values that you have used. The image is probably better as a background as it is decorative. 
```
header {
  background: url("gds_logo.png") no-repeat;
  background-size: contain;
  background-position: top 0 left 7px;
  padding-left: 45px;
  border-left: solid 2px #005abb;
}
```

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
davidtrussler: > Not sure it would actually be possible with the nature of the rendering; I can't think of a way to make the Govspeak renderer apply the tags etc without actually creating a link?

Yes, I think that's true unfortunately. At this stage we can maybe leave it for Desk Check but I think the requirement was to render them without any functionality. That's from memory though - I don't think it's documented. But let's see if that's the case in the Desk Check.

AlexRyanShep: As said in the desk check, I did a bit of testing and Govspeak will preserve HTML tags given to it as input, so I should hopefully be able to apply a span (?) tag to these links and have that inherit govuk-link styling.

davidtrussler (approved):
Nice, looks good! 👍

PR #3312, opened 2026-05-20
Content block rendering duplication bugfix

AlexRyanShep:
Fixes a bug in `HtmlRenderer` when a document contains multiple occurences of the same content block embed code

`render_html `was operating off a list of every single embed code in the document, which meant when there were multiple occurrences of the same one it ran over that block twice. Since the rendered block content includes the code, this caused duplication.

Fixed this by limiting the list running that iteration to unique entries.

### DIFF
--- a/test/unit/services/html_renderer_test.rb
+++ b/test/unit/services/html_renderer_test.rb
@@ -58,19 +58,43 @@ class HtmlRendererTest < ActiveSupport::TestCase
       assert_equal expected_output, HtmlRenderer.render_html(input)
     end
 
-    should "call ContentBlockTools::ContentBlock.from_embed_code when a content block embed code is present" do
-      input = "Before block, {{embed:content_block_contact:content-item}}, after block"
-
-      # ContentBlockTools has some internal logic and an API call that need stubbing out here
-      mock_content_block = Minitest::Mock.new
-      mock_content_block.expect(:render, "block content")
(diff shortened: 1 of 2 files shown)


test/unit/services/html_renderer_test.rb
```diff
@@ -58,19 +58,43 @@ class HtmlRendererTest < ActiveSupport::TestCase
...
 
-      # ContentBlockTools has some internal logic and an API call that need stubbing out here
-      mock_content_block = Minitest::Mock.new
-      mock_content_block.expect(:render, "block content")
-      ContentBlockTools::ContentBlock.expects(:from_embed_code).with("{{embed:content_block_contact:content-item}}").returns(mock_content_block)
+    context "when the document contains content blocks" do
+      setup do
+        @mock_render = '<span class="content-block" data-embed-code="{{embed:content_block_contact:content-item}}">block content</span>'
+        @mock_content_block = stub(render: @mock_render)
+        @mock_render_two = '<span class="content-block" data-embed-code="{{embed:content_block_contact:content-item}}">block content</span>'
```
young-rob: Would it be worth making `@mock_render_two` slightly different from `@mock_render_one` just so we can be sure they're coming out in the right order?

AlexRyanShep: Agh! That was originally how I wrote it, but at some point during my 4th or 5th time rewriting those snippets to make them work with Minitest's weird handling of stubs and scope I must've missed it on the copy & paste.

Fixed now

young-rob (approved):
LGTM 👍

PR #3319, opened 2026-05-22
[MAIN-7397] Parallelise tests

young-rob:
[MAIN-7397](https://gov-uk.atlassian.net/browse/MAIN-7397)
[MAIN-1236](https://gov-uk.atlassian.net/browse/MAIN-1236)

- Enable Rails' built-in parallel test runner with workers set to `:number_of_processors`. This brings down the time taken for the test suite to run locally and on CI from ~9 minutes to ~4 minutes.

- Following the [pattern on Whitehall](https://github.com/alphagov/whitehall/blob/c94022d66e82a727eeb572036091092995966f2f/test/test_helper.rb#L72), add parallelize_setup/parallelize_teardown hooks to give each test worker a unique SimpleCov command_name, so worker coverage results merge instead of overwriting each other.

- Rewrite test files that were using Minitest::Spec::DSL (describe/it/let). The anonymous classes that DSL creates cannot be Marshalled across worker processes, which the parallel runner requires.

- Remove id from create params in guide_parts_controller_test. Explicitly setting the id isn't something the create action does in reality, and was causing test flakiness when the id was already taken.

[MAIN-7397]: https://gov-uk.atlassian.net/browse/MAIN-7397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MAIN-1236]: https://gov-uk.atlassian.net/browse/MAIN-1236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/functional/guide_parts_controller_test.rb
+++ b/test/functional/guide_parts_controller_test.rb
@@ -65,7 +65,7 @@ class GuidePartsControllerTest < ActionController::TestCase
 
           post :create, params: {
             edition_id: edition.id,
-            part: { id: 1, body: "test body", slug: "test-slug", title: "test title" },
+            part: { body: "test body", slug: "test-slug", title: "test title" },
             save: "save",
           }
 
@@ -78,7 +78,7 @@ class GuidePartsControllerTest < ActionController::TestCase
 
           post :create, params: {
             edition_id: edition.id,
-            part: { id: 1, body: "test body", slug: "test-slug", title: "test title" },
+            part: { body: "test body", slug: "test-slug", title: "test title" },
             save: "save and summary",
           }
 
@@ -97,7 +97,7 @@ class GuidePartsControllerTest < ActionController::TestCase
 
           post :create, params: {
             edition_id: edition.id,
-            part: { id: 1, body: "test body", slug: "test-slug", title: "test title" },
+            part: { body: "test body", slug: "test-slug", title: "test title" },
             save: "save and summary",
           }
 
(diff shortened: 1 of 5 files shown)


hezzimo (approved):
Looks great! Amazing work, this will make so much difference. 

One question - was moving the local variables out to the setup blocks necessary to parallelise the tests, or was that general clean up? It's looking much better either way, just curious.

young-rob:
> Looks great! Amazing work, this will make so much difference.
> 
> One question - was moving the local variables out to the setup blocks necessary to parallelise the tests, or was that general clean up? It's looking much better either way, just curious.

Thanks for asking this question, it's made me read into it a bit more to better understand it:

Those tests kept failing after I set the tests to be parallelised and after a bit of an investigation it seems to be because they had `extend Minitest::Spec::DSL` - which allowed them to be written using `let/describe/it` (a bit like RSpec). Apparently, behind the scenes this leverages anonymous (i.e. unnamed - try doing Class.new) classes to create an isolated class for each `describe` context.

When the test are parallelised, the tests run in separate worker processes which run the tests and send the test results back to the parent process. The parent process and the worker processes are memory isolated so they can't communicate by referring to in-memory Ruby objects (like the test result objects).

They get around this by having the worker process serialise the test result object to bytes using this library called [Marshal](https://rubyapi.org/4.0/o/marshal), and having the parent process reconstruct the bytes into a test result object. However, it needs a class name to do this, so it can't do that for an anonymous class because it has no name, causing errors.

There might be another way around this but I thought the easiest thing would just be to rewrite those tests ☺️

PR #3333, opened 2026-06-02
Fix invalid email address issue in FCM

hezzimo:
[MAIN-7460](https://gov-uk.atlassian.net/browse/MAIN-7460)

Upgrade Regex to account for apostrophes and use a standard method for regex - .match?

- This method is purely for checking if a pattern exists, is more memory efficient and therefore faster.
- Allows us to use the negation at the start of the method chain, to improve readability.
- Uses .to_s to avoid any issues with address being nil (as to_s will convert this to an empty string)

[MAIN-7460]: https://gov-uk.atlassian.net/browse/MAIN-7460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/lib/fact_check_request_form.rb
+++ b/app/lib/fact_check_request_form.rb
@@ -25,7 +25,7 @@ class FactCheckRequestForm
   validate :deadline_in_range, on: :send
   validate :deadline_present, on: :send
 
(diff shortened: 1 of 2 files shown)


young-rob (approved):
LGTM 👍

PR #3343, opened 2026-06-09
[MAIN-7249] Call edition#order_parts upon deleting a part

young-rob:
[MAIN-7249](https://gov-uk.atlassian.net/browse/MAIN-7249)

### Changes

We do not currently bump down the order numbers for a guide edition's parts upon deleting a part e.g. deleting the middle part for a guide edition having part order values of 1,2,3 results in the remaining parts having order 1,3. This is unlikely to result in an ordering issue because we do this whenever we add a new part or via the reorder functionality, however it is good to keep the orders in the right state.

The deletion and re-ordering are wrapped in a transaction so they're atomic.

[MAIN-7249]: https://gov-uk.atlassian.net/browse/MAIN-7249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/guide_parts_controller.rb
+++ b/app/controllers/guide_parts_controller.rb
@@ -102,13 +102,14 @@ class GuidePartsController < InheritedResources::Base
 
   def destroy
     @part = Part.find(params[:id])
-    if @part.destroy!
-      UpdateWorker.perform_async(@edition.id.to_s)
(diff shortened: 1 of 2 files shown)


baisa (approved):
LGTM

PR #3350, opened 2026-06-11
Remove legacy controllers, views and, tests

AlexRyanShep:
[MAIN-7427](https://gov-uk.atlassian.net/browse/MAIN-7427)

This PR is a first pass at attempting to remove as much legacy content as possible from Publisher. A large quantity of the tests could not be removed, but some redundant tests were found and deleted, along with a number of views and two controllers (LegacyArtefactsController and PublicationsController)

Supporting docs:

- [Investigation into PublicationsController](https://gov-uk.atlassian.net/wiki/spaces/GOVUK/pages/5488967718/PublicationsController+-+What+is+it+what+does+it+do+can+we+delete+it)
- [Code to be removed/reviewed in the future](https://gov-uk.atlassian.net/wiki/spaces/GOVUK/pages/5481332742/Legacy+code+to+be+deleted)

[MAIN-7427]: https://gov-uk.atlassian.net/browse/MAIN-7427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/routes.rb
```diff
@@ -19,9 +19,9 @@
 
   get "downtimes" => "downtimes#index"
 
-  resources :artefacts, only: %i[create new]
+  resources :artefacts, only: %i[create update new]
   post "artefacts/new/content-details", to: "artefacts#content_details", as: :new_artefact_content_details
-  resources :artefacts, controller: "legacy_artefacts", only: %i[update]
+  # resources :artefacts, controller: "legacy_artefacts", only: %i[update]
```
young-rob: We should delete this completely
```suggestion
```

config/routes.rb
young-rob: We might as well remove the routes (`resources :publications`) now the controller has been removed

app/controllers/artefacts_controller.rb
```diff
@@ -71,6 +102,7 @@ def slug_with_prefix(artefact_params)
   def create_artefact_and_edition
     ActiveRecord::Base.transaction do
       if @artefact.update_as(current_user, artefact_params.merge({ slug: slug_with_prefix(artefact_params) })) && create_edition.persisted?
+        UpdateWorker.perform_async(@artefact.latest_edition.id.to_s)
```
young-rob: Maybe instead of calling this here, we should be calling it from the `create` method if this method returns true, once the records have been committed to the database (this is the what we're doing elsewhere) e.g. in `create`

```
if create_artefact_and_edition
  UpdateWorker.perform_async(@artefact.latest_edition.id.to_s)
  redirect_to edition_path(@artefact.latest_edition)
else...
```

config/routes.rb
AlexRyanShep: Yes definitely! Sorry, this got missed in all the other changes when I was tidying up after testing.

app/controllers/artefacts_controller.rb
AlexRyanShep: Can do! 

I don't think there's any actual functional difference (where it is now, this Async call is happening after the `create_edition.persisted?` check which is what checks the records have been saved right? So, at this point, we're onto the success branch and off the Rollback branch, but it is a bit neater where you've suggested it.

young-rob: Yeah I think it's probably not a problem, just best practice as the transaction hasn't been committed yet so perhaps there's a small chance that the Sidekiq worker (which I believe has a separate connection to the DB) wouldn't be able to find the record if the job runs really quickly before the commit happens (e.g. if maybe at some point we decide to do some other stuff in that transaction block which is a bit slower).

[This was an interesting read](https://medium.com/@esperonleticia/transactional-push-the-sidekiq-feature-you-need-to-enable-right-now-d3fc1787f6ce)

young-rob (approved):
LGTM 👍

app/controllers/artefacts_controller.rb
AlexRyanShep: That is an interesting little read, could be something that we should consider turning on? It looks like it was introduced in version 6.5 and we're running 7.x I believe in most of our apps

PR #3355, opened 2026-06-12
Fix ordering of new simple smart answer options

young-rob:
### Changes

When a user clicked "Add answer" on a question, the new option was being saved with `order` 1 rather than a value placing it after the existing answers; this resulted in it appearing at the top of the list rather than at the end.

This was happening because the hidden order field for each option is rendered with the value `o.index.to_i + 1` which is fine for options already persisted but the nested_form gem renders the new options with an index of "new_options" which results in 0 when calling `to_i`, so every dynamically added option had an order of 1.

This change sets the order of newly added options via Javascript by finding the highest order of the question's answers.

### DIFF
--- a/app/assets/javascripts/smart-answers.js
+++ b/app/assets/javascripts/smart-answers.js
@@ -123,7 +123,20 @@
       }
     },
     initOption: function (e) {
-      var node = $(e.field).parents('.node').first()
+      var option = $(e.field)
+      var node = option.parents('.node').first()
+      var maxOrder = 0
+
+      // Find the maximum order value in the list of existing options
+      node.find('.option:visible .option-order')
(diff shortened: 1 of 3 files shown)


davidtrussler (approved):
LGTM 👍 

I guess this won't be included when we update SSA to the Design System: it could do with updating somewhat and also with adding a JS Unit Test. But as it stands this looks fine to me.

young-rob:
> LGTM 👍
> 
> I guess this won't be included when we update SSA to the Design System: it could do with updating somewhat and also with adding a JS Unit Test. But as it stands this looks fine to me.

Yes, I'm looking forward to deleting all this jQuery once we've migrated it 🪚

PR #3366, opened 2026-06-18
Ga4 refactor form setup

davidtrussler:
The changes here refactor elements of GA4 From Tracking in order to make this more stable (particularly in tests).

These involve
- moving where the `ga4-form-tracker` data-module is applied from the `ga4-form-setup` JS module to the various forms which require it. Since this is a module that is initialised on page load it makes more sense and feels less flaky to do it in the view. Every view that requires it needed to have an empty data-module adding for this to work anyway so this does not add superfluous code to those views.
- also adding the `ga4-link-tracker` data-module to views where it is needed.
- updating the `ga4-form-setup` JS module to remove where the data-module was being added. The addition of various data-attributes is left here since that is not needed until the user begins to interact with the form. There are also some data-attributes added for the `ga4-link-tracker` data-module. 
- updating the `ga4-form-setup` unit test to reflect the changes in the module. 
- some tidying up of other aspects of some of the views such as adding the `data-ga4-section` attribute where that had been left out of previous changes

Since all the GA4 work is pretty comprehensively covered by automated tests I am confident that these changes will do what is required without any breakages.


hezzimo (approved):
LGTM!

PR #3371, opened 2026-06-24
Main 7408 ga4 send to fact check page

davidtrussler:
[MAIN-7408](https://gov-uk.atlassian.net/browse/MAIN-7408)

This adds the required GA4 tracking on the "Send to fact check" page (with the "Fact check manager api" feature toggle set to "On"), as listed in the Jira ticket (see the comment on 17 June 2026 at 11:07).

In fact the tracking is set correctly in previous work (form and error tracking) so this PR just adds a new test to the existing integration tests for the page. 

The values for user interactions can be tested locally by 
- running the following in the browser console: 
  - `window.GOVUK.analyticsGa4.showDebug = true` to enable real time debugging
  - `document.addEventListener('click',function(e){if(e.target.href)(e.preventDefault())}` to prevent page navigation when clicking on links
  - document.querySelector('form').addEventListener('submit',function(e){e.preventDefault()})` to prevent page navigation on form submission

To test the values for error messages the form must submit so the page should be refreshed and `window.dataLayer` can be run in the browser console for each error. Then the values for `event_data` can be inspected. It is expected that a single `form_error` event should be generated for each error. 

[MAIN-7408]: https://gov-uk.atlassian.net/browse/MAIN-7408?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/integration/ga4_tracking_edit_test.rb
+++ b/test/integration/ga4_tracking_edit_test.rb
@@ -535,7 +535,7 @@ class Ga4TrackingEditTest < JavascriptIntegrationTest
     end
   end
 
-  context "Send to fact check page" do
+  context "Send to fact check page (without Fact check manager api)" do
     setup do
       stub_holidays_used_by_fact_check
       @edition.state = "ready"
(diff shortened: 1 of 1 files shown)


Tetrino (approved):
LGTM

PR #3407, opened 2026-07-28
Update govuk_sidekiq to 12.0.0

young-rob:
### Changes

- Update govuk_sidekiq to 12.0.0
- **Remove Sidekiq pin from Gemfile** - We originally pinned Sidekiq to <7 to opt out of the v7 beta, but we are now using version 7+ so we no longer need to pin to a specific version. We can remove Sidekiq from the Gemfile and allow it to be brought in as a dependency of our other gems which require it.


stacy-hayes (approved):
LGTM!

PR #3411, opened 2026-07-29
Obtain content block for fact check from Publishing API rather than Content Store

edavey:
## Refactor: Obtain content block for fact check from Publishing API rather than Content Store

Jira [CR-37](https://gov-uk.atlassian.net/browse/CR-37)

---

[RFC-192: Restrict base_paths to a smaller character set][rfc-192] is restricting GOV.UK  base paths to `a-z 0-9 . / -` (no underscores, no uppercase). Content blocks are affected because their `base_path` includes the `document_type` (e.g. `content_block_time_period`) which contains underscores, as in this `base_path`:

```
/content-blocks/content_block_time_period/tax-year
```

Before the change in this PR, when rendering content blocks to send for fact check, Publisher was getting the block `#details` from Content Store. The process used `base_path` within the Content Block Tools gem via the call to `ContentBlockTools::ContentBlock.from_embed_code(code)` which constructs the (invalid) `base_path` to make the call to the Content Store:

`ContentBlockTools::ContentBlockReference`:
```
def content_store_identifier
  "/content-blocks/#{document_type}/#{identifier}"
end
```

With this change Publisher's `HtmlRenderer`, instead of building a renderable block with a single call to `ContentBlockTools::ContentBlock.from_embed_code(code).render`, now:

- obtains the block `details` from Publishing API
- builds a renderable block using `ContentBlockTools::ContentBlock.new(...).render`

[rfc-192]:
https://github.com/alphagov/govuk-rfcs/pull/192/

[CR-37]: https://gov-uk.atlassian.net/browse/CR-37?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/html_renderer.rb
+++ b/app/services/html_renderer.rb
@@ -9,12 +9,32 @@ class HtmlRenderer
     return "" if document.blank?
 
     html_content = Govspeak::Document.new(document).to_html
+    references = ContentBlockTools::ContentBlockReference.find_all_in_document(html_content).uniq
 
-    ContentBlockTools::ContentBlockReference.find_all_in_document(html_content).uniq.each do |content_block|
-      code = content_block.embed_code
-      html_content.gsub!(code, ContentBlockTools::ContentBlock.from_embed_code(code).render)
+    references.each do |ref|
+      block = build_from_publishing_api(ref)
+      html_content.gsub!(ref.embed_code, block.render)
     end
 
     html_content.strip
+  end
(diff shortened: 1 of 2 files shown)


young-rob (approved):
LGTM 👍

PR #3439, opened 2026-08-13
2 page fact check request flow

PeterHattyar:
[MAIN-7432](https://gov-uk.atlassian.net/browse/MAIN-7432)

2 page fact check request flow:

- Set up a 2nd page with appropriate routes for fact check
- Configured email preview to follow design
- Adjusted components and submission action to account for 2 Page flow

[MAIN-7432]: https://gov-uk.atlassian.net/browse/MAIN-7432?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -35,6 +35,7 @@ class EditionsController < InheritedResources::Base
                          schedule_page
                          send_to_fact_check
                          send_to_fact_check_page
+                         send_to_fact_check_email_preview_page
                          resend_fact_check_email_page
                          resend_fact_check_email
                          add_edition_note
@@ -56,7 +57,7 @@ class EditionsController < InheritedResources::Base
   before_action only: %i[schedule_page] do
     require_schedulable
   end
-  before_action only: %i[send_to_fact_check_page send_to_fact_check] do
+  before_action only: %i[send_to_fact_check_page send_to_fact_check_email_preview_page send_to_fact_check] do
     require_can_send_fact_check
   end
   before_action only: %i[show admin metadata tagging history related_external_links unpublish] do
@@ -101,7 +102,13 @@ class EditionsController < InheritedResources::Base
   end
 
   def send_to_fact_check_page
-    @form ||= FactCheckRequestForm.new
+    @form = FactCheckRequestForm.new(
+      permitted_fact_check_request_form_params.merge(
+        edition: @resource,
+        user: current_user,
+      ),
+    )
+
     render "secondary_nav_tabs/send_to_fact_check_page"
   end
 
@@ -254,16 +261,18 @@ class EditionsController < InheritedResources::Base
           ),
         )
 
-        if @form.valid?(:send) && send_to_fact_check_for_edition(@resource, @form, comment)
-          flash[:success] = "Sent to fact check"
-          redirect_to edition_path(resource)
+        unless @form.valid?(:send)
+          render "secondary_nav_tabs/send_to_fact_check_page"
           return
         end
 
-        flash.now[:danger] = SERVICE_REQUEST_ERROR_MESSAGE if @form.errors.none?
-
-        # TODO: These two lines not required once below block for legacy flow is removed.
-        render "secondary_nav_tabs/send_to_fact_check_page"
+        if send_to_fact_check_for_edition(@resource, @form, comment)
+          flash[:success] = "Sent to fact check"
+          redirect_to edition_path(@resource)
+        else
+          flash.now[:danger] = SERVICE_REQUEST_ERROR_MESSAGE
+          render "secondary_nav_tabs/send_to_fact_check_email_preview_page"
+        end
         return
       end
 
@@ -282,6 +291,21 @@ class EditionsController < InheritedResources::Base
     end
     flash.now[:danger] = SERVICE_REQUEST_ERROR_MESSAGE
     render "secondary_nav_tabs/send_to_fact_check_page"
+  end
+
+  def send_to_fact_check_email_preview_page
+    @form ||= FactCheckRequestForm.new(
+      permitted_fact_check_request_form_params.merge(
+        edition: @resource,
+        user: current_user,
+      ),
+    )
+
+    if @form.valid?(:send)
+      render "secondary_nav_tabs/send_to_fact_check_email_preview_page"
+    else
+      render "secondary_nav_tabs/send_to_fact_check_page"
+    end
   end
 
   def skip_review
@@ -642,6 +666,8 @@ private
   end
 
   def permitted_fact_check_request_form_params
+    return {} if params[:fact_check_request_form].blank?
+
     params.require(:fact_check_request_form).permit(:email_addresses,
                                                     :reason_for_change,
                                                     :zendesk_number,
--- a/app/lib/fact_check_request_form.rb
+++ b/app/lib/fact_check_request_form.rb
@@ -73,6 +73,30 @@ class FactCheckRequestForm
     }
   end
 
+  def reason_text
+    return if reason_for_change.blank?
+
+    escaped_reason = ERB::Util.html_escape(reason_for_change)
+    formatted_reason = escaped_reason.to_s.gsub(/\r?\n/, "<br>")
+    "Reason for change: “#{formatted_reason}”\n\n"
+  end
+
+  def zendesk_text
+    return if zendesk_number.blank?
+
+    "Zendesk ticket: <span class='non-functional-link'>#{zendesk_number}</span>\n\n"
+  end
+
+  def formatted_deadline
+    deadline ? deadline.strftime("%A %-e %B %Y").strip : "Not specified"
+  end
+
+  def split_email_addresses
+    return [] if email_addresses.blank?
+
+    email_addresses.split(Regexp.union(",", ";")).map(&:strip).compact_blank
+  end
+
 private
 
   def user_has_editor_permissions
@@ -111,11 +135,11 @@ private
     errors.add(:deadline, "Enter a deadline") if deadline.blank?
   end
 
-  def split_email_addresses
-    return if email_addresses.blank?
+  # def split_email_addresses
+  #   return if email_addresses.blank?
 
-    email_addresses.split(Regexp.union(",", ";")).map(&:strip)
-  end
+  #   email_addresses.split(Regexp.union(",", ";")).map(&:strip)
+  # end
 
   def current_content_presenter
     @current_content_presenter ||= EditionPresenterFactory.get_presenter(edition)
--- a/app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
+++ b/app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
@@ -1,57 +1,50 @@
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <%= render "govuk_publishing_components/components/heading", {
-      text: "Example email",
-      heading_level: 2,
-      font_size: "m",
-      padding: true,
-    } %>
-    <p class="govuk-body">This shows the format of the fact check email that will be sent.</p>
+<%= render "govuk_publishing_components/components/heading", {
+  text: "Fact check email preview",
+  heading_level: 2,
+  font_size: "m",
+  padding: true,
+} %>
+<p class="govuk-body">This shows the fact check email that will be sent.<br><br>You cannot open the links in this preview. They will work in the fact check email.
+</p>
 
-    <article class="edit--send-fact-check-email-preview">
-      <header>
-        <p class="govuk-body">Government Digital Service</p>
-      </header>
+<article class="edit--send-fact-check-email-preview">
+  <header>
+    <p class="govuk-body">Government Digital Service</p>
+  </header>
 
-      <%= render "govuk_publishing_components/components/govspeak", {} do %>
-        <%= Govspeak::Document.new(<<~GOV_SPEAK).to_html.html_safe
-            Hi,
+  <%= render "govuk_publishing_components/components/govspeak", {} do %>
+    <%= Govspeak::Document.new(<<~GOV_SPEAK).to_html.html_safe
+        Hi,
 
-            We need you to check the factual accuracy of changes made to ‘#{@edition.title}’ before it’s published on GOV.UK.
+        We need you to check the factual accuracy of changes made to ‘#{@edition.title}’ before it’s published on GOV.UK.
 
-            Reason for change: Your input will go here.
+        #{@form.reason_text}#{@form.zendesk_text}^ Deadline: #{@form.formatted_deadline}
 
-            Zendesk ticket: Your input will go here.
+        ---
 
-            ^ Deadline: DD Month YYYY
+        ## Review the changes
+        <span class="non-functional-link">View the fact check – #{@edition.title}</span>
 
-            ---
+        You can share this link with the subject matter experts.
 
-            ## Review the changes
-            <span class="non-functional-link">View the fact check - #{@edition.title}</span>
+        They can view the information and check the changes, but are not able to submit a response.
 
-            You can share this link with the subject matter experts. The link expires when the content is published.
+        When fact checking, the subject matter experts should:
 
-            They can view the information and check the changes, but are not able to submit a response.
+        * check the changed content for factual accuracy
+        * tell us if anything is wrong, and why
 
-            When fact checking, they should:
+        They do not need to suggest new wording.
 
-            * check the changed content for factual accuracy
-            * tell us if anything is wrong, and why
+        ---
 
-            They do not need to suggest new wording.
+        ## Submit your fact check response
+        <span class="non-functional-link">Respond to the fact check</span>
 
-            ---
+        Only one fact check response can be submitted.
 
-            ## Submit your fact check response
-            <span class="non-functional-link">Respond to the fact check</span>
-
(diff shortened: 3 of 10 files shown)


test/integration/ga4_tracking_edit_test.rb
```diff
@@ -715,6 +715,38 @@ class Ga4TrackingEditTest < JavascriptIntegrationTest
...
+
+        # 4. Trigger GA4 events on Page 2
+        click_link "Cancel"
+        click_button "Confirm and send for fact check"
+
+        event_data = get_event_data
+
+        # Page 2 "Cancel" link navigation event
+        assert_equal "navigation", event_data[0]["event_name"]
+        assert_equal "Cancel", event_data[0]["text"]
```
davidtrussler: Could we test for the same parameters as the "Cancel" and "Continue" clicks on the first page here (and below)? So for "Cancel" that would be the `external`, `link_domain`, `method`, `type` and `url` values and for the submit the `text`, `tool_name` and `type` values. 

I think they would mostly be the same though for "Cancel" the value for `url` would differ (I think we wouldn't need the whole query string for this so maybe just test for the value up to that, unless it's easy to do) and for submitting the form the calue for text would just be "{}" or something like it because there are. no values the user adds to the form.

PeterHattyar: Thank you, applied changes as discussed, looks good, can easily be modified if any request later from James.

app/controllers/editions_controller.rb
```diff
@@ -284,6 +292,21 @@ def send_to_fact_check
     render "secondary_nav_tabs/send_to_fact_check_page"
   end
 
+  def send_to_fact_check_email_preview_page
```
AlexRyanShep, line 296: I think this method will need adding to the `require_editor_permissions` and `require_can_send_fact_check` `before_action` blocks up at the top of the file

app/controllers/editions_controller.rb
```diff
@@ -254,16 +260,18 @@ def send_to_fact_check
...
+        unless @form.valid?(:send)
+          render "secondary_nav_tabs/send_to_fact_check_page"
           return
         end
 
-        flash.now[:danger] = SERVICE_REQUEST_ERROR_MESSAGE if @form.errors.none?
-
-        # TODO: These two lines not required once below block for legacy flow is removed.
-        render "secondary_nav_tabs/send_to_fact_check_page"
+        if send_to_fact_check_for_edition(@resource, @form, comment)
```
AlexRyanShep, line 269: I reckon the overall flow of this method is a decent bit easier to read now :-)

app/lib/fact_check_request_form.rb
```diff
@@ -73,6 +73,24 @@ def update_content_payload
...
+
+    escaped_reason = ERB::Util.html_escape(reason_for_change)
+    formatted_reason = escaped_reason.to_s.gsub(/\r?\n/, "<br>")
+    "Reason for change: “#{formatted_reason}”\n\n"
+  end
+
+  def zendesk_text
+    return if zendesk_number.blank?
+
+    "Zendesk ticket: [#{zendesk_number}](https://govuk.zendesk.com/tickets/#{zendesk_number})\n\n"
```
AlexRyanShep:
```suggestion
    "Zendesk link: [#{zendesk_number}](https://govuk.zendesk.com/tickets/#{zendesk_number})\n\n"
```

app/lib/fact_check_request_form.rb
```diff
@@ -73,6 +73,24 @@ def update_content_payload
...
+  end
+
+  def formatted_deadline
+    deadline ? deadline.strftime("%e %B %Y").strip : "Not specified"
```
AlexRyanShep:
```suggestion
    deadline ? deadline.strftime("%A %-e %B %Y").strip : "Not specified"
```
Fact Check Manager uses a very slightly different strftime string when it builds the emails

app/lib/fact_check_request_form.rb
```diff
@@ -73,6 +73,24 @@ def update_content_payload
     }
   end
 
+  def reason_text
+    return if reason_for_change.blank?
+
+    escaped_reason = ERB::Util.html_escape(reason_for_change)
+    formatted_reason = escaped_reason.to_s.gsub(/\r?\n/, "<br>")
```
AlexRyanShep, line 80: Worth looking into and testing a bit, but I think that Rails has a helper which can replace these two lines

`simple_format(reason_for_change)`

app/lib/fact_check_request_form.rb
```diff
@@ -73,6 +73,24 @@ def update_content_payload
...
+  def zendesk_text
+    return if zendesk_number.blank?
+
+    "Zendesk ticket: [#{zendesk_number}](https://govuk.zendesk.com/tickets/#{zendesk_number})\n\n"
+  end
+
+  def formatted_deadline
+    deadline ? deadline.strftime("%e %B %Y").strip : "Not specified"
+  end
+
```
AlexRyanShep, line 99: I've left a couple comments on a few bitty specifics to make sure we're exactly matching the current emails produced by FCM, but overall this is looking much better! Good job on the refactoring, these make much more sense here 👍

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
```diff
@@ -1,57 +1,59 @@
...
-            ^ Deadline: DD Month YYYY
+        ---
 
-            ---
+        ## Review the changes
+        <span class="non-functional-link">View the fact check - #{@edition.title}</span>
 
-            ## Review the changes
-            <span class="non-functional-link">View the fact check - #{@edition.title}</span>
+        You can share this link with the subject matter experts. The link expires when the content is published.
```
AlexRyanShep:
```suggestion
        You can share this link with the subject matter experts.
```
Not a change you've made, but while we're updating this file we might as well update it: this sentence has been removed from the email template

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
```diff
@@ -1,57 +1,59 @@
...
 
-            You can share this link with the subject matter experts. The link expires when the content is published.
+        They can view the information and check the changes, but are not able to submit a response.
 
-            They can view the information and check the changes, but are not able to submit a response.
+        When fact checking, they should:
```
AlexRyanShep:
```suggestion
        When fact checking, the subject matter experts should:
```
As for previous comment, wording in live email has been updated a bit

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
```diff
@@ -1,57 +1,59 @@
...
 
-            Zendesk ticket: Your input will go here.
+        #{@form.reason_text}#{@form.zendesk_text}^ Deadline: #{@form.formatted_deadline}
 
-            ^ Deadline: DD Month YYYY
+        ---
 
-            ---
+        ## Review the changes
+        <span class="non-functional-link">View the fact check - #{@edition.title}</span>
```
AlexRyanShep:
```suggestion
        <span class="non-functional-link">View the fact check – #{@edition.title}</span>
```

app/views/editions/secondary_nav_tabs/send_to_fact_check_email_preview_page.html.erb
```diff
@@ -0,0 +1,50 @@
+<% @edition = @resource %>
+<% title = "Check email and send for fact check" %>
+<% content_for :title_context, @edition.title %>
+<% content_for :page_title, title %>
+<% content_for :title, title %>
+
+<% content_for :back_link, send_to_fact_check_page_edition_path(
```
AlexRyanShep, line 7: As discussed, use this for now and ignore the field length issue that we found 🙂 Once I've got my solution for that working on FCM (and this is merged), I'm going to raise a quick PR updating this page to use the fixed version.

app/views/editions/secondary_nav_tabs/send_to_fact_check_email_preview_page.html.erb
```diff
@@ -0,0 +1,50 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <% email_items = @form.email_addresses.to_s.split(/[,;]/).map(&:strip).compact_blank %>
+
+    <%= render "govuk_publishing_components/components/inset_text" do %>
+      <p class="govuk-body">The fact check email will be sent to:</p>
+      <%= render "govuk_publishing_components/components/list", {
+        visible_counters: true,
+        items: email_items,
+      } %>
+    <% end %>
```
AlexRyanShep: `FactCheckRequestForm` already has a helper method for this, `split_email_addresses`

It's currently a private method as it was only used inside the class previously, but I think it would make sense to move it now to be a public method and use it here rather than duplicating the code.

app/controllers/editions_controller.rb
```diff
@@ -284,6 +292,21 @@ def send_to_fact_check
...
 
+  def send_to_fact_check_email_preview_page
+    @form ||= FactCheckRequestForm.new(
+      permitted_fact_check_request_form_params.merge(
+        edition: @resource,
+        user: current_user,
+      ),
+    )
+
+    if @form.valid?(:send)
```
AlexRyanShep, line 304: This is a good catch to have!

test/functional/editions_controller_test.rb
```diff
@@ -1325,6 +1325,49 @@ class EditionsControllerTest < ActionController::TestCase
           edition.reload
           assert_equal "ready", edition.state
         end
+
+        should "render Page 2 preview when valid parameters are provided" do
```
AlexRyanShep: Could we get one more test here please testing that "User somehow makes it to page 2 with an invalid form object" scenario?

AlexRyanShep: Oh and just to keep the tests orderly, can that test and this one go in their own context up from this one? 

If you look up a context to line 1142, this is currently inside a context specifically testing the `#send_to_fact_check` method, so this ideally would go in a `#send_to_fact_check_email_preview_page` context on the same level as that one

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
```diff
@@ -1,57 +1,59 @@
...
-        <p class="govuk-body">Government Digital Service</p>
-      </header>
+<%# Note-begin: Temporarily placing this block here to adjust non functional link behaviour in the preview, as discussed during desk check %>
+<style>
+  .edit--send-fact-check-email-preview .non-functional-link {
+    pointer-events: none;
+    cursor: default;
+  }
+</style>
+<%# Note-end %>
```
AlexRyanShep: I think this block needs moving out into `app/assets/stylesheets/edit.scss`?

AlexRyanShep (changes requested):
This is looking good Peter, good work with the refactor!

I've left a couple of comments, mostly minor things, then I reckon this is good to go!

app/lib/fact_check_request_form.rb
PeterHattyar: I agree with this, however I'm afraid it says 'ticket' because that's what it says on the [figma design](https://www.figma.com/design/UUNr4fuY8A9KYQ0efK0sz6/Mainstream?node-id=5007-14923&t=B31zGBl82xrwGSSk-1).

app/lib/fact_check_request_form.rb
PeterHattyar: Great catch, thanks, I also missed this must have changed on the design at some point, now rectified, displaying the name of the day of the week.

app/controllers/editions_controller.rb
PeterHattyar: That makes perfect sense now that I think about it. Can you explain how come it didn't break while testing it, even while missing from these collections?

app/lib/fact_check_request_form.rb
PeterHattyar: While **simple_format** sounds like a good idea, I think I can't particularly use it here specifically for several reasons:

- It is set up this way to retain formatting as requested. Introducing **simple_format** breaks it with HTML paragraph tags it would wrap it in:

Current: Reason for change: “text< br>text”\n\n
(I've put a space in the tags here just so you can see the tag as the comment would otherwise literally apply it to my content here as html :) )

Using **simple_format**: Reason for change: “< p>text< br />text< /p>”\n\n <- this will cause inline HTML breaks.

- Another thing I've found is that unfortunately this text helper belongs to ActionView. Form objects don't include these helpers by default, so it'd need to be pulled in/referenced which would introduce extra lines of code for this sole reason. But I admit that is a secondary issue, the main problem seems to be it would break formatting. 

I think I'd like to stick with the current approach, but if you'd like to push for using simple_format still, please do tell!

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
PeterHattyar: You're absolutely right, it's gone from the design too, I have missed this, thank you for calling it out!
Removed the sentence.

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
PeterHattyar: Correct, like the previous one, amended.

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
PeterHattyar: Uhmmm...sorry could you help me out with this? At first I thought there is no change here...but I can vaguely see the incoming dash is longer than the original? Is that a different character? 

My editor definitely recognises it as something different as it highlights it as such: 
_"The character U+2013 "–" could be confused with the ASCII character U+002d "-", which is more common in source code."_ 

How did you spot this? Shall I still change it?

app/controllers/editions_controller.rb
AlexRyanShep: Looking at it again I realise there's some tests missing, the ones where the user doesn't have editor permissions/where the edition isn't in the right state. You should be able to mostly copy those from other methods though

app/lib/fact_check_request_form.rb
AlexRyanShep: I'm gonna say go with reality rather than the out of date design 😅 Or double-check with Jonny

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
AlexRyanShep: Yeah the email is meant to have the longer em-dash rather than the shorter hyphen that it had here

AlexRyanShep: It's really hard to see in whatever font is used by GitHub isn't it?! If you copy and paste it from the green diff you'll get the right character though

app/views/editions/secondary_nav_tabs/send_to_fact_check_email_preview_page.html.erb
PeterHattyar: Thank you, that makes absolute perfect sense!

I have now removed the **email_items** line, and using **split_email_addresses** in the component.

A bit of a quick thing to pick your brain about, while looking into this, I got suggestions to improve on **split_email_addresses** (to make it safer), and these are the suggestions. Wanting to be careful, I'm currently using our method, as it was, and now public, but do you think these suggestions make sense?

`return if email_addresses.blank?` -> `return [] if email_addresses.blank?`
It says: _View components (like GOV.UK list) expect an Array for items:. Passing nil can cause rendering warnings or errors._

`email_addresses.split(Regexp.union(",", ";")).map(&:strip)` -> `email_addresses.split(/[,;]/).map(&:strip).compact_blank`

It refactored the Regexp.union into a regex on its own for some reason, and suggests adding compact_blank, to account for potential trailing commas entered by a user, which **(&:strip)** would turn into a **""**.
compact_blank is supposed to clear those, in case they have occured.

What do you think? Are these useful changes for **split_email_addresses**, or should we leave it as is?

app/views/editions/secondary_nav_tabs/_send_fact_check_email_preview.html.erb
PeterHattyar: I believe this needs to be here for now, because _edit.scss_ already has a default behaviour for **.non-functional-link**.

This section here modifies **.non-functional-link** behaviour (by adding **pointer-events: none**; and **cursor: default;**) strictly for this preview page only, as requested during desk check. That's why I didn't want to tamper with the all over  **.non-functional-link** in _edit.scss._, as elsewhere these particular behaviours might not be required where **.non-functional-link** is being utilised.

app/lib/fact_check_request_form.rb
PeterHattyar: Consulted with Jonny, it'll stay as Zendesk ticket, due to feedback as follows:

_"The content designer Gaynor who has been helping us approved and signed off all the content. I remember discussions around this and we didn't feel it was necessary to state that it is a link, as it's visually prominent, and users are familiar with the 'Zendesk ticket' terminology."_

AlexRyanShep (approved):
Thanks for adding the extra tests, looks good to me! Good work on this Peter! :-)

PR #3454, opened 2026-08-27
Remove `:restrict_access_by_org` feature toggle

young-rob:


hezzimo (approved):
Looks great! That was super speedy.

PR #3463, opened 2026-09-04
[MAIN-7536] Send every editable field for other content types

young-rob:
[MAIN-7536](https://gov-uk.atlassian.net/browse/MAIN-7536)

### Changes
Place, transaction and local transaction editions were sent to Fact Check Manager as a single merged body, which excluded some fields for visible content. Each format now declares `FIELD_LABELS` and sends one block per editable field, in the order they appear on the edit form and using that form's own labels. Blank fields are dropped, so Fact Check Manager marks a field the editor has emptied as "(REMOVED)" rather than showing an empty section.

Guides already sent a block per part and now override the same `fact_check_blocks` method rather than the whole presenter method. An empty guide part keeps its heading, because that section still exists on the page.

Also fixes `GOVSPEAK_FIELDS` on `LocalTransactionEdition`, which named `before_text` and `after_text` when the columns are `before_results` and `after_results`, so neither field was checked by `LinkValidator` or `SafeHtml`. Editions with a malformed link in either will now fail validation on save.

[MAIN-7536]: https://gov-uk.atlassian.net/browse/MAIN-7536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/local_transaction_edition.rb
+++ b/app/models/local_transaction_edition.rb
@@ -15,7 +15,7 @@ class LocalTransactionEdition < ApplicationRecord
 
   after_validation :merge_errors
 
-  GOVSPEAK_FIELDS = %i[introduction more_information need_to_know before_text after_text].freeze
+  GOVSPEAK_FIELDS = %i[introduction more_information need_to_know before_results after_results].freeze
 
   validates :lgsl_code, presence: { message: "Enter a LGSL code" }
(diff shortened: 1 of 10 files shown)


stacy-hayes (approved):
Nicely done!

